PR #5, opened 2023-07-20
Fix some permlinks

xianzhidu:


axlearn/vision/augment.py
```diff
@@ -86,7 +86,7 @@ def _convert_translation_to_transform(translations: tf.Tensor) -> tf.Tensor:
 
     Returns:
         A transformation matrix of shape (num_images, 8) to be used by
-        https://github.com/keras-team/keras/blob/v2.9.0/keras/layers/preprocessing/image_preprocessing.py#L898-L985
```
markblee, line 89: FWIW I think tags are ok too.

markblee (approved):

markblee (approved):

PR #20, opened 2023-08-03
torch: attention softmax as float32 for bf16

gyin94:


ruomingp (approved):
Is there a unittest that can be updated?

gyin94:
@ruomingp good question. the unit test is running on CPU. May need to figure out how pytorch community is running bfloat16 for their implementation

PR #32, opened 2023-08-07
Adds T5-UL2 configuration

mehrnaztavan:
add UL2 architecture https://arxiv.org/pdf/2205.05131.pdf


markblee (approved):
Thanks! Should we also make the PR title a bit more specific (e.g., "Adds T5-UL2 configuration")? IIRC, UL2 can also refer to the pretraining objective, which seems to be out of scope of this PR.

PR #35, opened 2023-08-11
Add Request ID to HTTP requests

jiya-zhang:
Potentially fixing recent error while creating queued resource

### DIFF
--- a/axlearn/gcp/tpu.py
+++ b/axlearn/gcp/tpu.py
@@ -7,6 +7,7 @@ import json
 import pathlib
 import re
 import time
+import uuid
 from dataclasses import dataclass
 from datetime import datetime
 from typing import Any, Dict, List, Optional, Sequence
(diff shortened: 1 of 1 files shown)


markblee (commented):
As sync'ed offline, it seems like we're still hitting the 409's even with the unique ID.

ruomingp:
Do we still need this PR?

jiya-zhang:
@ruomingp Sorry this thread hasn't been the top priority but I can look into this now after my ooo. I believe some users are still seeing an error to create TPU on the first try and we'd like to improve the user experience, but the account team is stuck on reproducing the issue (haven't been able to repro)

ruomingp:
> @ruomingp Sorry this thread hasn't been the top priority but I can look into this now after my ooo. I believe some users are still seeing an error to create TPU on the first try and we'd like to improve the user experience, but the account team is stuck on reproducing the issue (haven't been able to repro)

Got it. Thank you!

PR #37, opened 2023-08-18
fix alibi logits bias order

gyin94:
In `apply_attention_logit_biases` function, attention_logit_biases would be cased to logits dtype. So the order matters to pass logits as the first argument
```python
return logits + attention_logit_biases.to(dtype=logits.dtype, device=logits.device)
```

### DIFF
--- a/axlearn/common/adapter_torch.py
+++ b/axlearn/common/adapter_torch.py
@@ -1387,7 +1387,7 @@ class ALiBiAttentionLogitBiasLayer(CausalAttentionLogitBiasLayer):
         )
         # Causal mask.
         return apply_attention_logit_biases(
-            alibi_mask, super().forward(segment_ids=segment_ids, positions=positions)
+            super().forward(segment_ids=segment_ids, positions=positions), alibi_mask
         )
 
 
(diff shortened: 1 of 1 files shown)


axlearn/common/adapter_torch.py
```diff
@@ -1387,7 +1387,7 @@ def forward(
         )
         # Causal mask.
         return apply_attention_logit_biases(
```
ruomingp, line 1389: Great catch! Could you change `apply_attention_logit_biases` to take `apply_attention_logit_biases` as a keyword arg? This also better matches https://docs.google.com/document/d/1tK3MyfZQgXyrvWNg3rt6NQCItuFnFw_FBd6-C7pE2Wk/edit#heading=h.d1xks5sf41jd.

Feel free to do so in a follow-up PR.

ruomingp (approved):
Is there a unittest that can catch this bug?

axlearn/common/adapter_torch.py
gyin94: sure. added a commit for this. please help review again. thanks.

ruomingp: Thanks! Is there any other call site of `apply_attention_logit_biases`?

ruomingp (approved):

ruomingp:
> Is there a unittest that can catch this bug?

Is it possible to add a unittest?

gyin94:
@ruomingp I think so. I will add a follow up PR on this

PR #41, opened 2023-08-23
Make logits partition specs configurable in decoder.

xianzhidu:


axlearn/common/decoder.py
```diff
@@ -422,6 +422,12 @@ class Config(BaseLayer.Config):
         lm_head: Optional[InstantiableConfig] = None
         pad_token_id: int = 0  # Int ID of the inputs to be masked for self-attention.
         eos_token_id: int = 1  # Int ID of the end of sequence token id.
+        # Specifies how to partition the output logits of shape [batch, max_seq_len, vocab_size].
+        logits_partition_spec: Tuple[Union[Optional[str], Tuple[Optional[str]]], ...] = (
```
markblee, line 426: This is fine for now, but I wonder if the inner Tuple should also support len > 2 in theory.

markblee (approved):

PR #45, opened 2023-08-24
move set_layer_norm_eps_recursively from bert.py to layer.py

zhiyun:
move set_layer_norm_eps_recursively from bert.py to layer.py and add unit test.


markblee (approved):

markblee (commented):
Are we missing some tests?

zhiyun:
Good catch!

markblee (approved):

PR #51, opened 2023-08-29
Removes timestamp information in vprint

ruomingp:
... because the timestamp reflects when the python code is traced, not when the compiled program actually runs, so it can be more misleading than useful.

Also addresses review comments on read_state_spec.


ruomingp:
A gentle ping. Thanks!

markblee (approved):

PR #89, opened 2023-09-30
Adds support for torch param conversion for LayerNormStateless.

ruomingp:


axlearn/common/param_converter_test.py
```diff
@@ -176,6 +188,26 @@ def test_layer_norm(self):
             ref_layer=hf_layer,
             test_inputs=[inputs],
             ref_inputs=as_torch_tensor(inputs),
+            test_torch_to_axlearn=True,
+        )
+        self.assertNestedAllClose(out, hf_out)
+
+    def test_layer_norm_stateless(self):
```
markblee, line 195: Consider parameterizing the `test_layer_norm` test? It may also make it easier to check the invalid elementwise_affine codepath.

axlearn/common/param_converter_test.py
```diff
@@ -583,7 +615,7 @@ def test_roundtrip(self, arch, repeat, fuse_qkv):
                 hf_layer_copy(**hf_inputs).logits,
             ),
         )
-        self.assertNestedAllClose(expected, actual)
+        self.assertNestedAllClose(expected, actual, atol=1e-5, rtol=1e-3)
```
markblee, line 618: Intended change?

markblee (approved):

axlearn/common/param_converter_test.py
ruomingp: Good point. Done.

axlearn/common/param_converter_test.py
ruomingp: Yes, the original test was too strict, it seems, and failed without the change.

markblee (approved):

PR #102, opened 2023-10-06
fix issue where flax structs were not recursed into by tree_paths

apghml:


axlearn/common/utils.py
```diff
@@ -158,6 +159,8 @@ def visit(tree, prefix):
                 (k, visit(v, _concat(prefix=prefix, suffix=k, separator=separator)))
                 for k, v in tree.items()
             )
+        elif isinstance(tree, flax.struct.PyTreeNode):
```
ruomingp, line 162: Is there a test that can be added to catch the issue?

ruomingp (approved):

axlearn/common/utils.py
apghml: Added.

markblee (approved):

PR #121, opened 2023-10-13
upgrade transformers version

gyin94:
- the latest transformers (4.28+) would allow an overwrite of build_alibi_tensor


ruomingp (commented):
Thanks! This change is already in #119.

PR #122, opened 2023-10-14
Add query_scale and key_scale to MultiheadAttention.

apghml:


ruomingp (approved):

markblee (approved):

axlearn/common/attention.py
```diff
@@ -1108,6 +1110,28 @@ def forward(self, x: Tensor) -> Tensor:
...
+        value: The value to scale by.
+
+    Example:
+        `query_scale = config_for_function(constant_scale).set(value=0.01)`
+
+    Returns:
+        A config that scales by `value`.
+    """
+
+    def constant_function(_: float, value: float) -> float:
```
ruomingp, line 1129:
```suggestion
    def constant_function(per_head_dim: int, value: float) -> float:
```

ruomingp (approved):

axlearn/common/attention.py
apghml: Oops, it got automerged before I made this change.

However, making this change would cause a pytype failure due to the unused argument. Should I make the change and tell pytype to ignore it?

PR #128, opened 2023-10-17
Fixes large memory usage by pipeline

ruomingp:
Pipeline takes carry inputs of shape [M=num_microbatches, microbatch_size, ...].

Currently the code pads carry tensors to shape [M + N - 1, N=num_layers, microbatch_size, ...] to fully materialize the inputs to every layer on each iteration. This blows up memory usage when carry is large, e.g., one of the tensors can be [batch, num_heads, seq_len, seq_len], representing ALiBi attention logit biases.

Full materialization before the loop is actually unnecessary. Instead, we can pad carry to [M + N - 1, 1, microbatch_size, ...]. Within each iteration, we can generate inputs of shape [N, microbatch_size, ...] by concatenating the inputs to layer 0 with
the outputs of layer [0..N-2].


markblee (approved):
Thanks!

PR #135, opened 2023-10-21
Implement rank_by_value with support for tied ranks.

tuzhucheng:
Implement rank_by_value with support for tied ranks.

This is useful to produce ranks for `ranking_pairwise_loss`, https://github.com/apple/axlearn/blob/4a77cd89ea25a8f0e92029272b6824338fcf8c83/axlearn/common/loss.py#L1203 where the ranks are obtained by sorting another field.

### DIFF
--- a/axlearn/common/input_ranking.py
+++ b/axlearn/common/input_ranking.py
@@ -1,0 +1,68 @@
+# Copyright © 2023 Apple Inc.
+
+"""Input processing utilities on tf.data for ranking-related tasks."""
+from typing import Dict
+
+import seqio
+import tensorflow as tf
+
+from axlearn.common import input_tf_data
+
+
+def rank_by_value(
+    *, input_field: str, output_field: str, ascending: bool = True, allow_ties: bool = False
+) -> input_tf_data.DatasetToDatasetFn:
+    """Returns a DatasetToDatasetFn that stores the ranks of input_field in output_field.
+
+    Note the rank starts at 1.
+
+    Args:
+        input_field: The field whose value will be ranked.
+        output_field: The field to store the ranks into.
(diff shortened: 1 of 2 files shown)


axlearn/common/input_ranking.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Dict
+
+import seqio
+import tensorflow as tf
+
+from axlearn.common import input_tf_data
+
+
+def rank_by_value(
+    *, input_key: str, output_key: str, ascending: bool = True, allow_ties: bool = False
```
ruomingp, line 13: Nit: remove default values, as they are not obvious choices, so spelling them out explicitly at call sites will make the code more readable.

```suggestion
    *, input_key: str, output_key: str, ascending: bool, allow_ties: bool,
```

axlearn/common/input_ranking.py
```diff
@@ -0,0 +1,54 @@
...
+) -> input_tf_data.DatasetToDatasetFn:
+    """Returns a DatasetToDatasetFn that stores the ranks of input_field in output_field.
+
+    Note the rank starts at 1.
+
+    Args:
+        input_key: The field whose value will be ranked.
+        output_key: The field to store the ranks into.
+        ascending: True to rank in ascending order or False to rank in descending order.
+        allow_ties: If true, multiple elements could have the same rank. Ranks could have gaps
```
ruomingp, line 23: Comment on how we break ties when allow_ties=False.

tuzhucheng: Done.

axlearn/common/input_ranking.py
tuzhucheng: Done.

ruomingp (approved):

PR #154, opened 2023-10-29
Adds GroupedQueryAttention.

markblee:


axlearn/common/attention_test.py
```diff
@@ -1528,32 +1595,110 @@ def test_data_types(self, dtype: jnp.dtype, per_dim_scale: Optional[PerDimScale.
...
+        input_linear=(
+            None,  # Use the default linear.
+            attention.QKVLinear.default_config(),
+            attention.GroupedQKVLinear.default_config().set(num_kv_heads=4),
+            attention.FusedGroupedQKVLinear.default_config().set(num_kv_heads=4),
+        ),
     )
-    def test_extend_step(
-        self, dtype: jnp.dtype, per_dim_scale: Optional[PerDimScale.Config], atten_logit_cap: float
+    def test_gqa_forward(
```
ruomingp, line 1605: Do we test the cases when bias=False?

ruomingp (approved):

axlearn/common/attention_test.py
markblee: Good catch, added.

PR #165, opened 2023-11-04
Moves conformer layers to axlearn/audio.

markblee:
Moves the conformer layers to axlearn/audio + minor docstring updates.


ruomingp (commented):
Strictly speaking Conformer is not audio-specific. It can also be used in LMs.

PR #178, opened 2023-11-13
Adds experiments/text/gpt and a trainer config to train a GPT model on C4.

ruomingp:


axlearn/cloud/common/config.py
```diff
@@ -176,7 +176,6 @@ def _locate_user_config_file() -> Optional[str]:
     config_file = None
     for path in search_paths:
         if os.path.exists(path):
-            logging.log_first_n(logging.INFO, "Found user config at %s", 1, path)
```
markblee, line 179: Missing a rebase?

ruomingp: Done.

markblee (approved):

PR #197, opened 2023-11-23
Supports propagating forward outputs across calculators.

markblee:


axlearn/common/evaler.py
```diff
@@ -366,20 +393,65 @@ def __init__(
...
+        for src, dst, dst_key in self._dependencies():
+            if src in self._calculator_dag[dst]:
+                raise ValueError(f"Encountered duplicate edge ({src}, {dst}).")
+            self._calculator_dag[dst].add(src)
+            self._edge_names[(src, dst)] = dst_key
+
+            # Make sure we don't have duplicate keys for the same dst.
+            if dst_key in keys_by_dst[dst]:
+                raise ValueError(f"Encountered duplicate key {dst_key} for {dst}.")
+            keys_by_dst[dst].add(dst_key)
```
ruomingp, line 413: Thanks for adding this check. I envision a more strict check where `dst_key` is required to be globally unique.

Strictly speaking it's not necessary, but I think it will make the data model clearer by avoiding confusion on which src produces the dst_key. WDYT?

markblee: Hm, it seems useful to support fan out cases like (calculator1, calculator[2-N], "forward_outputs"), unless I misunderstood the comment?

ruomingp: Oh, I meant that we shouldn't have multiple src's that produce the same dst_key. Right now it's possible as long as the dst sets do not overlap.

ruomingp (commented):
ack.

axlearn/common/evaler.py
markblee: I see. Updated.

ruomingp (approved):

PR #200, opened 2023-11-24
Fix typo in causal_lm.

tgunter:


markblee (approved):
Thanks! Maybe add a test?

tgunter:
> Thanks! Maybe add a test?

@markblee We have a test, but it requires a mesh with >1 device.  Will need to add functionality to mock the mesh | devices as a followup.

PR #203, opened 2023-11-27
Adds `fuji-7B-single`.

ruomingp:


axlearn/experiments/text/gpt/c4_trainer.py
```diff
@@ -19,6 +19,22 @@
...
         --mesh_selector=$INSTANCE_TYPE
+
+wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh; \
+bash Miniconda3-latest-Linux-x86_64.sh; \
+bash
+conda create -n axlearn python=3.10; \
+conda activate axlearn; \
+git clone https://github.com/apple/axlearn.git; \
+cd axlearn; \
+git fetch origin pull/203/head:debug_gpu; \
```
markblee, line 30: Is this intended for merge, or just to share w/ jax team?

markblee (approved):

axlearn/experiments/text/gpt/c4_trainer.py
markblee: On GPU we may need to add some details about how we build jaxlib cuda (or see if we can repro with a pre-built jaxlib cuda version).

markblee (approved):

axlearn/experiments/text/gpt/c4_trainer.py
ruomingp: Agreed. If you could follow up with a PR to fix the reproduction sequence, that will be great.

PR #205, opened 2023-11-27
Relax jax version in dependencies.

markblee:
Copy of https://github.com/apple/axlearn/pull/204 which seems to be stuck processing updates.


markblee:
Looks like an ongoing github incident... https://www.githubstatus.com/

PR #211, opened 2023-11-30
CTC loss is only valid when input_length >= output_length + num_of_consecutive duplications in label sequence

yqwangustc:
For invalid sequence, optax.ctc_loss will return -log(eps), default -1e5 as the loss. Currently we only drop sequence is the input_length is shorter than output_length, but there are edge cases, this could  result in invalid sequence used in the loss calculation, and a loss value  spike. In the backward pass, the offending sequence get 0 gradient, but the other sequence in the same batch get a very high gradient. 

For example, if the input sequence has 3 frames, and the label sequence is [a, a, b], CTC cannot align the label sequence with the input sequence, because it is require from a to a, it must travel a blank symbol.

### DIFF
--- a/axlearn/audio/asr_decoder.py
+++ b/axlearn/audio/asr_decoder.py
@@ -40,21 +40,21 @@ def _is_valid_ctc_seq(
     considered as an invalid sequence and the loss should be ignored.
 
     A validity check is passed if for an example when:
-      input.length >= labels.length + num(consecutive dup label tokens)
+        input.length >= labels.length + num(consecutive dup label tokens)
 
     Args:
-      paddings: A 0/1 tensor of shape [batch_size, num_frames], indicating whether
-        an input frame is a padding.
-      target_labels: A int32 tensor of shape [batch_size, num_frames].
-      target_paddings: A 0/1 tensor of shape [batch_size, num_frames], indicating
-        whether a label is a padding. Note that at the moment, `target_paddings`
-        must be left-justified, i.e., it must starts with 0 and followed by 1, and
-        not transition back to 0.
-        TODO(yqw): support generic target_paddings.
+        paddings: A 0/1 tensor of shape [batch_size, num_frames], indicating whether
+            an input frame is a padding.
+        target_labels: A int32 tensor of shape [batch_size, num_frames].
+        target_paddings: A 0/1 tensor of shape [batch_size, num_frames], indicating
+            whether a label is a padding. Note that at the moment, `target_paddings`
+            must be left-justified, i.e., it must starts with 0 and followed by 1, and
+            not transition back to 0.
+            TODO(yqw): support generic target_paddings.
 
     Returns:
-      A float tensor of [batch_size, ] indicating if each (input, label) pair is valid,
-      with a value of 1.0 indicating valid and 0.0 otherwise.
+        A float tensor of [batch_size, ] indicating if each (input, label) pair is valid,
+        with a value of 1.0 indicating valid and 0.0 otherwise.
     """
     # [batch_size, ]
     label_lengths = jnp.sum(1.0 - target_paddings, axis=-1)
--- a/axlearn/audio/asr_decoder.py
+++ b/axlearn/audio/asr_decoder.py
@@ -47,7 +47,10 @@ def _is_valid_ctc_seq(
         an input frame is a padding.
       target_labels: A int32 tensor of shape [batch_size, num_frames].
       target_paddings: A 0/1 tensor of shape [batch_size, num_frames], indicating
-        whether a label is a padding.
+        whether a label is a padding. Note that at the moment, `target_paddings`
+        must be left-justified, i.e., it must starts with 0 and followed by 1, and
+        not transition back to 0.
+        TODO(yqw): support generic target_paddings.
 
     Returns:
       A float tensor of [batch_size, ] indicating if each (input, label) pair is valid,
--- a/axlearn/audio/asr_decoder.py
+++ b/axlearn/audio/asr_decoder.py
@@ -28,31 +28,35 @@ from axlearn.common.module import Module
 from axlearn.common.utils import Nested, Tensor
 
 
-def is_valid_ctc_seq(logitpaddings, labels, labelpaddings):
+def _is_valid_ctc_seq(
+        paddings: Tensor,
+        target_labels: Tensor,
+        target_paddings: Tensor) -> Tensor:
     """Returns for per example sequence if it passes validity check.
 
-    Note that the above `ctc_loss_with_alignments` returns logeps
-    (usually a very large number) if the input length is smaller than
-    the label length plus number of consectutive duplications.
-    However, in that case, we should ignore the loss.
+    Note that `optax.ctc_loss` returns logeps (default to 1e5) if the
+    input length is smaller than the label length plus number of 
+    consectutive duplications. However, in that case, we should 
+    ignore the loss.
 
     A validity check is passed if for an example when :
       input.length >= labels.length + num(consecutive dup label tokens)
 
     Args:
-      logitpaddings:  [b, t], 0/1 JTensor.
-      labels:         [b, t], int32 JTensor.
-      labelpaddings:  [b, t], 0/1 JTensor.
+      paddings:              [b, t], 0/1 Tensor.
+      target_labels:         [b, t], int32 Tensor.
+      target_paddings:       [b, t], 0/1 Tensor.
+
     Returns:
       A shape [b] float tensor indicating if each (input, label) pair is valid,
-        with a value of 1.0 indicating valid and 0.0 otherwise.
+      with a value of 1.0 indicating valid and 0.0 otherwise.
     """
     # [b]
-    label_lengths = jnp.sum(1.0 - labelpaddings, axis=-1)
+    label_lengths = jnp.sum(1.0 - target_paddings, axis=-1)
     # [b]
-    input_lengths = jnp.sum(1.0 - logitpaddings, axis=-1)
+    input_lengths = jnp.sum(1.0 - paddings, axis=-1)
     # [b, t-1]
-    dups = (1.0 - labelpaddings[:, 1:]) * (labels[:, :-1] == labels[:, 1:])
+    dups = (1.0 - target_paddings[:, 1:]) * (target_labels[:, :-1] == target_labels[:, 1:])
     # [b]
     num_consecutive_dups = jnp.sum(dups, axis=-1)
     # [b]
@@ -218,7 +222,7 @@ class CTCDecoderModel(BaseModel):
         )
 
         # Drop examples with targets longer than inputs.
-        per_example_weight = is_valid_ctc_seq(paddings, target_labels, target_paddings)
+        per_example_weight = _is_valid_ctc_seq(paddings, target_labels, target_paddings)
         per_example_weight = per_example_weight.astype(per_example_loss.dtype)
 
         # Compute weighted loss.
--- a/axlearn/audio/asr_decoder.py
+++ b/axlearn/audio/asr_decoder.py
@@ -29,37 +29,36 @@ from axlearn.common.utils import Nested, Tensor
 
 
 def is_valid_ctc_seq(logitpaddings, labels, labelpaddings):
-  """Returns for per example sequence if it passes validity check.
+    """Returns for per example sequence if it passes validity check.
 
-  Note that the above `ctc_loss_with_alignments` returns logeps
-  (usually a very large number) if the input length is smaller than
-  the label length plus number of consectutive duplications.
-  However, in that case, we should ignore the loss.
+    Note that the above `ctc_loss_with_alignments` returns logeps
+    (usually a very large number) if the input length is smaller than
+    the label length plus number of consectutive duplications.
+    However, in that case, we should ignore the loss.
 
-  A validity check is passed if for an example when :
-    input.length >= labels.length + num(consecutive dup label tokens)
+    A validity check is passed if for an example when :
+      input.length >= labels.length + num(consecutive dup label tokens)
 
-  Args:
-    logitpaddings:  [b, t], 0/1 JTensor.
-    labels:         [b, t], int32 JTensor.
-    labelpaddings:  [b, t], 0/1 JTensor.
-  Returns:
-    A shape [b] float tensor indicating if each (input, label) pair is valid,
-      with a value of 1.0 indicating valid and 0.0 otherwise.
-  """
-  # [b]
-  label_lengths = jnp.sum(1.0 - labelpaddings, axis=-1)
-  # [b]
-  input_lengths = jnp.sum(1.0 - logitpaddings, axis=-1)
-  # [b, t-1]
-  dups = (1.0 - labelpaddings[:, 1:]) * (labels[:, :-1] == labels[:, 1:])
-  # [b]
-  num_consecutive_dups = jnp.sum(dups, axis=-1)
-  # [b]
-  is_valid = (label_lengths + num_consecutive_dups) <= input_lengths
-  is_valid = is_valid.astype(jnp.float32)
-  return is_valid
-
+    Args:
+      logitpaddings:  [b, t], 0/1 JTensor.
+      labels:         [b, t], int32 JTensor.
+      labelpaddings:  [b, t], 0/1 JTensor.
+    Returns:
+      A shape [b] float tensor indicating if each (input, label) pair is valid,
+        with a value of 1.0 indicating valid and 0.0 otherwise.
+    """
+    # [b]
+    label_lengths = jnp.sum(1.0 - labelpaddings, axis=-1)
+    # [b]
+    input_lengths = jnp.sum(1.0 - logitpaddings, axis=-1)
+    # [b, t-1]
+    dups = (1.0 - labelpaddings[:, 1:]) * (labels[:, :-1] == labels[:, 1:])
+    # [b]
+    num_consecutive_dups = jnp.sum(dups, axis=-1)
+    # [b]
+    is_valid = (label_lengths + num_consecutive_dups) <= input_lengths
+    is_valid = is_valid.astype(jnp.float32)
+    return is_valid
 
 
 class CTCPrefixMerger(PrefixMerger):
--- a/axlearn/audio/asr_decoder.py
+++ b/axlearn/audio/asr_decoder.py
@@ -28,25 +28,29 @@ from axlearn.common.module import Module
 from axlearn.common.utils import Nested, Tensor
 
 
-def _is_valid_ctc_seq(paddings: Tensor, target_labels: Tensor, target_paddings: Tensor) -> Tensor:
+def _is_valid_ctc_seq(
+    *, paddings: Tensor, target_labels: Tensor, target_paddings: Tensor
+) -> Tensor:
     """Returns for per example sequence if it passes validity check.
 
     Note that `optax.ctc_loss` returns -logeps (default to 1e5) if the
     input length is smaller than the label length plus number of
-    consectutive duplications, because we need an blank to token to transit
-    between the same tokens. When this condition is not met, it should be
+    consecutive duplications, because we need a blank label to transition
+    between the same labels. When this condition is not met, it should be
     considered as an invalid sequence and the loss should be ignored.
 
-    A validity check is passed if for an example when :
+    A validity check is passed if for an example when:
       input.length >= labels.length + num(consecutive dup label tokens)
 
     Args:
-      paddings:              [batch_size, num_frames], 0/1 Tensor, paddings of input frames.
-      target_labels:         [batch_size, num_frames], int32 Tensor.
-      target_paddings:       [batch_size, num_frames], 0/1 Tensor.
+      paddings: A 0/1 tensor of shape [batch_size, num_frames], indicating whether
+        an input frame is a padding.
+      target_labels: A int32 tensor of shape [batch_size, num_frames].
+      target_paddings: A 0/1 tensor of shape [batch_size, num_frames], indicating
+        whether a label is a padding.
 
     Returns:
-      A shape [b] float tensor indicating if each (input, label) pair is valid,
+      A float tensor of [batch_size, ] indicating if each (input, label) pair is valid,
       with a value of 1.0 indicating valid and 0.0 otherwise.
     """
     # [batch_size, ]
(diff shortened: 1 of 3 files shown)


axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,40 @@
 from axlearn.common.utils import Nested, Tensor
 
 
+def is_valid_ctc_seq(logitpaddings, labels, labelpaddings):
```
markblee, line 31:
```suggestion
def _is_valid_ctc_seq(*, paddings: Tensor, target_labels: Tensor, target_paddings: Tensor) -> Tensor:
```

In general, can we add type+return annotations and follow existing naming conventions (following https://github.com/apple/axlearn/blob/main/CONTRIBUTING.md) here and elsewhere?

axlearn/audio/asr_decoder_test.py
```diff
@@ -88,6 +91,80 @@ def test_map_label_sequences(
...
+    per_seq_loss = optax.ctc_loss(
+        logits, logitpaddings, labels, labelpaddings)
+    # per_seq_loss[0:1] and per_seq_loss[3] should near log(400) * 15, while
+    # per_seq_loss[2] should be around logepsilon
+    print(per_seq_loss)
+    per_seq_validality = is_valid_ctc_seq(logitpaddings, labels, labelpaddings)
+    self.assertAllClose(per_seq_validality, [1.0, 1.0, 0.0, 1.0])
+
 
 class CTCPrefixMergerTest(TestCase):
```
markblee, line 163: Shall we also update `test_forward` to make sure we cover the training codepath?

axlearn/common/test_utils.py
```diff
@@ -230,6 +238,16 @@ def assertNestedEqual(self, a, b):
             if hasattr(a_value, "dtype"):
                 self.assertEqual(a_value.dtype, b_value.dtype)
 
+    def assertAllClose(self, x, y, check_dtypes=True, rtol=1e-5, atol=1e-5, **kwargs):
```
markblee, line 241: Do we need this, or can we use `assertNestedAllClose`?

axlearn/audio/asr_decoder_test.py
```diff
@@ -89,6 +91,75 @@ def test_map_label_sequences(
         )
 
 
+class ValidCtcSeqTest(TestCase):
+    def get_logits_and_labels(self, batchsize, timesteps, labelsteps, nclasses):
```
markblee, line 95: Same comment as above re types and naming -- also, should we use `jnp`/`jax.random` here and elsewhere? Besides consistency, we can avoid `np.random` calls that may depend on a global seed?

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,39 @@
...
+    A validity check is passed if for an example when :
+      input.length >= labels.length + num(consecutive dup label tokens)
+
+    Args:
+      logitpaddings:  [b, t], 0/1 JTensor.
+      labels:         [b, t], int32 JTensor.
+      labelpaddings:  [b, t], 0/1 JTensor.
+    Returns:
+      A shape [b] float tensor indicating if each (input, label) pair is valid,
+        with a value of 1.0 indicating valid and 0.0 otherwise.
```
markblee, line 48: Fix docstring following rest of repo?

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,39 @@
 from axlearn.common.utils import Nested, Tensor
 
 
+def is_valid_ctc_seq(logitpaddings, labels, labelpaddings):
+    """Returns for per example sequence if it passes validity check.
+
+    Note that the above `ctc_loss_with_alignments` returns logeps
```
markblee, line 34: What is `ctc_loss_with_alignments` referring to?

axlearn/audio/asr_decoder_test.py
```diff
@@ -89,6 +91,75 @@ def test_map_label_sequences(
...
+        batchsize = 4
+        timesteps = 10
+        labelsteps = 11
+        nclasses = 400
+        # generate logits and labels, which has logits shorter than labels
+        logits, logitpaddings, labels, labelpaddings = self.get_logits_and_labels(
+            batchsize, timesteps, labelsteps, nclasses
+        )
+        per_seq_loss = optax.ctc_loss(logits, logitpaddings, labels, labelpaddings, blank_id=0)
+        print(per_seq_loss)
```
markblee, line 112: nit: Remove debugging prints here and elsewhere?

markblee (commented):
Thanks! Left some initial comments -- could we also add/update a test s.t. it would fail without this change?

axlearn/audio/asr_decoder_test.py
yqwangustc: Here, I try to show that when presnets an invalid sequence, `optax.ctc_loss` returns somethong abnormal. On the other hand, I don't want to assert this loss value, as it can change anytime. What's the best strategy here ? Maybe I can using logging.info instead ?

axlearn/audio/asr_decoder.py
zhiyun: hmm that seems to be a follow up PR.

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,43 @@
...
+    label_lengths = jnp.sum(1.0 - target_paddings, axis=-1)
+    # [b]
+    input_lengths = jnp.sum(1.0 - paddings, axis=-1)
+    # [b, t-1]
+    dups = (1.0 - target_paddings[:, 1:]) * (target_labels[:, :-1] == target_labels[:, 1:])
+    # [b]
+    num_consecutive_dups = jnp.sum(dups, axis=-1)
+    # [b]
+    is_valid = (label_lengths + num_consecutive_dups) <= input_lengths
+    is_valid = is_valid.astype(jnp.float32)
```
zhiyun, line 64: Do we need to cast type here?

axlearn/audio/asr_decoder_test.py
zhiyun: would assert_greater() works?

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,43 @@
 from axlearn.common.utils import Nested, Tensor
 
 
+def _is_valid_ctc_seq(
+        paddings: Tensor,
+        target_labels: Tensor,
+        target_paddings: Tensor) -> Tensor:
+    """Returns for per example sequence if it passes validity check.
+
+    Note that `optax.ctc_loss` returns logeps (default to 1e5) if the
```
zhiyun, line 37: Can we give a reference link to optax implementation on this?

yqwangustc:
> Thanks! Left some initial comments -- could we also add/update a test s.t. it would fail without this change?

Without this change, it won't fail, but it will silently fail with a wrong loss value. I have updated the test cases to reflect the value before and after applying this `is_valid_ctc_seq` check.

yqwangustc:
> Thanks! Left some initial comments -- could we also add/update a test s.t. it would fail without this change?

Without this change, the forward pass will not fail, but it will give an incorrect loss value. With this cl, we will filter out the incorrect loss (i.e., give a loss value of 0). I have also updated the unitest cases, showing the effect of this change (w/o it gives a loss value of 1e5, with this change, the 1e5 loss value will be masked out).

axlearn/audio/asr_decoder_test.py
```diff
@@ -89,6 +91,97 @@ def test_map_label_sequences(
         )
 
 
+class ValidCtcSeqTest(TestCase):
+    def get_logits_and_labels(
+        self, batch_size: int, time_steps: int, target_steps: int, nclasses: int
+    ) -> tuple[Tensor, Tensor, Tensor, Tensor]:
```
zhiyun, line 97: time_steps -> input_lengths
target_steps -> target_lengths
nclasses -> vocab_size

axlearn/audio/asr_decoder_test.py
```diff
@@ -89,6 +91,97 @@ def test_map_label_sequences(
...
+        )
+        target_paddings = jnp.zeros(shape=(batch_size, target_steps), dtype=jnp.int32)
+        return logits, paddings, target_labels, target_paddings
+
+    def test_label_longer_than_input(self):
+        batchsize = 4
+        timesteps = 10
+        labelsteps = 11
+        nclasses = 400
+        # generate logits and labels, which has logits shorter than labels
```
zhiyun, line 116: Nit: Use full sentence for comment: start with Capitalize, and end with period `.`

axlearn/audio/asr_decoder_test.py
```diff
@@ -89,6 +91,97 @@ def test_map_label_sequences(
...
+        logits, paddings, target_labels, target_paddings = self.get_logits_and_labels(
+            batchsize, timesteps, labelsteps, nclasses
+        )
+        per_seq_loss = optax.ctc_loss(logits, paddings, target_labels, target_paddings, blank_id=0)
+        for x in per_seq_loss:
+            # because these are invalid sequence loss, the optax.ctc_loss will return
```
zhiyun, line 122: Same here. Use complete sentence for comment.

axlearn/audio/asr_decoder_test.py
zhiyun: +1. can we check the forward method returns the correct masked loss?

axlearn/common/test_utils.py
```diff
@@ -230,6 +238,16 @@ def assertNestedEqual(self, a, b):
...
+    def assertTensorAllClose(self, x, y, rtol=1e-5, atol=1e-5, check_dtypes=True, **kwargs):
+        """assert Tensor are all close."""
+        x = np.asarray(x)
+        y = np.asarray(y)
+        if check_dtypes:
+            self.assertDtypesMatch(x, y)
+        x = x.astype(np.float32) if x.dtype == jnp.bfloat16 else x
+        y = y.astype(np.float32) if y.dtype == jnp.bfloat16 else y
+        np.testing.assert_allclose(x, y, rtol=rtol, atol=atol, **kwargs)
+
```
zhiyun, line 250: @markblee to comment if we want to add this util.

zhiyun (commented):
Thanks Yongqiang!

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,40 @@
...
+    A validity check is passed if for an example when :
+      input.length >= labels.length + num(consecutive dup label tokens)
+
+    Args:
+      paddings:              [batch_size, num_frames], 0/1 Tensor, paddings of input frames.
+      target_labels:         [batch_size, num_frames], int32 Tensor.
+      target_paddings:       [batch_size, num_frames], 0/1 Tensor.
+
+    Returns:
+      A shape [b] float tensor indicating if each (input, label) pair is valid,
```
ruomingp, line 49:
```suggestion
      A float tensor of shape [batch_size] indicating if each (input, label) pair is valid,
```

axlearn/common/test_utils.py
ruomingp: Can we use `assertNestedAllClose` directly?

ruomingp (commented):
Nice!

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,40 @@
...
 
 
+def _is_valid_ctc_seq(paddings: Tensor, target_labels: Tensor, target_paddings: Tensor) -> Tensor:
+    """Returns for per example sequence if it passes validity check.
+
+    Note that `optax.ctc_loss` returns -logeps (default to 1e5) if the
+    input length is smaller than the label length plus number of
+    consectutive duplications, because we need an blank to token to transit
+    between the same tokens. When this condition is not met, it should be
+    considered as an invalid sequence and the loss should be ignored.
```
markblee, line 38:
```suggestion
    consecutive duplications, because we need a blank to token to transition
    between the same labels. When this condition is not met, it should be
    considered as an invalid sequence and the loss should be ignored.
```

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,40 @@
...
+
+    A validity check is passed if for an example when :
```
markblee, line 40:
```suggestion
    A validity check is passed if for an example when:
```

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,40 @@
...
+      input.length >= labels.length + num(consecutive dup label tokens)
+
+    Args:
+      paddings:              [batch_size, num_frames], 0/1 Tensor, paddings of input frames.
+      target_labels:         [batch_size, num_frames], int32 Tensor.
+      target_paddings:       [batch_size, num_frames], 0/1 Tensor.
```
markblee, line 46:
```suggestion
      paddings: A 0/1 Tensor of shape [batch_size, num_frames] representing paddings of input frames.
      target_labels: An int Tensor of shape [batch_size, num_labels].
      target_paddings: A 0/1 Tensor of shape [batch_size, num_labels].
```

axlearn/audio/asr_decoder_test.py
```diff
@@ -89,6 +91,99 @@ def test_map_label_sequences(
...
+        target_lengths = 10
+        vocab_size = 400
+        logits, paddings, _, target_paddings = self.get_logits_and_labels(
+            batch_size, input_lengths, target_lengths, vocab_size
+        )
+        # there are 12 timesteps, and 10 labels. If the consecutive duplicates in
+        # one sequence is larger than 2, then the pair become non-valid
+        target_labels = np.array(
+            [
+                [0, 1, 2, 3, 4, 5, 6, 7, 8, 9],  # no duplicates
```
markblee, line 165: Consider having a test w/ duplicates but not consecutively?

axlearn/audio/asr_decoder_test.py
```diff
@@ -89,6 +91,99 @@ def test_map_label_sequences(
...
+
+    def test_label_shorter_than_input(self):
+        batch_size = 4
+        input_lengths = 15
+        target_lengths = 10
+        vocab_size = 400
+        logits, paddings, _, target_paddings = self.get_logits_and_labels(
+            batch_size, input_lengths, target_lengths, vocab_size
+        )
+        # to make sure there is no duplicate in the labels
```
markblee, line 140: Please see @zhiyun 's comment re. comment formatting here and below.

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,40 @@
 from axlearn.common.utils import Nested, Tensor
 
 
+def _is_valid_ctc_seq(paddings: Tensor, target_labels: Tensor, target_paddings: Tensor) -> Tensor:
+    """Returns for per example sequence if it passes validity check.
```
markblee, line 32:
```suggestion
def _is_valid_ctc_seq(*, paddings: Tensor, target_labels: Tensor, target_paddings: Tensor) -> Tensor:
    """Returns whether each input sequence passes validity check.
```

axlearn/common/test_utils.py
markblee: +1, I think I left another comment on this as well.

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,40 @@
...
+    Returns:
+      A shape [b] float tensor indicating if each (input, label) pair is valid,
+      with a value of 1.0 indicating valid and 0.0 otherwise.
+    """
+    # [batch_size, ]
+    label_lengths = jnp.sum(1.0 - target_paddings, axis=-1)
+    # [batch_size, ]
+    input_lengths = jnp.sum(1.0 - paddings, axis=-1)
+    # [batch_size, num_frames - 1]
+    dups = (1.0 - target_paddings[:, 1:]) * (target_labels[:, :-1] == target_labels[:, 1:])
```
markblee, line 57: I wonder whether we should consider `jnp.logical_or(target_paddings[:, :-1], target_paddings[:, 1:])` when dropping dups?

markblee (commented):
Thanks @yqwangustc! A friendly reminder to follow https://github.com/apple/axlearn/blob/main/CONTRIBUTING.md (e.g., re coding style and resolving comments).

axlearn/common/test_utils.py
yqwangustc: Unfortunately not. `assertNestedAllClose` will first `flatten_items`. For example, if we do 

```
self.assertTensorAllClose(jnp.array[0, 0, 0, 0], [0] * 4)
```

it will flatten the first item to 

```
[('', Array([1., 1., 1., 1.], dtype=float32))]
```
and the second item to
```
[('0', 1.0), ('1', 1.0), ('2', 1.0), ('3', 1.0)]
```

Obviously, they are different, so the test will fail. 

I also tried something more verbose: 

```
x = {'tensor': jnp.array([1.0] * 4)}
y = {'tensor': [1.0] * 4}
self.assertNestedAllClose(x, y)
```
Again, `flatten` will expand them differently. 

Another possibility is to use something like 

```
x = jnp.array([1.0] * 4)
y = [1.0] * 4
self.assertNestedAllClose(x, jnp.array(y))
```

This will work, but under the hook, it is expanding the jnp/np array to a list of elements, and compare them one-by-one. Of course, if we do

```
x = jnp.array([1.0] * 4)
y = [1.0] * 4
self.assertNestedAllClose(
    {'tensor': x},
    {'tensor': jnp.array(y)})
```

it will work, but this is too troublesome from my biased perspective. 

It is pretty common to compare a jnp/numpy tensor with an expected tensor or list of elements. I have seen people start to use something like https://github.com/apple/axlearn/blob/1a045eae42bda6fcefbf5f4decbebc1b7969ce78/axlearn/vision/nms_test.py#L38, but that requires to subclass from tf.test.TestCase, which brings even more problems from my point of view.

axlearn/audio/asr_decoder_test.py
zhiyun: Current test_forward examples do not explicitly cover duplicate labels cases, feel free to modify the test inputs there? I think that's why Mark referred to.

not sure if i locate Ln 175 to Ln 178 correctly..

axlearn/audio/asr_decoder.py
zhiyun: Oh i mean add this reference in code comment explicitly.

zhiyun (commented):
Thanks! overall LGTM. Wait for resolving Mark and Ruoming's comment.

axlearn/audio/asr_decoder.py
markblee: As discussed offline, `dups` may be miscounted if target paddings appear on the leading duplicate labels. In normal cases, this shouldn't matter since paddings will be inferred from an OOV label, thus the paddings should cover both dups -- but in this case the API is taking a separate `target_paddings` Tensor which may not follow this assumption.

yqwangustc: Thanks. I have added a note in the docstring to explain the expectation of target_paddings,  and a TODO to support any paddings in the future.

axlearn/common/test_utils.py
markblee:
```
x = jnp.array([1.0] * 4)
y = jnp.array([1.0] * 4)
self.assertNestedAllClose(x, y)
```
seems reasonable to me (we treat the Tensors as leaves). Can you elaborate on the concern?

yqwangustc: First we need to convert python list to jax numpy array, which needs extra work in writing test; second this requires the dtype of both x and y must match, but what we only mean to be compare the numerical value.

ruomingp: We can change `assertNestedAllClose` to be more relaxed on dtypes. There are also ways to make it support `[1.0] * 4`. 

We should decouple this change to a separate PR and discuss separately.

yqwangustc: Thanks for the comment. I have removed the change in `test_utils.py`. will follow up in another cl if necessary.

axlearn/audio/asr_decoder.py
markblee: Are the changes pushed?

axlearn/audio/asr_decoder.py
```diff
@@ -28,6 +28,47 @@
...
+    input length is smaller than the label length plus number of
+    consecutive duplications, because we need a blank label to transition
+    between the same labels. When this condition is not met, it should be
+    considered as an invalid sequence and the loss should be ignored.
+
+    A validity check is passed if for an example when:
+      input.length >= labels.length + num(consecutive dup label tokens)
+
+    Args:
+      paddings: A 0/1 tensor of shape [batch_size, num_frames], indicating whether
```
markblee, line 46: nit -- please use 4 space indents.

axlearn/audio/asr_decoder_test.py
markblee: Yeah, ideally we can cover the forward codepath. IIUC, if we don't invoke the `_is_valid_ctc_seq` helper function at all, the forward tests still pass.

axlearn/audio/asr_decoder.py
markblee: Sounds good, thanks.

markblee (commented):
Mostly lgtm, although it seems like the latest updates have not been pushed?

markblee (approved):

axlearn/audio/asr_decoder.py
yqwangustc: Done.

axlearn/audio/asr_decoder.py
yqwangustc: thanks for catching this. just pushed another one to fix it.

axlearn/audio/asr_decoder.py
yqwangustc: Thanks, Done.

axlearn/common/test_utils.py
yqwangustc: Remove this function. Considered done for this thread.

axlearn/audio/asr_decoder.py
yqwangustc: Sorry I think there is no explicit mention in optax how to deal with inaccessible path, so I prefer not to mention specific line in the code.

axlearn/common/test_utils.py
yqwangustc: Done.

axlearn/audio/asr_decoder.py
yqwangustc: Done

axlearn/audio/asr_decoder.py
yqwangustc: Done

yqwangustc:
> Mostly lgtm, although it seems like the latest updates have not been pushed?

Thanks for catching. Still get myself familiar with the reviewing workflow. Add another commit to fix some leftover issues.

markblee (approved):

zhiyun (approved):

PR #222, opened 2023-12-04
Increases FD limit to 100000.

ruomingp:


axlearn/cloud/gcp/jobs/tpu_runner.py
```diff
@@ -350,7 +350,8 @@ def _run_command(self):
         # Set env vars, run the command and pipe outputs to run log.
         # Depending on command returncode, emit either success or failure flag.
         # Note that we use PIPESTATUS[0] to check the returncode of the first command in the pipe.
-        cmd = f"""mkdir -p {self._output_dir}; echo "Starting command..." >> {self._run_log};
+        cmd = f"""ulimit -n 100000;
```
markblee, line 353: Does it work if we set it in `start_tpu.sh`?

ruomingp: Good question. I'm not sure because:

The ulimit command in Unix-like systems is used to control the resources available to the shell and processes started by it. It's important to note that ulimit settings are specific to the shell session in which they are set and do not affect other shell sessions.

PR #227, opened 2023-12-05
Ensures that dataflow is launched from the same image as worker.

markblee:


axlearn/cloud/gcp/jobs/dataflow.py
```diff
@@ -29,7 +29,7 @@
     axlearn gcp dataflow start \
         --name=$USER-dataflow \
         --bundler_spec=dockerfile=Dockerfile \
-        --bundler_spec=base_image=apache/beam_python3.8_sdk:2.38.0 \
+        --bundler_spec=base_image=apache/beam_python3.10_sdk:2.52.0 \
```
zhiyun, line 32: 3.9_sdk:2.52.0

zhiyun (approved):

PR #235, opened 2023-12-08
Changes in preparation for jax 0.4.21.

markblee:


axlearn/common/utils_spmd.py
```diff
@@ -46,29 +48,41 @@ def setup(
...
-            )
+            ):
+                raise ValueError(
+                    "distributed_coordinator, num_processes, and process_id "
+                    "should all be None for tpu backend."
+                )
             jax.distributed.initialize(
                 coordinator_address=_infer_tpu_coordinator_address(),
                 num_processes=jax.process_count(),
                 process_id=jax.process_index(),
```
markblee, line 62: Will remove these when we actually bump the jax version.

PR #251, opened 2023-12-15
update dataflow script

zhiyun:
This PR makes 2 changes:
1. disable running from docker on Mac M1, to avoid qemu core dump issue https://github.com/docker/for-mac/issues/5342.
2. remove default number_of_worker_harness_threads=1. When user submit job that requires this flag, it can be set by passing --dataflow_spec=number_of_worker_harness_threads=1.

### DIFF
--- a/axlearn/cloud/gcp/jobs/dataflow.py
+++ b/axlearn/cloud/gcp/jobs/dataflow.py
@@ -72,6 +72,7 @@ https://cloud.google.com/dataflow/docs/quickstarts/create-pipeline-python#run-th
 """
 # pylint: disable=protected-access
 
+import platform
 import re
 import shlex
 import signal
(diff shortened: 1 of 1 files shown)


axlearn/cloud/gcp/jobs/dataflow.py
```diff
@@ -229,13 +229,25 @@ def _delete(self):
...
-
         # Run the setup command locally, but the launch command via docker.
         # This is to ensure that the launch environment matches the worker environment.
-        cmd = (
-            "docker run --rm --entrypoint /bin/bash "
-            f"{self._bundler.id(cfg.name)} -c '{cfg.command}'"
-        )
+        processor = platform.processor().lower()
+        if "arm" in processor:
+            # Disable running from docker on Mac M1 chip due to quemu core dump bug.
```
markblee, line 236:
```suggestion
            # Disable running from docker on Mac M1 chip due to qemu core dump bug.
```

markblee (approved):

axlearn/cloud/gcp/jobs/dataflow.py
zhiyun: Thanks! Done.

zbwglory (approved):

PR #256, opened 2023-12-18
Moves the max_start_seconds check so that we start checking summaries…

ruomingp:
… from the beginning for faster debugging.


JianyuWangV (approved):
Thanks!

PR #259, opened 2023-12-19
Increases file descriptor limit for root to 100K in start_tpu.sh.

ruomingp:


axlearn/cloud/gcp/scripts/start_tpu.sh
```diff
@@ -12,6 +12,9 @@ echo "=== AXLearn start_tpu.sh ==="
 # Random sleep to prevent all TPU-VMs overwhelming pypi etc for large slices.
 sleep $((1 + $RANDOM % 30))
 
+sudo sh -c "echo 'root soft nofile 100000' >> /etc/security/limits.conf"
```
tgunter, line 15: nit: Is it worth a comment to explain?

tgunter (approved):

samos123 (approved):

axlearn/cloud/gcp/scripts/start_tpu.sh
ruomingp: Done.

axlearn/cloud/gcp/scripts/start_tpu.sh
```diff
@@ -12,6 +12,10 @@ echo "=== AXLearn start_tpu.sh ==="
 # Random sleep to prevent all TPU-VMs overwhelming pypi etc for large slices.
 sleep $((1 + $RANDOM % 30))
 
+# Increase file descriptor limits for `root` to avoid "Too many open files" errors.
```
samos123, line 15: nit: the change the comment to an echo statement so someone reading the logs knows what's going on. That would have helped in catching whether the correct startup script was run as well.

ruomingp: Good point! Done.

PR #264, opened 2023-12-23
omit default value for list in config

gyin94:


axlearn/common/config.py
```diff
@@ -387,6 +387,8 @@ def process_kv(key: str, val: Any):
                 default_val = field.default
                 if val is default_val and default_val in omit_default_values:
                     return
+            elif field is None and val in omit_default_values:
```
ruomingp, line 390: I wonder if we could apply this logic only to dataclass fields.

axlearn/experiments/testdata/axlearn.experiments.text.gpt.c4_trainer/fuji-7B-single.txt
```diff
@@ -223,7 +221,6 @@ model.decoder.transformer.layer.self_attention.attention.output_linear.param_par
 model.decoder.transformer.layer.self_attention.attention.output_linear.param_partition_spec[0][1]: 'fsdp'
 model.decoder.transformer.layer.self_attention.attention.output_linear.param_partition_spec[0][2]: 'seq'
 model.decoder.transformer.layer.self_attention.attention.output_linear.param_partition_spec[1]: 'model'
-model.decoder.transformer.layer.self_attention.attention.output_linear.param_partition_spec[2]: None
```
ruomingp, line 226: Applying it to list/tuple/dict seems going too far, e.g., here it will not be easy for readers to tell that `param_partition_spec` actually has three elements rather than two. WDYT?

ruomingp (commented):
Thanks!

axlearn/common/config.py
gyin94: changed

axlearn/experiments/testdata/axlearn.experiments.text.gpt.c4_trainer/fuji-7B-single.txt
gyin94: sounds good

axlearn/common/config.py
```diff
@@ -371,7 +371,14 @@ def to_flat_dict(self, *, omit_default_values: Collection[Any]) -> Dict[str, Any
         result = {}
 
         def enter(key: str, val: Any, default_result: Optional[List]) -> Optional[List]:
-            if key and isinstance(val, ConfigBase):
+            if dataclasses.is_dataclass(val) and not isinstance(val, type):
```
gyin94, line 374: this is similar to the condition check of _default_enter_fn. https://github.com/apple/axlearn/blob/4c9ecd92c8d1a411934a239763a778c2f80825e8/axlearn/common/config.py#L487

axlearn/common/config.py
```diff
@@ -371,7 +371,14 @@ def to_flat_dict(self, *, omit_default_values: Collection[Any]) -> Dict[str, Any
...
+                omitted_result = []
```
ruomingp, line 375:
```suggestion
                kvs_to_traverse = []
```

axlearn/common/config.py
```diff
@@ -371,7 +371,14 @@ def to_flat_dict(self, *, omit_default_values: Collection[Any]) -> Dict[str, Any
...
+                for cur_key, cur_val in default_result:
+                    if cur_val in omit_default_values:
+                        continue
```
ruomingp, line 378: We should omit only if cur_val is the default value of the dataclass field, e.g.,

```
@dataclass 
class foo:
    x: Optional[int] = None
    y: Optional[int] = -1
```

We should omit if `x is None` but not if `y is None`. This is to make it consistent with the attribute values below.

ruomingp (commented):
Thanks!

axlearn/common/config.py
gyin94: made a change and added a test like this

axlearn/common/config.py
```diff
@@ -371,7 +371,23 @@ def to_flat_dict(self, *, omit_default_values: Collection[Any]) -> Dict[str, Any
         result = {}
 
         def enter(key: str, val: Any, default_result: Optional[List]) -> Optional[List]:
-            if key and isinstance(val, ConfigBase):
+            if dataclasses.is_dataclass(val) and not isinstance(val, type):
+                kvs_to_traverse = []
+                default_result_dict = dict(default_result)
+                for field in dataclasses.fields(val):
```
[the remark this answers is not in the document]
gyin94: made a change to iterate default_result

axlearn/common/config.py
gyin94: changed

ruomingp (approved):

PR #274, opened 2024-01-05
Update dataflow script to be able to run Mac M1.

zhiyun:
This PR makes 2 changes:
1. disable running from docker on Mac M1, to avoid qemu core dump issue https://github.com/docker/for-mac/issues/5342.
2. remove default number_of_worker_harness_threads=1. When user submit job that requires this flag, it can be set by passing --dataflow_spec=number_of_worker_harness_threads=1.


yqwangustc (approved):

ruomingp:
Hi @zhiyun, in retrospect, the removal of `number_of_worker_harness_threads=1` was not a good choice---it costs the team hours to figure out the solution again. If we remove it from dataflow command, we should add it back to the specific script that needs it.

PR #280, opened 2024-01-12
Adds some tolerance to the check about the sum of resources.

ruomingp:
Do not raise ValueError. Instead, just log a warning.

### DIFF
--- a/axlearn/cloud/common/quota_test.py
+++ b/axlearn/cloud/common/quota_test.py
@@ -17,8 +17,8 @@ _mock_config = {
         "resource_type2": 8,
     },
     "project_resources": {
-        "team1": {"resource_type1": 0.3},
-        "team2": {"resource_type1": 0.7, "resource_type2": 1.0},
+        "team1": {"resource_type1": 0.30000001},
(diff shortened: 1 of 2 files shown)


axlearn/cloud/common/quota_test.py
```diff
@@ -17,8 +17,8 @@
         "resource_type2": 8,
     },
     "project_resources": {
-        "team1": {"resource_type1": 0.5},
-        "team2": {"resource_type1": 0.5, "resource_type2": 1.0},
+        "team1": {"resource_type1": 0.3},
```
apghml, line 20: Does this test fail without the change? I get:

```
> 4.8+11.2==16
True
> 12.8+11.2==24
True
> 0.3+0.7==1
True
```

Maybe it would make sense to use:
```
> 0.1 + 0.2 <= 0.3
False
```

ruomingp: Good point. Found a way to make a test. PTAL?

apghml (approved):
Thanks!

apghml (approved):

PR #293, opened 2024-01-25
Allow casting dtype while restoring checkpoints

gyin94:
- This allows casting to target dtype when restoring the checkpoint. It is similar to pytorch model with `.to(torch.float16)` or `.to(torch.bfloat16)`

### DIFF
--- a/axlearn/common/checkpointer.py
+++ b/axlearn/common/checkpointer.py
@@ -18,7 +18,7 @@ import threading
 import time
 from concurrent import futures
 from types import TracebackType
-from typing import Any, Dict, List, NamedTuple, Optional, Protocol, Tuple, Type, Union
+from typing import Any, Callable, Dict, List, NamedTuple, Optional, Protocol, Tuple, Type, Union
 
 import jax
 import jax.numpy as jnp
@@ -404,7 +404,7 @@ class TensorStoreStateStorage(StateStorage):
         ckpt_dir: str,
         validation: CheckpointValidationType = CheckpointValidationType.EXACT,
         concurrent_gb: int = 32,
-        dtype: Optional[jnp.dtype] = None,
+        dtype_cast_func: Optional[Callable[[jnp.dtype], jnp.dtype]] = None,
     ) -> NestedTensor:
         """Restore checkpoints in tensorstore format from a target directory.
 
@@ -414,8 +414,8 @@ class TensorStoreStateStorage(StateStorage):
             ckpt_dir (str): The path to checkpoint directory.
             validation (CheckpointValidationType): Validation type after loading weights.
             concurrent_gb (int): Max concurrent size to load for tensorstore checkpoints.
-            dtype (Optional[jnp.dtype]): Target dtype to cast all tensors.
-                By default, it is None and would infer from index file.
+            dtype_cast_func (Optional[Callable[[jnp.dtype], jnp.dtype]]):
+                Function to cast dtype to other dtype.
 
         Returns:
             NestedTensor: Restored model weights.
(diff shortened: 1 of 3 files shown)


axlearn/common/checkpointer.py
```diff
@@ -416,11 +431,19 @@ def restore_from_dir(
             spec.tf_ckpt_map, dir=os.path.join(ckpt_dir, f"tf_{jax.process_index()}")
         )
 
+        # Override dtype to target cast dtype.
+        # From jax docs,
+        # Cast while reloading on process to avoid 2 copies on device if the
+        # casting is done on device.
+        spec_dtypes = spec.dtypes
+        if dtype is not None:
+            spec_dtypes = [dtype] * len(spec_dtypes)
```
ruomingp, line 440: How would this handle integer vs float dtypes?

ruomingp (changes requested):

axlearn/common/checkpointer.py
gyin94: how about we add a more complicated check? And only add this for bfloat16 and float16 to float32 weights. wdyt?

ruomingp: How about taking an `Optional[Callable[[jnp.dtype], jnp.dtype]]` that maps one dtype to another?

gyin94: Sounds good. Let me try

axlearn/common/checkpointer.py
```diff
@@ -404,7 +404,22 @@ def restore_from_dir(
...
+        """Restore checkpoints in tensorstore format from a target directory.
+
+        Args:
+            step (int): Step number to restore.
+            state (Union[NestedTensor, NestedTensorSpec]): Model states.
+            ckpt_dir (str): The path to checkpoint directory.
+            validation (CheckpointValidationType): Validation type after loading weights.
+            concurrent_gb (int): Max concurrent size to load for tensorstore checkpoints.
+            dtype_cast_func (Optional[Callable[[jnp.dtype], jnp.dtype]]):
+                Function to cast dtype to other dtype.
```
ruomingp, line 418:
```suggestion
            dtype_cast_func (Optional[Callable[[jnp.dtype], jnp.dtype]]):
                Function to map saved dtypes to restored dtypes.
```

axlearn/common/checkpointer_test.py
```diff
@@ -559,17 +559,27 @@ def test_read_state_spec(self):
...
-    def test_save_and_restore_from_dir(self, restore_floats_as: jnp.dtype):
+    @parameterized.parameters(
+        [jnp.float32, None],
+        [jnp.bfloat16, None],
+        [jnp.int32, None],
+        [jnp.int16, None],
+        [jnp.float32, jnp.bfloat16],
+        [jnp.float32, jnp.float16],
+    )
+    def test_save_and_restore_from_dir(self, restore_floats_as: jnp.dtype, cast_dtype: jnp.dtype):
```
ruomingp, line 570: Comment on the difference between `restore_floats_as` and `cast_dtype`? Shouldn't they be the same?

axlearn/common/checkpointer.py
```diff
@@ -404,7 +404,22 @@ def restore_from_dir(
         ckpt_dir: str,
         validation: CheckpointValidationType = CheckpointValidationType.EXACT,
         concurrent_gb: int = 32,
+        dtype_cast_func: Optional[Callable[[jnp.dtype], jnp.dtype]] = None,
```
ruomingp, line 407: Reading the unittest again, I'm not sure we need this new arg. It seems that user can manipulate `state` by casting one type to another instead. Would that work?

gyin94: do you mean we can move it to TensorStoreStateStorageBuilder to manipulate the state? 

https://github.pie.apple.com/foundation-models/axlearn/blob/6f5d47153067bb3186b070f370b3e34db87c862b/axlearn/common/state_builder.py#L390

ruomingp: Maybe with a chain state builder? Let's discuss more offline. I'd like to understand your use case better.

gyin94: I checked the chain state builder. But I still this could be a native feature of inference runner when loading checkpoints. I made another change here as a reference here. https://github.com/apple/axlearn/pull/296. This might also explain why the inference is kind of slow and uses more memory since it has a data type cast for each iter.

PR #296, opened 2024-01-28
cast dtype before loading tensors in inference runner

gyin94:


axlearn/common/inference.py
```diff
@@ -370,21 +382,15 @@ def _inference_iter(
         **kwargs,
     ) -> Tuple[Tensor, NestedTensor, NestedTensor]:
         """Implements inference for a single input batch."""
-        cfg = self.config
         new_prng_key, iter_key = jax.random.split(prng_key)
 
-        def inference_cast(in_tree: NestedTensor) -> NestedTensor:
```
gyin94, line 376: instead of casting params for each inference run (also copies the tensor that needs more memory)

axlearn/common/inference.py
```diff
@@ -219,6 +219,9 @@ def __init__(self, cfg: Config, *, parent: Optional[Module]):
         with self.mesh():
             self._add_child("model", cfg.model)
             self._model_param_specs = self.model.create_parameter_specs_recursively()
+            if cfg.inference_dtype is not None:
+                self._model_param_specs = self._inference_cast(self._model_param_specs)
```
ruomingp, line 223:
```suggestion
            self._model_param_specs = self._inference_cast(self._model_param_specs)
```

axlearn/common/utils.py
```diff
@@ -830,9 +832,13 @@ def cast_floats(in_tree: NestedTensor, to_dtype: Optional[jnp.dtype]) -> NestedT
...
 
-    def cast(x: Tensor) -> Tensor:
+    def cast(x: Union[Tensor, TensorSpec]) -> [Tensor, TensorSpec]:
         if x.dtype == from_dtype:
-            return x.astype(to_dtype)
+            if isinstance(x, Tensor):
+                return x.astype(to_dtype)
+            else:
+                x.dtype = to_dtype
+                return x
```
ruomingp, line 841: To avoid mutating `x`:

```suggestion
                return dataclasses.replace(x, dtype=to_dtype)
```

gyin94: changed

axlearn/common/inference.py
gyin94: changed

ruomingp (approved):

PR #300, opened 2024-02-01
Use Peter Novig spell corrector for similar_names

wangkuiyi:
Guoli Yin pointed me to this code snippet for field name correction. I instantly thought of Peter Novig's spell checker, which is available at https://norvig.com/spell-check.html. This pull request makes advantage of Peter Novig's technique for name suggestions.

I'm preserving the original function name,'similar_names'. However, the new algorithm only suggests one name. I think this makes more sense than suggesting multiple choices because a class's collection of properties is typically far smaller than the English vocabulary, and users may prefer one accurate correction when they misspell.

Please let me know if I mis-interpret the problem here. Thanks!

### DIFF
--- a/axlearn/common/config.py
+++ b/axlearn/common/config.py
@@ -125,26 +125,40 @@ def is_attrs(x: Any):
 
 
 def similar_names(name: str, candidates: Iterable[str]) -> List[str]:
-    """Return a sorted list of candidates that are similar to name."""
+    """Use Peter Novig's spell correcter at https://norvig.com/spell-correct.html"""
+    word_count = Counter([_ for _ in candidates])
 
-    def overlaps(name: str, key: str) -> float:
-        """The fraction of 3-char substrings in <name> that appear in key."""
-        matches = 0
-        trials = 0
-        for i in range(len(name) - 2):
-            trials += 1
-            if name[i : i + 3] in key:
-                matches += 1
-        return float(matches) / max(trials, 1)
+    def P(word, N=sum(word_count.values())):
+        "Probability of `word`."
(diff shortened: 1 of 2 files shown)


axlearn/common/config.py
```diff
@@ -125,26 +125,40 @@ class Foo:
...
-    # Filter out candidates below 0.5 overlap threshold.
-    pairs = [pair for pair in pairs if pair[0] > 0.5]
-    # Sort by highest overlap, breaking ties alphabetically.
-    pairs.sort(key=lambda pair: (-pair[0], pair[1]))
-    # Return just the keys.
-    return [key for _, key in pairs]
+    """Use Peter Novig's spell correcter at https://norvig.com/spell-correct.html"""
+    word_count = Counter([_ for _ in candidates])
+
+    def P(word, N=sum(word_count.values())):
```
gyin94, line 131: consider a type hint and return type?

wangkuiyi: I agree that we should add the type hints if we merge this pull request; however, after more thoughts, I am no longer  sure that we should merge Peter's algorithm.

This algorithm expands the misspelled word to a small set of similar words, and filter out those not in the vocabulary. This is because the vocabulary is too large to compute the editing distance between each word in it and the misspelled word.

However, a Python class wouldn't have this many candidate symbols as in a vocabulary. So it may be affordable to compute the pair-wise editing distance. What do you think?

gyin94: makes sense to me.

ruomingp:
Thanks, Yi. Let me close this PR.

PR #340, opened 2024-02-25
Requires flags to be parsed before `gcp_settings` is called.

ruomingp:
Motivation: the axlearn CLI sometimes does not respect the --zone flag. This often happens when gcp_settings is called before the --zone flag is parsed.

Since the return value of gcp_settings depends on flag values, it is more principled if we require the flag values to be parsed before gcp_settings can be called.

This PR fixes two common patterns of gcp_settings usage to set flag default values and config field default values. Instead, we set the default to None and call gcp_settings in from_flags or from_spec.

### DIFF
--- a/axlearn/cloud/common/bundler.py
+++ b/axlearn/cloud/common/bundler.py
@@ -199,7 +199,7 @@ class Bundler(Configurable):
         return temp_dir
 
     @classmethod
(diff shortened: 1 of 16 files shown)


markblee (approved):
Thanks for addressing this!

PR #346, opened 2024-02-28
Add option to use whitening or not in ImagenetInput

weiliu89:


axlearn/vision/input_image.py
```diff
@@ -335,6 +336,7 @@ def example_fn(example: Dict[str, Tensor]) -> Dict[str, Tensor]:
             randaug_magnitude=randaug_magnitude,
             randaug_exclude_ops=randaug_exclude_ops,
             erasing_probability=erasing_probability,
+            use_whitening=use_whitening,
```
markblee, line 339: Not for this PR, but a more 'configurable' way would be to take a `config_for_function(crop_augment_whiten)` as an input processor, configure it at the caller, and chain it here. This avoids the need to propagate all args in the fn signature.

markblee (approved):

xianzhidu (approved):
Thanks!

PR #349, opened 2024-03-01
Fix reading bastion flag.

markblee:


axlearn/cloud/gcp/jobs/launch.py
```diff
@@ -214,7 +214,7 @@ def from_flags(cls, fv: flags.FlagValues, action: str, **kwargs) -> Config:
         """
         cfg = super().from_flags(fv, **kwargs)
         if not cfg.bastion_name:
-            cfg.bastion_name = shared_bastion_name(fv)
+            cfg.bastion_name = fv.bastion or shared_bastion_name(fv)
```
zetaqubit, line 217: For future: might be good to centralize all of the config reading and overriding

zetaqubit (approved):

ruomingp (approved):
How about the other call sites of `shared_bastion_name`?

markblee:
> How about the other call sites of `shared_bastion_name`?

This is the only callsite.

PR #350, opened 2024-03-02
Makes ResourceLimitCalculator support proportional quotas.

ruomingp:


markblee (approved):

axlearn/cloud/common/scheduler.py
```diff
@@ -146,76 +146,96 @@ def calculate(
...
+                Available resources will be allocated proportional to quotas.
             demands: A mapping from project ids to demands. If a project id is missing, assume
                 demand of 0.
 
         Returns:
-            A mapping from project ids to resource limits.
+            A mapping from project ids to resource limits for each project id in `demands`.
 
         Raises:
-            ValueError: if total quota exceeds `limit`.
```
apghml, line 160: Would it make sense to add a test that it behaves correctly when the total quota exceeds `limit`, if there isn't already such a test?

apghml (approved):

axlearn/cloud/common/scheduler.py
markblee: I think the idea is that these quota values represent proportions, so not necessarily comparable to limit (see e.g L228)?

ruomingp: We have a test to show that quotas are proportional and do not have to add up to limit. We don't currently have a test to show that the sum of quotas can exceed the limit, but I can add one in a follow-up PR.

Meanwhile, let me merge this PR since we need it to simplify the process to add new quota groups.

PR #361, opened 2024-03-09
Print step time for each step

samos123:
This is helpful in cases where there is variable step time and looking at the logs would quickly allow you to identify such cases.

### DIFF
--- a/axlearn/common/trainer.py
+++ b/axlearn/common/trainer.py
@@ -141,6 +141,9 @@ class SpmdTrainer(Module):
         # By default, only trace on host 0.
         start_trace_process_indices: Union[Literal["all"], Sequence[int]] = [0]
 
+        # Log the step time for each individual step
+        log_step_time: bool = False
+
         # Prune empty state updates.
         prune_empty_state_updates: bool = True
 
@@ -432,7 +435,9 @@ class SpmdTrainer(Module):
 
                     self._step = self._step + 1
 
-                    step_start_time = time.perf_counter()
+                    if cfg.log_step_time:
+                        step_start_time = time.perf_counter()
+
                     self.vlog(3, "Start step %s", self.step)
                     output = self._run_step(
                         utils.host_to_global_device_array(input_batch),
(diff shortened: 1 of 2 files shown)


markblee (commented):
Hm, I'm not sure that we want to print every step by default (since it'll introduce a significant amount of logs/summaries, and since avg step time is usually pretty stable after the first couple hundred steps)?

samos123:
That makes sense. Should I hide it behind an option? For me it was important to troubleshoot a variable step time issue. Or would you rather totally leave this out of the code base. Note I'm fine with that too.

samos123:
I've added it as a config parameter and made it false by default

samos123:
I have been using this for 2 use cases:
* Ensure step time is stable across steps. In the past on the GPU the networking becomes unstable where collectives sometimes take longer between steps. E.g. step time varies between 5 second to 7 seconds.
* Quickly evaluate whether there is a performance gain without having to wait for 100 steps

Please let me know if there is any interest in merging it.

axlearn/common/trainer.py
```diff
@@ -141,6 +141,9 @@ class Config(Module.Config):
         # By default, only trace on host 0.
         start_trace_process_indices: Union[Literal["all"], Sequence[int]] = [0]
 
+        # Log the step time for each individual step
+        log_step_time: bool = False
```
markblee, line 145: Maybe it'll be more general to do:

```suggestion
        # Interval to log average step time. If None, defaults to 100.
        log_step_time_every_n_steps: Optional[int] = None
```

Note that we default to None instead of 100 to avoid triggering a large number of golden config updates.

axlearn/common/trainer.py
```diff
@@ -439,6 +446,12 @@ def run(
...
                     )
                     self.vlog(3, "Done step %s", self.step)
+
+                    if cfg.log_step_time:
+                        step_time = time.perf_counter() - step_start_time
+                        self._step_log("Step %s took %s seconds", self.step, step_time)
+                        self.summary_writer(self.step, {"step_time": step_time})
+
                     num_steps += 1
                     if num_steps % 100 == 0:
```
markblee, line 456:
```suggestion
                    if num_steps % (cfg.log_step_time_every_n_steps or 100) == 0:
```

ruomingp (commented):
Will defer to @markblee for review.

changlan:
Closing this PR due to inactivity. Feel free to reopen if you would like to continue the work.

PR #362, opened 2024-03-09
increase timeout from 300s to 900s

samos123:
The scale up of GPU nodes for large jobs can take up to 10 minutes. So increasing this timeout is needed to wait for all pods to become available.

This change is required to run on 64 nodes with autoscaling enabled. The image pull and VM scale up time are the main reason it takes up to 10 minutes. In some cases the coordinator launches directly while other pods require scale up. In those cases, it's critical that the timeout is set to at least 10 minutes or more.


samos123:
Closing because Mark added this in a separate PR

PR #373, opened 2024-03-17
Makes adastar_optimizer compute `raw_update_norm` …

ruomingp:
…even when raw_update_clipping_threshold is None.


axlearn/common/optimizers.py
```diff
@@ -1502,7 +1506,7 @@ def adastar_optimizer(
         eps: (float) regularization constant added to the square root of smoothed_gradient_squares.
         eps_square: (float) regularization constant added to gradient_squares.
         raw_update_clipping_threshold: If not None, clips the norms of the raw updates
-            to this value.
+            to this value. `raw_update_norm` summaries will be clipped either way.
```
tgunter, line 1509: Did you mean:
```suggestion
            to this value. `raw_update_norm` summaries will be logged either way.
```
?

ruomingp: Indeed. Thanks!

markblee (approved):

PR #383, opened 2024-03-25
Add more Adastar optimizer stats

taolei87:
- Add smoothed update_norm and raw_grad_norm stats to the Adastar optimizer
- Add correlation stats to the Adastar optimizer


ruomingp (approved):
Let's remember to change the training code to verbosity.

PR #384, opened 2024-03-25
Dataflow changes

jiya-zhang:
I changed a few things in this PR:

1. Add Dockerfile entrypoint for Dataflow (this is needed for Dataflow worker to start up successfully)
2. Fix stdout and stderr in subprocess.popen (it used to always yield None to stdout and stderr)
3. Mount gcloud config folder to docker container (so that container will have the credentials to access GCS, etc.)
4. Add `dataflow_word_count.sh` script. This is what I used to test a simple job running on GCP.


Dockerfile
```diff
@@ -73,6 +73,10 @@ ENV RUN_PYTHON_SDK_IN_DEFAULT_ENVIRONMENT=1
 RUN pip install .[gcp,dataflow]
 COPY . .
 
+COPY --from=apache/beam_python3.9_sdk:2.52.0 /opt/apache/beam /opt/apache/beam
+ENTRYPOINT ["/opt/apache/beam/boot"]
```
markblee, line 77: Can you comment on the behavior observed without this change?

axlearn/cloud/gcp/jobs/dataflow.py
```diff
@@ -243,13 +243,13 @@ def _execute(self):
...
             cmd = (
-                "docker run --rm --entrypoint /bin/bash "
+                "docker run --rm --mount type=bind,src=$HOME/.config/gcloud,dst=/root/.config/gcloud --entrypoint /bin/bash "
                 f"{self._bundler.id(cfg.name)} -c '{cfg.command}'"
             )
         cmd = f"{cfg.setup_command} && {cmd}"
         cmd = f"bash -c {shlex.quote(cmd)}"
         logging.info("Executing in subprocess: %s", cmd)
-        with subprocess.Popen(cmd, shell=True, text=True) as proc:
+        with subprocess.Popen(cmd, shell=True, text=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE) as proc:
```
markblee, line 252: Does this mean the user will not be able to see the outputs in realtime?

dataflow_word_count.sh
```diff
@@ -0,0 +1,16 @@
+# chmod +x dataflow_word_count.sh
```
markblee, line 1: Let's remove this file (I think the usage is in docstring)? Long-term we can add some dedicated docs on running dataflow.

markblee (commented):
Thanks! Some initial comments/questions.

Dockerfile
jiya-zhang: Yes. Before this change, the Dataflow worker will not start up properly and the job appears to be hung for a long time without useful error messages. Product team suggested this change based on documentation on using a custom container: https://cloud.google.com/dataflow/docs/guides/build-container-image#use_a_custom_base_image

axlearn/cloud/gcp/jobs/dataflow.py
jiya-zhang: Before the change, I observed that stdout and stderr are always `None`. This is the documentation that suggested the change: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate

`Note that if you want to send data to the process’s stdin, you need to create the Popen object with stdin=PIPE. Similarly, to get anything other than None in the result tuple, you need to give stdout=PIPE and/or stderr=PIPE too.`

However, for both before and after the change, users can't seem to see the outputs in realtime.

jiya-zhang: Okay I tested this again and since after the change user can't see real time output, I reverted this change

markblee (commented):
Thanks! A few additional suggestions while testing this PR:
1. On L175, let's change to `cfg.command = f"{cfg.command.strip()} {dataflow_flags}"` so that trailing newlines in the command do not cause `dataflow_flags` to be interpreted as a separate command.
2. L249, similar change to `cmd = f"{cfg.setup_command.strip()} && {cmd}"`.
3. You may want to rebase onto latest `main` (which fixes the bundler specs) for local testing.

markblee (approved):
Thanks, also update the PR description?

markblee (approved):

jiya-zhang:
@markblee Do you mind trigger the CI tests again? The error should disappear now that dependency issue was fixed and merged. (I don't have permission to rerun CI tests)

markblee:
> @markblee Do you mind trigger the CI tests again? The error should disappear now that dependency issue was fixed and merged. (I don't have permission to rerun CI tests)

It failed again with the same issue. Did you rebase?

markblee (approved):

PR #404, opened 2024-04-16
Add skip_bundle to reuse images.

markblee:


axlearn/cloud/common/bundler.py
```diff
@@ -306,6 +308,7 @@ def from_spec(cls, spec: List[str], *, fv: Optional[flags.FlagValues]) -> Config
         - platform: The image target platform.
         - allow_dirty: Whether to ignore dirty git status.
         - cache_from: A comma-separated list of cache sources.
+        - skip_bundle: Whether to skip the build + push.
```
ruomingp, line 311: When is it safe to enable `skip_bundle`? Please add a comment.

markblee: Added some clarification: 
> This option is intended to be used when an image has already been pre-built offline, in which case we may still want to leverage the install commands implemented by the bundler.

E.g. when launching training we can specify `--bundler_spec=skip_bundle=True`.

ruomingp (approved):

PR #406, opened 2024-04-18
Jax upgrade

kelvin-zou:
Upgrade Jax to 0.4.25, and fix the broken unit tests.


pull_request_template.md
```diff
@@ -0,0 +1 @@
+Please read through the [Contribution Guidelines](https://quip-apple.com/R3XmAn4kGnT4) before submitting a pull request.
```
samos123, line 1: I don't think this is supposed to be here

markblee (approved):

PR #408, opened 2024-04-18
make mel energy floor value required

yqwangustc:
This is the first step towards using a small mel energy value (such as 1e-6). We first make this value required. Once every experiment using a smaller value we can set a default value like 1e-6.


markblee (approved):

ruomingp (approved):
Can you prepare an ajax PR and make sure that it passes CI before merging this one?

yqwangustc:
> Can you prepare an ajax PR and make sure that it passes CI before merging this one?

An internal PR is prepared (#9767) and it shows that all the tests have passed after patching that PR.

PR #411, opened 2024-04-19
Replace Matmul in Linear Layer with Variable Shape Einsum

jiarui-lu2:


ruomingp (commented):
As discussed offline, let's run one more experiment to validate this change.

ruomingp (approved):

PR #416, opened 2024-04-23
Initial GKE runner and launcher.

markblee:


axlearn/cloud/gcp/job.py
```diff
@@ -75,16 +81,46 @@ def _get_job_credentials(
         )
 
 
-class TPUJob(GCPJob):
+@config_class
+class AcceleratorConfig(ConfigBase):
+    """Configures job resources, e.g. TPU or GPU.
+
+    Attributes:
+        instance_type: Instance type, e.g. tpu-v4-8.
```
samos123, line 89: add example for what this should be on GPU (can be AWS or GCP)

markblee: We can follow the format used in the mesh rules, so e.g. https://github.com/apple/axlearn/blob/9da0b26064563a6f783f9f0d32ad0594f8dfad87/axlearn/experiments/text/gpt/fuji.py#L69

zhzhyi (approved):

ruomingp (approved):

PR #417, opened 2024-04-23
fix absl-py version

zhiyun:


pyproject.toml
```diff
@@ -13,7 +13,7 @@ requires-python = ">=3.9"
 # Every time we upgrade JAX, we should try to bring the rest to the newest versions.
 dependencies = [
     "attrs>=23.1.0", # We use `type` in `attrs.field`
-    "absl-py",
+    "absl-py<2",  # breaks axlearn.cli.utils_test on 2.1.0
```
ruomingp, line 16: Can we pin to a specific version of absl-py?

zhiyun: Done.

kelvin-zou (approved):

PR #423, opened 2024-04-26
allow _map_label_sequences to set `remove_repeats=False`

zhiyun:
to include RNN-T use case.

### DIFF
--- a/axlearn/audio/asr_decoder.py
+++ b/axlearn/audio/asr_decoder.py
@@ -439,11 +439,18 @@ class CTCDecoderModel(BaseModel):
         )
 
 
-def _map_label_sequences(inputs: Tensor, *, blank_id: int = 0, pad_id: int = 0) -> Nested[Tensor]:
-    """Removes blanks, paddings, and repeats from the input sequences, as seen in CTC.
+def _map_label_sequences(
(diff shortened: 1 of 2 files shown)


axlearn/audio/asr_decoder.py
```diff
@@ -439,11 +439,18 @@ def _postprocess_outputs(self, *, sequences: Tensor, paddings: Tensor, scores: T
         )
 
 
-def _map_label_sequences(inputs: Tensor, *, blank_id: int = 0, pad_id: int = 0) -> Nested[Tensor]:
-    """Removes blanks, paddings, and repeats from the input sequences, as seen in CTC.
+def _map_label_sequences(
+    inputs: Tensor, *, remove_repeats: bool = True, blank_id: int = 0, pad_id: int = 0
```
ruomingp, line 443: Should we leave `remove_repeats` without a default value, since neither is the best value?

```suggestion
    inputs: Tensor, *, remove_repeats: bool, blank_id: int = 0, pad_id: int = 0
```

zhiyun: Done.

markblee (approved):

PR #428, opened 2024-04-28
Adds versions V1,V2,V3 for the Fuji models.

ruomingp:


axlearn/experiments/text/gpt/fuji.py
```diff
@@ -6,12 +6,13 @@
 
 The fuji models are set up to imitate LLaMA-1 (https://arxiv.org/abs/2302.13971).
```
kelvinzou, line 7: Nit, update this comment?

axlearn/experiments/text/gpt/common.py
```diff
@@ -444,19 +444,23 @@ def evaler_config_dict(
...
         arch: The architecture of the model.
         model_size: The number of transformer parameters (not including vocab embeddings).
+        version: An optional version string.
 
     Returns:
-        f"{arch}-{model_size}".
+        f"{arch}-{model_size}" or f"{arch}-{model_size}-{version}".
     """
-    return f"{arch}-{model_size}"
+    name = f"{arch}-{model_size}"
```
kelvinzou, line 460: Nit, should we make version = v1 as an default for backward compatibility? If looking at your other change, maybe make it ?
```
    if version:
        name += f"-{version}"
    else:
         name += f"-v1"
```

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -54,17 +105,21 @@ def get_trainer_kwargs(model_size: str, *, vocab_size: int) -> Dict[str, Any]:
...
+            train_batch_size=train_batch_size,
+            max_step=max_step,
             mesh_shape=mesh_shape_from_axes(fsdp=-1),
             mesh_rules=(
                 # tpu-v4. step time: 3.03s.
                 ("tpu-v4-(1024|2048)", mesh_shape_from_axes(data=-1, fsdp=16)),
                 # tpu-v5e. step time: TBD.
                 ("tpu-v5litepod-256", mesh_shape_from_axes(data=-1, fsdp=16)),
                 # H100/A100 80G. Maximum per-node batch size = 64, hence need >= 32 nodes.
+                # p5.48xlarge 8x64. v1 step time: 1.54s.
```
kelvinzou, line 122: Nit, maybe add a comment stating that global batch size = 512 for v3 model and need to adjust the mesh size? 
(1024 GPUs won't work here)

kelvinzou (approved):

axlearn/experiments/text/gpt/fuji.py
kelvin-zou, line 7: nit, fix comment?

axlearn/experiments/text/gpt/common.py
kelvin-zou, line 460: nit: add v1 as an default for backward compatibility?

axlearn/experiments/text/gpt/fuji.py
kelvin-zou, line 122: Nit, add a note for v3 model? I believe 1024 GPUs won't work for v3 model since global bs is only 512 due to 8k seq length.

kelvin-zou (approved):
Thank you for making the change!

ruomingp:
Thanks for the review. Will address the comments in a follow-up PR.

PR #437, opened 2024-05-02
Make GKE jobset rescheduling more robust.

markblee:


ruomingp (approved):

ruomingp:
What's the status of this PR? Do we want to merge it?

markblee:
> What's the status of this PR? Do we want to merge it?

Thanks for checking -- I held off because of the unintended scheduling changes discussed internally. Once I validate that this works with the updated tpu provisioner I will merge.

PR #444, opened 2024-05-04
Improves bastion log sync logic

ruomingp:
- Adds TMPDIR to avoid `No space left on device`.
- Makes tpu_runner.py start log sync before installing the bundle. 
- Pipes bundle installation stderr and stdout to run.log. This allows us to see the bundle installation process in the GCS logs.

### DIFF
--- a/axlearn/cloud/gcp/jobs/bastion_vm.py
+++ b/axlearn/cloud/gcp/jobs/bastion_vm.py
@@ -69,7 +69,7 @@ Examples:
 To test changes to bastion:
 
     # 1. Build a custom image.
-    axlearn gcp bundle --bundler_type=docker \
+    axlearn gcp bundle --bundler_type=artifactregistry \
         --name=$USER-bastion \
         --bundler_spec=image=base \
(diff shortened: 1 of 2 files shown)


axlearn/cloud/gcp/jobs/bastion_vm.py
```diff
@@ -249,6 +249,9 @@ def _gcloud_storage_rsync(
             timeout=timeout_s,
             capture_output=True,
             text=True,
+            # Avoid "No space left on device":
+            # https://cloud.google.com/knowledge/kb/error-message-while-running-the-command-gsutil-rsync-000004577
+            env={"TMPDIR": f"{_LOG_DIR}/rsync"},
```
markblee, line 254: nit -- I wonder if we should use /var/tmp/rsync explicitly, given that log dir is often the directory being rsync'ed itself?

ruomingp: Good point. Done.

markblee (approved):

PR #446, opened 2024-05-07
Adds BaseLayer.Config.tensor_stats

ruomingp:
The `tensor_stats` abstraction allows users to compute stats (e.g., mean, RMS norm, max abs) on tensors such as layer inputs/outputs and add them to summaries.

The abstraction decouple which tensors to collect stats on, which will be controlled by the layer implementation via `_add_tensor_stats(name, value)` calls, vs. how to compute and report the stats, which will be controlled by `Config.tensor_stats` and configured on a per-experiment basis.

Inspired by the discussion between @taolei87 and @apghml on an internal PR.

### DIFF
--- a/axlearn/common/base_layer.py
+++ b/axlearn/common/base_layer.py
@@ -16,6 +16,7 @@ from axlearn.common.metrics import WeightedScalar
 from axlearn.common.module import Module, child_context, current_context, new_output_collection
 from axlearn.common.param_init import DefaultInitializer, FanAxes
 from axlearn.common.utils import (
+    Nested,
     NestedTensor,
     PartitionSpec,
     Tensor,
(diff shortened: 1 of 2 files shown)


axlearn/common/base_layer.py
```diff
@@ -296,6 +365,8 @@ def initialize_parameters_recursively(
                     parameter_spec=spec,
                 )
         for name, child in self._children.items():
+            if not isinstance(child, BaseLayer):
+                continue
```
apghml, line 369: Could you add a comment about why this is needed and a test that fails without this change?

apghml (commented):
Thanks!

axlearn/common/base_layer.py
```diff
@@ -126,6 +127,61 @@ def apply(self, prng_key: Tensor, params: NestedTensor) -> NestedTensor:
...
+class TensorStats(Module):
+    """An abstract Module to add summaries about the given Tensors."""
+
+    def add_stats(self, name: str, value: Nested[Tensor]):
+        """Subclasses must implement this method."""
+        raise NotImplementedError(type(self))
+
+
+class CompositeTensorStats(TensorStats):
+    """A TensorStats consists of multiple child TensorStats."""
```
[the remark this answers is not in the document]
ruomingp: Done.

axlearn/common/base_layer.py
ruomingp: Added comment. The tensor state unittest would fail without this `if`.

apghml (approved):

PR #450, opened 2024-05-08
Add new model config for smaller tests

jesus-orozco:
Adding a new model configuration for text experiments. The goal is to get an early termination model for fuji-test to accelerate infrastructure validation.

+ @jiya-zhang


jiya-zhang:
@jesus-orozco is still working on this PR, but it would be helpful to get some early feedback from @markblee - Thanks!

axlearn/experiments/text/gpt/c4_trainer.py
```diff
@@ -149,4 +152,50 @@ def make_single_host_config(base_config_name: str) -> SpmdTrainer.Config:
...
+                        evaler.set(
+                            eval_policy=config_for_function(eval_every_n_steps_policy).set(
+                                n=eval_every_n_steps
+                            )
+                        )
+                    cfg.max_step = max_step
+                    cfg.mesh_shape = mesh_shape_from_axes(data=-1, fsdp=4)
+                    cfg.summary_writer.write_every_n_steps = eval_every_n_steps
+                    cfg.checkpointer.save_policy = config_for_function(every_n_steps_policy).set(
+                        n=eval_every_n_steps
```
jiya-zhang, line 193: Is it possible to save checkpointer more frequently than eval? Something like save ckpt every 500 steps, eval every 1500 steps. This allows us to identify issues separately if the job hangs

jesus-orozco: Thanks for the feedback! Added the custom policy to save checkpoints more often to differentiate from eval steps

axlearn/experiments/text/gpt/c4_trainer.py
```diff
@@ -149,4 +152,51 @@ def make_single_host_config(base_config_name: str) -> SpmdTrainer.Config:
...
+                            eval_policy=config_for_function(eval_every_n_steps_policy).set(
+                                n=eval_every_n_steps
+                            )
+                        )
+                    cfg.max_step = max_step
+                    cfg.mesh_shape = mesh_shape_from_axes(data=-1, fsdp=4)
+                    cfg.summary_writer.write_every_n_steps = eval_every_n_steps
+                    cfg.checkpointer.save_policy = config_for_function(every_n_steps_policy).set(
+                        n=save_every_n_steps
+                    )
```
markblee, line 195: For these, have you considered tweaking the kwargs here directly?
https://github.com/apple/axlearn/blob/c2c8a935a8ea339cdf0e0ffad6d48e005455dbe4/axlearn/experiments/text/gpt/fuji.py#L85-L104

It looks like we can add a mesh rule for the accelerators that you are testing on, too.
For reference, the kwargs will be passed to https://github.com/apple/axlearn/blob/c2c8a935a8ea339cdf0e0ffad6d48e005455dbe4/axlearn/experiments/text/gpt/common.py#L466-L482.

markblee (commented):
Thanks @jesus-orozco !

axlearn/experiments/text/gpt/c4_trainer.py
jesus-orozco: Thanks for the inputs Mark! 
For the initial pass I was basing the new config on the existing one-off configuration, but it does make more sense to have dedicated kwargs for the new model config as this would allow us to more easily create variations to test multiple TPU shapes reusing get_trainer_config_fn().
@jiya-zhang @markblee PTAL at the updated kwargs to double-check it makes sense.

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -140,6 +140,29 @@ def get_trainer_kwargs(model_size: str, *, vocab_size: int, version: Version) ->
                 ),
             ),
         )
+    elif model_size == "simple":
```
markblee, line 143: Thanks! Does this need to be separate from `"test"` (which is itself intended to be the testing configuration)? 

In particular, we can configure `mesh_rules` for the accelerator that you are testing on. This way, it'll run on both CPU and the target testing hardware.

The only other differences seem to be batch sizes and eval/saving more frequently, which seem tolerable as defaults. WDYT?

jiya-zhang: Re mesh_rules, yeah sometimes we test on v4-8, and we need something like (-1,1,4,1,1). 

However, re eval/saving/max step, we do want to have a config that terminate training early. As long as the training runs for a few thousands steps without problems, then we know the jax testing passes

jesus-orozco: Thanks! That sounds reasonable, adding the new defaults to the "test" configuration instead for frequent saving/early termination. 
On mesh rules, I'll leave the default for it to work on CPU, but can you clarify how we can configure the rules for specific accelerators? as Maggie mentioned, we'd be testing mainly on smaller TPU shapes like v4-8.

jesus-orozco: Thanks @markblee !
Committed the changes you suggested, it makes sense to add data=-1 to the default configuration.

markblee (commented):
FYI, you might need to run golden config updates: https://github.com/apple/axlearn/blob/main/docs/01-start.md#testing

jesus-orozco:
> FYI, you might need to run golden config updates: https://github.com/apple/axlearn/blob/main/docs/01-start.md#testing
Thanks!
Ran golden config, tests should pass now with available test data.

markblee:
> > FYI, you might need to run golden config updates: https://github.com/apple/axlearn/blob/main/docs/01-start.md#testing
> > Thanks!
> > Ran golden config, tests should pass now with available test data.

Thanks, did you commit the changes? I don't see it in the PR diff.

jesus-orozco:
> > > FYI, you might need to run golden config updates: https://github.com/apple/axlearn/blob/main/docs/01-start.md#testing
> > > Thanks!
> > > Ran golden config, tests should pass now with available test data.
> 
> Thanks, did you commit the changes? I don't see it in the PR diff.

My bad, typo on the branch push, fixed now!

markblee (approved):
Thanks!

PR #454, opened 2024-05-09
Triton flash attention

kelvin-zou:
This PR contains 2 changes:
1. Fix the NaN issue on flash attention
2. Add a segment ID (2d-bias) support in flash attention.

Also contains updated tests, and refactoring the existing code.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,8 +109,8 @@ dataflow = [
     "google-apitools",  # for beam pipeline
     "orjson==3.9.10",
 ]
-# Triton kernel dependency.
-triton = [
+# GPU custom kernel dependency.
+gpu = [
     "triton==2.1.0",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,11 +109,9 @@ dataflow = [
     "google-apitools",  # for beam pipeline
     "orjson==3.9.10",
 ]
-
+# Triton kernel dependency.
 triton = [
-    "matplotlib",
     "triton==2.1.0",
-    "pytest",
 ]
 
 [tool.flit.module]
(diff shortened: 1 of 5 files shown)


samos123:
How would I enable it in fuji 7b? Could you add a config for fuji 7b on GPU that uses flash attention?

pyproject.toml
```diff
@@ -109,6 +109,10 @@ dataflow = [
     "google-apitools",  # for beam pipeline
     "orjson==3.9.10",
 ]
+# Triton kernel dependency.
+triton = [
```
samos123, line 113: maybe rename `triton` to `gpu` if this is a gpu specific dependency? Not sure if there will be other gpu only dependencies

markblee (approved):
Thanks, can we internally link the tb with the comparison between flash/no-flash?

markblee:
> How would I enable it in fuji 7b? Could you add a config for fuji 7b on GPU that uses flash attention?

The [FlashAttention](https://github.com/apple/axlearn/blob/ef4c3820e92b0d4d83db23554eed2fdacf9d5f72/axlearn/common/flash_attention/layer.py#L37) layer should be (mostly) a drop-in replacement for `MultiheadAttention`/`GroupedQueryAttention` layers, with caveats mentioned in the class docstring. 

You can follow a similar idea as [set_norm_recursively](https://github.com/apple/axlearn/blob/ef4c3820e92b0d4d83db23554eed2fdacf9d5f72/axlearn/common/layers.py#L1593) to replace instances of `MultiheadAttention.Config` with your flash attention config.

kelvin-zou:
> Thanks, can we internally link the tb with the comparison between flash/no-flash?

Done, please check internal PR.

markblee (approved):

PR #456, opened 2024-05-10
Bug fix. Update layers to use _add_tensor_stats() 

taolei87:


axlearn/common/attention.py
```diff
@@ -2358,7 +2361,10 @@ class Config(BaseLayer.Config):
         add_dead_neuron_summary: Optional[bool] = None
 
         # Adds summary of RMS norms of the specified values. Supported value are:
+        # - "inputs": inputs of the layer.
+        # - "linear1_outputs": outputs of linear1.
         # - "linear2_outputs": outputs of linear2.
+        # TODO: deprecate this feature since we use TensorStats.
```
markblee, line 2367:
```suggestion
        # TODO(tlei3): deprecate this feature since we use TensorStats.
```
here and elsewhere?

markblee (approved):

axlearn/common/attention.py
taolei87: Done.

axlearn/common/base_layer.py
```diff
@@ -182,6 +182,18 @@ def add_stats(self, name: str, value: Nested[Tensor]):
...
+class DefaultTensorStats(CompositeTensorStats):
+    """Default tensor stats that compute RMS norm and max value."""
+
+    @config_class
+    class Config(CompositeTensorStats.Config):
+        tensor_stats: Dict[str, TensorStats.Config] = {
+            "norm": TensorRMSNorm.default_config(),
+            "max": TensorMaxAbs.default_config(),
+        }
+        inline_child_summaries: bool = True
```
ruomingp, line 194: Nit: do we need class? Maybe a function is enough:

```
def default_tensor_stats_config() -> TensorStats.Config:
```

ruomingp (approved):
Thanks!

PR #461, opened 2024-05-13
Perf optimize with unroll=8

kelvin-zou:
Enable unroll for GPU for better communication interleaving


kelvin-zou:
Close it due to recent upgrade in XLA, no longer needed.

PR #465, opened 2024-05-13
Gradient Accumulation in Axlearn

apoorvtintin:
Gradient accumulation allows training with higher batch sizes without scaling out.

Added a new learner type ```learner.klass: 'axlearn.common.learner.AccumulatedLearner'```

At a high level the optimization does the following:
1. Input batch is split into even microbatches.
2. Creates a buffer for gradients and metrics.
3. Runs forward and backward pass for each microbatch in a loop summing up the gradients and aggregating metrics.
4. Average gradients across microbatches and normalize metrics.

Configuration changes:
- Number of microbatches are specified during configuration through option ```micriobatches``` in the learner.

### DIFF
--- a/axlearn/common/learner.py
+++ b/axlearn/common/learner.py
@@ -6,14 +6,13 @@
 - Applying updates on non-differentiable params such as batch norm stats;
 - Maintaining Polyak averages of model params (if enabled).
 """
-
 import dataclasses
 import enum
-import logging
 from typing import Callable, Dict, Mapping, NamedTuple, Optional, Protocol, Sequence, Tuple
 
 import jax
 import optax
+from absl import logging
 from jax import numpy as jnp
 
 from axlearn.common.base_layer import NestedParameterSpec
@@ -487,9 +486,10 @@ class AccumulatedLearner(Learner):
 
         # The number of microbatches to accumulate per optimizer step.
         microbatches: Required[int] = REQUIRED
-        # tuple of key-value pairs specifying custom aggregation and normalization
-        # for a specific metric
+        # Tuple of key-value pairs specifying custom aggregation and normalization
+        # for a specific metric.
         metrics_accumulation_key_ops: Sequence[Dict[str, Optional[MetricsAccumulationOp]]] = []
+        # Data type used for microbatch gradient buffer and accumulation.
         gradient_dtype: Optional[jnp.dtype] = jnp.bfloat16
 
     def forward_and_backward(
@@ -508,6 +508,11 @@ class AccumulatedLearner(Learner):
         """
         should_compute_gradients = self.should_update_with_optimizers(opt_params)
         model_params = jax.tree_util.tree_map(lambda opt_param: opt_param.value, opt_params)
+
+        # create evenly sized accumulation microbatches, keep sequence dimension as it is.
+        inputs = jax.tree_map(
+            lambda x: x.reshape(self.config.microbatches, -1, *x.shape[1:]), inputs
+        )
 
         # create buffer for metrics
         input0 = jax.tree_map(lambda x: x[0], inputs)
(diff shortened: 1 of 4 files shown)


ruomingp (commented):
Could you explain why this is needed? We find usually it's more efficient to use either a larger mesh or a smaller batch size.

axlearn/common/learner.py
```diff
@@ -94,8 +95,7 @@ def _prune_empty(in_tree: NestedTensor) -> NestedTensor:
     return prune_tree(in_tree, lambda _, v: isinstance(v, dict) and not v)
 
 
-@dataclasses.dataclass
-class ForwardOutputs:
+class ForwardOutputs(NamedTuple):
```
apghml, line 98: What is the purpose of this change?

axlearn/common/learner.py
```diff
@@ -444,6 +444,153 @@ def _mask_tree(tree: dict, *, keep: dict) -> dict:
     )
 
 
+class MetricsAccumulationOp(NamedTuple):
```
apghml, line 447: Axlearn already has metric accumulation classes that are used by evalers. Could those be reused here instead of defining new classes?

axlearn/common/learner.py
```diff
@@ -9,7 +9,8 @@
 
 import dataclasses
 import enum
-from typing import Callable, Mapping, Optional, Protocol, Sequence, Tuple
+import logging
```
apghml, line 12: We typically use absl logging instead of python logging.

axlearn/common/learner.py
```diff
@@ -444,6 +444,153 @@ def _mask_tree(tree: dict, *, keep: dict) -> dict:
...
+    @config_class
+    class Config(Learner.Config):
+        """Configures Learner."""
+
+        # The number of microbatches to accumulate per optimizer step.
+        microbatches: Required[int] = REQUIRED
+        # tuple of key-value pairs specifying custom aggregation and normalization
+        # for a specific metric
+        metrics_accumulation_key_ops: Sequence[Dict[str, Optional[MetricsAccumulationOp]]] = []
+        gradient_dtype: Optional[jnp.dtype] = jnp.bfloat16
```
apghml, line 493: Does the existing learner class use this? If not, we should try to be consistent with its API.

axlearn/common/learner.py
```diff
@@ -444,6 +444,153 @@ def _mask_tree(tree: dict, *, keep: dict) -> dict:
...
+        and metrics into buffers.
+
+        Args:
+            fn (ForwardFn): Model forward pass
+            inputs (NestedTensor): input microbatches with a leading microbatch axis
+            opt_params (NestedOptParam): params which don't need gradients
+
+        Returns:
+            ForwardBackwardOutputs: pytree containing gradients and metrics
+        """
```
apghml, line 508: I wonder if instead of having a separate learner for microbatching, it would be more flexible to have a generic way of wrapping a ForwardFn so that it uses Jax.lax.map to run the microbatches. Beyond avoiding the need to add a new learner, it would also allow for other microbetching uses outside of learner, eg inference or in second order optimizers.

apghml:
I’m ooo this week. I have left some preliminary comments and will revisit on Monday.

axlearn/common/learner.py
apoorvtintin: The interface of jax.lax.scan does not consider "_dataclass_" a valid JAX type.

axlearn/common/learner.py
apoorvtintin: jax.lax.map gives no guarantees of sequential execution of microbatches which is the key quality of gradient accumulation.

axlearn/common/learner.py
apoorvtintin: These metric accumulation classes are stateless to be usable as carry by jax.lax.scan unlike the ones in the evaler, I can make the class structure similar though

axlearn/common/learner.py
apoorvtintin: Could you please be a little more specific, is the concern naming of variables?

apghml:
Closing since gradient accumulation functionality has been implemented via https://github.com/apple/axlearn/pull/614

PR #476, opened 2024-05-18
Fixes GroupedQueryAttention support in Fuji.

ruomingp:


axlearn/experiments/text/gpt/fuji.py
```diff
@@ -186,8 +188,10 @@ def model_config(
     if ffn_dim is None:
         ffn_dim = scaled_hidden_dim(scale=8 / 3, round_up_to_multiples_of=256)
     if num_kv_heads:
+        atten_cfg = GroupedQueryAttention.default_config()
```
markblee, line 191: May be worth adding a unit test?

markblee (approved):

axlearn/experiments/text/gpt/fuji.py
ruomingp: Done.

PR #481, opened 2024-05-22
Regression test for recent s3 issue.

apghml:


axlearn/common/env_test.py
```diff
@@ -0,0 +1,17 @@
+"""Tests for AXLearn environment."""
+# pylint: disable=no-self-use,redundant-keyword-arg,too-many-function-args
```
markblee, line 2: OOI where did `redundant-keyword-arg,too-many-function-args` come from?

markblee (approved):

axlearn/common/env_test.py
apghml: Removed, thanks! They were from another test that I copied over as a template.

markblee (approved):

PR #483, opened 2024-05-23
add a manually sharded fft, and allow FFT to be configurable in speech frontend

zhiyun:


markblee (commented):
Thanks -- seems to be missing some changes from internal review?

markblee (approved):
Approving since the missing changes are mostly nit...

PR #496, opened 2024-05-28
Enables tcmalloc to mitigate host OOM.

markblee:


zhzhyi (approved):
Do we need to set PYTHONMALLOC=malloc too or is this change only for xla ?

PR #521, opened 2024-06-13
bump up hf version to 4.41.2

gyin94:


axlearn/common/bert_test.py
```diff
@@ -68,10 +68,7 @@ def dummy_inputs_for_mlm(
     np.putmask(input_ids, attention_mask, padding_input_id)
     # hf expects the opposite masking scheme as we do:
     # A float value of 0. represents padding and 1. represents non-padding.
-    attention_mask = np.logical_not(attention_mask)
-    hf_attention_mask = np.logical_and(
-        attention_mask[:, None, :], attention_mask[:, :, None]
-    ).astype(jnp.float32)
+    hf_attention_mask = np.logical_not(attention_mask)
```
gyin94, line 71: _prepare_4d_attention_mask_for_sdpa in huggingface transformers has changed for bert model. we only need a shape of [batch_size, seq_len] now instead of [batch_size, seq_len, seq_len] for unit tests

markblee (approved):

PR #525, opened 2024-06-18
Request memory for TPUGKEJob pod based on TPU machine type

HaijingFu:
Set memory request for worker pod according to its TPU machine type. This can make the pod less vulnerable to eviction.

1. Create a <machine_type to memory> mapping
2. Set the memory using the mapping in _build_cotianiner()

Memory setting reference: https://cloud.google.com/tpu/docs/tpus-in-gke

### DIFF
--- a/axlearn/cloud/gcp/job.py
+++ b/axlearn/cloud/gcp/job.py
@@ -26,7 +26,10 @@ from axlearn.cloud.common.utils import parse_kv_flags, subprocess_run
 from axlearn.cloud.gcp.config import default_project, default_zone, gcp_settings
 from axlearn.cloud.gcp.node_pool import PRE_PROVISIONER_LABEL
 from axlearn.cloud.gcp.scopes import DEFAULT_TPU_SCOPES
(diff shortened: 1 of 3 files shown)


axlearn/cloud/gcp/job.py
```diff
@@ -425,6 +428,15 @@ def _build_container(self) -> Nested[Any]:
...
             env_vars["ENABLE_ICI_RESILIENCY"] = str(cfg.enable_tpu_ici_resiliency).lower()
 
+        resources = {"limits": {"google.com/tpu": system.chips_per_vm}}
+        # Set request memory by host machine type.
+        machine_memory_gb = GCE_MACHINE_TYPE_TO_REQUEST_MEMORY_CHARACTERISTICS.get(
+            system.gce_machine_type, None
+        )
+        if machine_memory_gb is not None:
+            resources["limits"]["memory"] = f"{machine_memory_gb}G"
+            resources["requests"] = {"memory": f"{round(machine_memory_gb * 0.8, 2)}G"}
```
jinglu1, line 438: Does it mean that we are reserving 20% of memory for system software?

HaijingFu: Yes.

PR #536, opened 2024-06-21
Fix VDict compatibility with tree_map.

apghml:


markblee (approved):
Thanks!

ruomingp (commented):
Thanks! Could you check whether this will affect checkpoint backwards compatibility?

apghml:
> Thanks! Could you check whether this will affect checkpoint backwards compatibility

Backwards compatibility unit test added.

apghml:
Test asserts have been fixed.

ruomingp:
> > Thanks! Could you check whether this will affect checkpoint backwards compatibility
> 
> Backwards compatibility unit test added.
> 
> The reason it doesn't affect backwards compatibility is the AXLearn checkpoint saving/restoring code doesn't track the order of dictionary keys.

Ah OK. Thanks for checking!

ruomingp (approved):

chongw (approved):

PR #542, opened 2024-06-26
Add github action support

samos123:
Testing here until GHA is enabled on axlearn: https://github.com/samos123/axlearn/pull/1

The GHA are working except clip_test.py seems to be broken but I suspect it's broken on circleCI too.

I would prefer to merge this with excluding clip_test.py. I or the whoever wrote/knows clip_test.py can do a follow up PR.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10"]
     outputs:
       pytest-files: ${{ steps.find-test-files.outputs.pytest-files }}
 
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -49,12 +49,13 @@ jobs:
       - uses: actions/checkout@v3
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
-      - name: Build and Test
-        env:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yaml
```diff
@@ -0,0 +1,61 @@
...
+  build-and-test:
+    needs: set-matrix
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: ${{ fromJson(needs.set-matrix.outputs.matrix) }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - uses: docker/setup-buildx-action@v3
```
madrob, line 52: Is this right that we set up twice?

.github/workflows/build-test.yaml
```diff
@@ -0,0 +1,61 @@
...
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: define pytest testing matrix
+        id: set-matrix
+        run: |
+          echo "matrix=$(python3 .github/workflows/set_matrix.py)" >> $GITHUB_OUTPUT
+
+  build-and-test:
```
madrob, line 43: Is there a way to set the name here to something nicer than the list of files? I'm imagining `build-and-test (1/10)` and then first step can be to echo list of files to output for debugging later?

samos123:
@madrob Thanks for the feedback! I probably won't be able to attend to this PR for a few weeks. Feel free to take make a separate PR and discard this PR.

ruomingp (commented):
Will defer to @madrob and @markblee for review.

PR #556, opened 2024-06-28
add kv_state arg to forward function in RoFormerQKVLinear

chongw:


axlearn/common/attention.py
```diff
@@ -1306,6 +1306,7 @@ def forward(
         *,
         key: Optional[Tensor] = None,
         value: Optional[Tensor] = None,
+        kv_state: Optional[KVState] = None,
```
ruomingp, line 1309: If `kv_state` is not None, how is it used? Should it be passed to `i_proj` below? Otherwise we should throw a NotImplementedError if kv_state is not None.

ruomingp, line 1309: Add a test that would fail without this change?

chongw: the function will take kv_state (None or not None) but it will just ignore it.

PR #572, opened 2024-07-02
Preserve microbatch sharding of carry inputs in Pipeline.

markblee:
Previously, carry inputs were explicitly padded to `M + N - 1` and scanned over directly. While simple, the dim cannot be sharded because it's being indexed by scan, and we also waste memory due to padding values.

### DIFF
--- a/axlearn/common/pipeline.py
+++ b/axlearn/common/pipeline.py
@@ -36,6 +36,7 @@ from axlearn.common.base_layer import BaseLayer, FactorizationSpec, NestedParame
 from axlearn.common.config import REQUIRED, InstantiableConfig, Required, config_class
 from axlearn.common.module import Module, NestedTensor, Tensor, child_context, new_output_collection
 from axlearn.common.utils import (
+    Nested,
     NestedPartitionSpec,
     VDict,
     get_or_none,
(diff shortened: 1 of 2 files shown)


axlearn/common/pipeline.py
```diff
@@ -305,80 +310,99 @@ def vmap_fn(
...
+                # Note that vmap outputs are typically sharded over stages and may incur extra
+                # communication per-iteration (e.g. from broadcasting last stage outputs).
+                return carry_out, vmap_out
+
+            state_t0 = dict(
+                # Current loop iteration.
+                t=jnp.array(0, dtype=jnp.int32),
+                # [N, microbatch_size, ...].
+                carry_output_t_1=jax.tree_util.tree_map(
+                    lambda x: jnp.zeros((n,) + x.shape[1:]), carry
```
ptoulme-aws, line 380: see this PR: https://github.com/markblee/axlearn/pull/1

markblee: Thanks -- how can I reproduce this error? (Also added `x.dtype`, in case that resolves it.)

ruomingp (approved):

axlearn/common/pipeline.py
ruomingp: Do we have a test for bfloat16?

markblee: Added a test that fails without the change.

ruomingp (approved):
Thanks for adding the test!

PR #578, opened 2024-07-07
[6/N] refactor async_generate with request

gyin94:
This PR refactor async_generate to use `request` field instead of `messages`, `prompt`, `tools`.

- More flexible on using other content in the request
- Needed for some clients to sanitize the request with its own rule. For example, some clients need to cutoff the function name to be less than 32 characters. And we will sanitize it here and store it the response jsonl.

### DIFF
--- a/axlearn/open_api/gemini.py
+++ b/axlearn/open_api/gemini.py
@@ -65,7 +65,7 @@ class GeminiClient(BaseClient):
         """
         if "messages" not in request:
             raise ValidationError("Field messages must be in request.")
-        _santize_request(request=request)
+        _format_request(request=request)
         contents = _convert_openai_messages_to_gemini(messages=request["messages"])
         if request.get("tools", None) is not None:
             gemini_tools = _convert_openai_tools_to_gemini(tools=request["tools"])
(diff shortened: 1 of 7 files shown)


axlearn/open_api/gemini.py
```diff
@@ -188,6 +187,24 @@ def _aggregate_tool_role_messages(messages: List[Dict[str, Any]]) -> List[Dict[s
     return aggregated_messages
 
 
+def _santize_request(request: Dict[str, Any]):
+    """Santizes request to follow Gemini request rules."""
```
markblee, line 191:
```suggestion
def _format_request(request: Dict[str, Any]):
    """Formats request to follow Gemini request rules."""
```
Since it seems more like formatting than sanitization?

axlearn/open_api/openai.py
```diff
@@ -69,20 +65,22 @@ async def async_generate(
         """
         cfg: OpenAIClient.Config = self.config
         client: AsyncOpenAI = self._client
-        assert prompt is not None or messages is not None, ValidationError(
+        prompt = request.get("prompt", None)
+        messages = request.get("messages", None)
+        assert prompt is None or messages is None, ValidationError(
```
markblee, line 70: Is this correct? Previously we require one to be non-None, but now we check that one of them is None.

axlearn/open_api/gemini.py
```diff
@@ -188,6 +187,24 @@ def _aggregate_tool_role_messages(messages: List[Dict[str, Any]]) -> List[Dict[s
...
+    if "messages" in request:
+        new_messages = []
+        for message in request["messages"]:
+            new_messages.append(_format_tool_message(message=message))
+        request["messages"] = new_messages
```
markblee, line 196: nit -- could be simpler with list comprehensions.

axlearn/open_api/gemini.py
gyin94: changed

axlearn/open_api/openai.py
gyin94: changed. and added unit test

axlearn/open_api/gemini.py
gyin94: changed

markblee (approved):

PR #579, opened 2024-07-10
Gala base model + Gala sigmoid attention

WeersProductions:
- Adds Gala base model implementation
- Adds Gala sigmoid model implementation
- Adds Pajama-based trainers


axlearn/experiments/text/gpt/fuji.py
```diff
@@ -2,7 +2,7 @@
 
 """Utilities to set up the 'fuji' GPT model trainer configs.
 
-Architecture names follow apple varieties: fuji, gala, honeycrisp, etc.
+Architecture names follow apple varieties: Fuji, Gala, Honeycrisp, etc.
```
ruomingp, line 5: Could you confirm that this PR doesn't change the step times for the Fuji models?

markblee (commented):
Please see internal comment. Thanks!

PR #581, opened 2024-07-10
Add core dependency and keep a minimal dependency on axlearn

gyin94:


ruomingp (changes requested):
Could you address my question on the internal PR? Thanks.

markblee (approved):

ruomingp (approved):

PR #592, opened 2024-07-17
Add job_id field in BastionManagedJob

HaijingFu:
Add job_id field in the BastionManagedJob JobMetadata to better track and manage the job.

### DIFF
--- a/axlearn/cloud/common/utils.py
+++ b/axlearn/cloud/common/utils.py
@@ -62,6 +62,11 @@ def generate_job_name() -> str:
     return f"{os.environ['USER'].replace('_', '')}-{uuid.uuid4().hex.lower()[:6]}"
 
 
+def generate_job_id() -> str:
+    """Generate a unique job uuid."""
+    return str(uuid.uuid4())
+
+
 # TODO(markblee): Consider using git python.
 def get_git_revision(revision: str) -> str:
     """Gets the commit hash for the revision."""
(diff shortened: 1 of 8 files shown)


markblee (approved):

axlearn/cloud/common/utils.py
```diff
@@ -62,6 +62,11 @@ def generate_job_name() -> str:
     return f"{os.environ['USER'].replace('_', '')}-{uuid.uuid4().hex.lower()[:6]}"
 
 
+def generate_job_id() -> str:
+    """Generate a unique job uuid."""
+    return str(uuid.uuid4())
```
apghml, line 67: IIUC, uuid4 is a random uuid. Do we know if the randomness used is good? Would it make sense to instead uses `secrets.token_urlsafe()`, which is cryptographically random?

HaijingFu: 1 in 2.71 x 10^18 : Two random v4 UUIDs collide. Considering our job numbers, the risk should be fairly low.

apghml: Yes, the odds are something like that assuming the UUIDs are actually random.
I was wondering about whether the source of randomness was a good (i.e., cryptographic) source of randomness, which would suffice for that. If the source of randomness isn't good, then collisions can be much more probable than that. E.g., some PRNGs are periodic with a relatively small period.

I just took a look at the implementation of uuid4, and it does appear that it uses `os.urandom`, which is suitable for cryptographic use. However, this guarantee isn't documented. For future uuid generation, it might be better to use `secrets`.

PR #622, opened 2024-08-03
allow using on-demand instead of spot only

### LINKED ISSUE #621: Unable to use TPU on GKE using on-demand quota
Currently axlearn either adds a nodeSelector for spot=true or it adds a nodeSelector for reservation:
```
        if tier == "0" and cfg.reservation is not None:
            logging.info("Found tier=%s in env. Using reservation=%s", tier, cfg.reservation)
            selector.update({"cloud.google.com/reservation-name": cfg.reservation})
        else:
            logging.info("Found tier=%s in env. Using spot quota", tier)
            selector.update({"cloud.google.com/gke-spot": "true"})
            tolerations.append(
                {
                    "key": "cloud.google.com/gke-spot",
                    "operator": "Equal",
                    "value": "true",
                    "effect": "NoSchedule",
                }
            )
```

It should be possible to launch a job using on-demand TPU, however today that's not possible unless you remove this line:
```
selector.update({"cloud.google.com/gke-spot": "true"})
```

samos123:
This fixes #621

### DIFF
--- a/axlearn/cloud/gcp/job.py
+++ b/axlearn/cloud/gcp/job.py
@@ -371,6 +371,7 @@ class TPUGKEJob(GKEJob):
 
         accelerator: AcceleratorConfig = AcceleratorConfig()
         reservation: Optional[str] = None
+        enable_ondemand: Optional[bool] = None
         enable_tpu_ici_resiliency: Optional[bool] = None
         location_hint: Optional[str] = None
 
@@ -385,6 +386,13 @@ class TPUGKEJob(GKEJob):
             None,
             "Whether to enable TPU ICI resiliency. If None, the decision is left to GCP, as "
             "not all TPU types support this flag.",
+            **common_kwargs,
+        )
+        flags.DEFINE_boolean(
+            "enable_ondemand",
+            None,
+            "Allow the job to run using on-demand quota when no reservation is provided. "
+            "Without this flag a job can only land on spot quota.",
             **common_kwargs,
         )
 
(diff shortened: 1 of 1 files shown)


samos123:
@markblee could you review please? I need this in order to test on internal on-demand capacity.

samos123:
@Ethanlm would love your thoughts too

axlearn/cloud/gcp/job.py
```diff
@@ -508,8 +516,9 @@ def _build_pod(self) -> Nested[Any]:
             logging.info("Found tier=%s in env. Using reservation=%s", tier, cfg.reservation)
             selector.update({"cloud.google.com/reservation-name": cfg.reservation})
         else:
```
markblee, line 518: What about using `tier == "1"` for spot and `tier == "2"` for on-demand? You can configure the quota file s.t. there's only on-demand quota like
```
[[total_resources]]
v5litepod = 0

[[total_resources]]
v5litepod = 0

[[total_resources]]
v5litepod = X

...
```

Which corresponds to 0 quota for tier 0/1 and X quota on tier 2. More generally we can also make the logic for `tier -> selector` mapping configurable, for more flexible schemes. WDYT?

samos123: We generally don't run with bastion or any quota files. I thought some more about this. The ideal UX for a simple setup would be to not have to worry about any kind of bastion tiers, meaning the user does not set any BASTION_TIER env variable. In that case no special nodeSelectors should get added.

samos123:
Closed in favor of https://github.com/apple/axlearn/pull/1096

PR #623, opened 2024-08-03
Enable support for Kueue for GKETPUJob

samos123:
Internal V5E testing capacity requires using Kueue. This PR allows you to specify which Kueue Queue to use. The GPUGKEJob already supported this so moving the shared config to GKEJob.


markblee (approved):

samos123:
Do we need anything else to get this merged? Do we need another review from @ruomingp ?

PR #627, opened 2024-08-05
Add Orbax Checkpointing

jiya-zhang:
At this time, you need to use Orbax Checkpointing to test out in-memory checkpointing. It is also better integrated with Pathways, when the time to test that comes.

This PR is currently a draft. It does the following: (more notes are [here](https://quip-apple.com/cu5HA70gghxY))
- Adds the option to use an orbax checkpointer
- Renames the original checkpointer to `StateStorageCheckpointer`
- Sets `use_orbax` to `False` by default for models under `experiments/text/gpt` folder

To test orbax checkpointer: 
- Manually set `cfg.checkpointer.use_orbax = True`
- Launch training as usual and the checkpoints will be stored in the expected `checkpoints` folder in GCS

Planned work on my side: 
- Add more comments
- Upgrade `orbax-checkpoint` version and consider pinning it in `pyproject.toml`
- Test on TPUs
- Test in-memory checkpointing with Orbax

@markblee Do you mind taking a look and giving some feedback? Please feel free to edit branch directly.


axlearn/common/checkpointer.py
```diff
@@ -27,6 +27,8 @@
 from absl import logging
 from jax.experimental import maps, multihost_utils
 from jax.experimental.array_serialization import serialization as array_serialization
+from orbax.checkpoint.checkpoint_manager import CheckpointManager, CheckpointManagerOptions
```
cpgaffney1, line 30: You should always reference `ocp` rather than importing individual modules.

axlearn/common/checkpointer.py
```diff
@@ -619,16 +627,137 @@ class Config(Module.Config):
...
+            logging.info(f"Encountered the following error when restoring Orbax checkpoint at step {step}: {e}")
+            step = None
+            restored_state = state
+        # TODO (maggiejz): also try to restore from StateStorageCheckpoint format
+        return step, restored_state
+
+    def stop(self):
+        self._checkpoint_manager.wait_until_finished()
+
+    def transform_tensorspec(self, element):
```
cpgaffney1, line 743: "transform" has a bit of a different meaning - this can probably just be something like `to_shape_dtype_struct`.

axlearn/common/checkpointer.py
```diff
@@ -619,16 +627,137 @@ class Config(Module.Config):
...
+        # eg: {'weight': TensorSpec(shape=[32, 8], dtype=<class 'jax.numpy.float32'>, mesh_axes=PartitionSpec(None, 'model'))}
+        if isinstance(element, TensorSpec):
+            # change it into orbax compatible format
+            new_element = jax.ShapeDtypeStruct(
+                shape = element.shape,
+                dtype = element.dtype,
+                sharding = element.sharding
+            )
+        else:
```
cpgaffney1, line 752: Not sure if you intend to leave this else here.

axlearn/common/checkpointer.py
```diff
@@ -619,16 +627,137 @@ class Config(Module.Config):
...
+    def restore(
+        self,
+        *,
+        step: Optional[int] = None,
+        state: Union[NestedTensor, NestedTensorSpec],
+    ) -> Tuple[Optional[int], NestedTensor]:
+        try:
+            transformed_state = jax.tree.map(self.transform_tensorspec, state)
+            restored_state = self._checkpoint_manager.restore(
+                step=self._checkpoint_manager.latest_step(),
```
cpgaffney1, line 728: Redundant accesses of `latest_step`

axlearn/common/checkpointer.py
```diff
@@ -619,16 +627,137 @@ class Config(Module.Config):
...
+                args=ocp.args.StandardRestore(transformed_state)
+            )
+            step = self._checkpoint_manager.latest_step()
+            logging.info("Restored Orbax checkpoints at step %s", step)
```
cpgaffney1, line 732: Logs are a bit redundant since orbax is already logging this info.

axlearn/common/checkpointer.py
cpgaffney1: Also, you can specify step=None to automatically get the latest.

axlearn/common/checkpointer.py
jiya-zhang: Fixed

axlearn/common/checkpointer.py
jiya-zhang: Fixed

axlearn/common/checkpointer.py
jiya-zhang: It was for debugging reasons. I deleted it now

axlearn/common/checkpointer.py
jiya-zhang: I changed to `step=None` within the `restore`

axlearn/common/checkpointer.py
jiya-zhang: Fixed

jiya-zhang:
this PR is implemented in #651

PR #638, opened 2024-08-08
Set remat_spec.policy None for fuji v2 70B

samos123:
Please do not merge yet, but would like a review and start a discussion of the preferred approach in axlearn.

This is needed for v5e-256 otherwise we run out of HBM memory.


axlearn/experiments/text/gpt/fuji.py
```diff
@@ -188,6 +189,10 @@ def get_trainer_kwargs(
                 num_kv_heads=None if version == Version.V1 else 8,
                 rope_theta=rope_theta,
                 flash_attention=flash_attention,
+                remat_spec=RematSpec(
+                    prevent_cse=True,
```
samos123, line 193: not sure if we actually need prevent_cse=True. It may work with false too.

samos123:
@kelvin-zou you mind testing it standalone on your side as well? I've tested this as part of my broader v5e-perf-testing branch here: https://github.com/apple/axlearn/compare/main...samos123:axlearn:v5e-perf-testing

axlearn/experiments/text/gpt/fuji.py
samos123: we may also want this to be conditional based on instance type. E.g. if v5e then use remat_spec.policy = None otherwise use default remat_spec

samos123: It doesn't seem like there is an easy way to add such a conditional though

samos123:
One more thing that a teammate mentioned is using host offload instead of fully disabling jax.checkpointing

samos123:
Maxtext comes with pre-defined remat policies that you can use: https://github.com/google/maxtext/blob/644eb87ae90dd8b210ce17f1c16ca7a54e80fceb/MaxText/configs/base.yml#L132
and
https://github.com/google/maxtext/blob/35da4f98ef309869295b5a1f24a1dab3e169ae57/MaxText/layers/models.py#L302
That could be another approach to take with experiments/gpt

samos123:
Closing for now since host offload for fuji 2 70b is recommended instead. I will re-open if needed.

PR #641, opened 2024-08-08
set hostNetwork to True for TPUGKEJob

samos123:
This helped increase v5e-512 performance from 50% MFU to 58% MFU. The gains could be more significant when going beyond 2 slices.

Majority of Google TPU benchmarks ran on GKE use hostNetwork=true. That's also what XPK sets by default.

The performance increase is due to being able to bypass container networking. Setting hostNetwork=true allows us to directly utilize the host NICs without having to first traverse container NIC and linux network bridges.

### DIFF
--- a/axlearn/cloud/gcp/job.py
+++ b/axlearn/cloud/gcp/job.py
@@ -474,6 +474,7 @@ class TPUGKEJob(GKEJob):
         annotations, selector, volumes, tolerations = {}, {}, [], []
 
         if cfg.gcsfuse_mount:
+            # GCS Fuse does not work with hostNetwork True.
             # Mount a GCS bucket as a volume.
             annotations.update(
                 {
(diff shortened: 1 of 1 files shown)


kelvin-zou (approved):

axlearn/cloud/gcp/job.py
```diff
@@ -563,23 +563,29 @@ def _build_pod(self) -> Nested[Any]:
...
+                "cloud.google.com/gke-tpu-topology": system.topology,
+                **selector,
+            },
+            tolerations=tolerations,
+            containers=[self._build_container()],
+            serviceAccountName=cfg.service_account,
+            volumes=volumes,
+        )
+
+        # hostNetwork True and dnsPolicy do not work with Workload Identity and GCS Fuse.
```
markblee, line 581: I suppose we should check here that workload identity is not being used, not just gcs fuse --- as sync'ed offline, will do more testing to see whether necessary before merging.

samos123: GKE Workload Identity requires forcing metadata traffic to a specific pod. However, if you set hostNetwork true that's no longer possible so it doesn't work.

Note your GKE clusters can keep on using Workload Identity, however the issue is that it won't work for any pod that is using hostNetwork: true. All your other pods using hostNetwork: false will continue to be able to utilize workload identity just like before.

Ethanlm: We can attach the right service account (the same one used in the container) to GKE node pools instead of using the default one. Then it should work with hostNetwork=True

Ethanlm:
Closing this PR due to inactivity. Please re-open or file a new PR if this is still important.

PR #643, opened 2024-08-09
Enable remat checkpoints to host instead of TPU memory

samos123:
This allowed us to get MFU of fuji v2 70B from 58.50% to 61.83%


axlearn/common/attention.py
```diff
@@ -3874,6 +3874,7 @@ def build_remat_spec(
     ],
     self_attention: bool = True,
     feed_forward: bool = False,
+    offload: bool = False,
```
ruomingp, line 3877: Instead of a bool, should we allow the caller to customize `offload_dst` directly?

```suggestion
    offload_dst: Optional[Literal["pinned_host"]] = None,
```

This will make the API more extensible and closer to the JAX API.

ruomingp (commented):
Thanks!

axlearn/common/attention.py
samos123: Agree. I will take another stab at this PR with focus on staying closer to JAX API and extensibility.

samos123: Resolved, could you review again?

axlearn/common/attention.py
```diff
@@ -3891,6 +3904,7 @@ def build_remat_spec(
         stack_cfg: A transformer config.
         self_attention: Checkpoint self attention layer activations if true.
         feed_forward: Checkpoint feed-forward layer activations if true.
+        offload_dst: Destination of remat checkptoing offloading.
```
ruomingp, line 3907: Add a link to the JAX documentation on `offset_dst` on the potential values?

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -188,6 +188,7 @@ def get_trainer_kwargs(
                 num_kv_heads=None if version == Version.V1 else 8,
                 rope_theta=rope_theta,
                 flash_attention=flash_attention,
+                remat_offload_dst="pinned_host",
```
ruomingp, line 191: Add a comment on the observed MFU and step time?

ruomingp, line 191: Is this option limited to 70B? Do we want to apply it to 7B and other models?

ruomingp (commented):
Thanks!

axlearn/experiments/text/gpt/fuji.py
samos123: Only use remat checkpoint offload into host when you could benefit from the extra TPU memory.

If there is plenty TPU memory then remat checkpoint into TPU memory.
If TPU memory is low or you want to squeeze as much per device batch as possible, then use offload_dst=pinned_Host.

So yes, it could make a lot of sense for 7B too on V5e and Trilium since then we can possibly increase per device batch size.

samos123: Done

samos123: I would prefer to have this PR focus on enabling remat offload and 70B. As a follow up I can do the following as part of V5E perf benchmarking:
* Enable remat offload for 7B and compare performance before and after
* See if I can increase per device batch size for 7B after enabling remat_offload

Would that work for you?

axlearn/common/attention.py
ruomingp: Sounds good. Thanks!

axlearn/experiments/text/gpt/fuji.py
ruomingp: Sounds good. Thanks.

ruomingp (approved):

samos123:
@markblee could you trigger CI, review and merge if good?

axlearn/common/attention.py
markblee: Was this change pushed?

samos123: No ah I misunderstood Ruoming's comment and thought he was fine there being no docs. Let me add the link to maxtext as a comment.

samos123: Done

axlearn/common/attention.py
```diff
@@ -3891,6 +3904,10 @@ def build_remat_spec(
         stack_cfg: A transformer config.
         self_attention: Checkpoint self attention layer activations if true.
         feed_forward: Checkpoint feed-forward layer activations if true.
+        offload_dst: Destination of remat checkptoing offloading. Relevant Maxtext example:
+          https://github.com/google/maxtext/blob/ebd39aa64d670fa13a313b6f776e01ad9e450321/MaxText/layers/models.py#L230.
+
+
```
markblee, line 3910:
```suggestion
```

markblee (approved):

ruomingp (approved):

PR #651, opened 2024-08-14
An integration of orbax checkpointer.

markblee:
Depends on https://github.com/apple/axlearn/pull/650.

The relevant changes are in `checkpointer.py`, `checkpointer_test.py`, and `pyproject.toml`. It also depends on an unreleased commit for orbax (for concurrency bounded serialization).

### DIFF
--- a/axlearn/common/checkpointer.py
+++ b/axlearn/common/checkpointer.py
@@ -11,12 +11,25 @@ import threading
 import time
 from concurrent import futures
 from types import TracebackType
-from typing import Any, Dict, List, NamedTuple, Optional, Protocol, Tuple, Type, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    NamedTuple,
+    Optional,
+    Protocol,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
+)
 
 import jax
 import jax.numpy as jnp
 import tensorflow as tf
 from absl import logging
+from etils.epath.abstract_path import Path
 from jax.experimental import maps, multihost_utils
 from jax.experimental.array_serialization import serialization as array_serialization
 
@@ -38,7 +51,14 @@ from axlearn.common.module import (
     install_context_stack,
 )
 from axlearn.common.summary_writer import CheckpointerAction, SummaryWriter
-from axlearn.common.utils import NestedTensor, NestedTensorSpec, Tensor, TensorSpec, set_recursively
+from axlearn.common.utils import (
+    Nested,
+    NestedTensor,
+    NestedTensorSpec,
+    Tensor,
+    TensorSpec,
+    set_recursively,
+)
 
 
 class CheckpointValidationType(str, enum.Enum):
@@ -62,9 +82,15 @@ class CheckpointValidationType(str, enum.Enum):
     CONTAINS_STATE_UP_TO_DTYPE = "CONTAINS_STATE_UP_TO_DTYPE"
 
 
+# Number of digits in the step directory.
+_STEP_NUM_DIGITS = 8
+# Prefix for step directory.
+_STEP_PREFIX = "step"
+
+
 def parse_step_from_dir(step_dir: str) -> int:
     # TODO(markblee): use regex.
-    return int(step_dir[-8:])
+    return int(step_dir[-_STEP_NUM_DIGITS:])
 
 
 def check_state_structure(
@@ -135,6 +161,7 @@ def restore_tf_savables(value_map: Dict[str, Any], *, dir: str):
     for path, value in value_map.items():
         tf_checkpoint = tf.train.Checkpoint(value)
         tf_checkpoint.read(os.path.join(dir, path))
+    return value_map
 
 
 class StateStorageCommitCallback(Protocol):
@@ -185,6 +212,12 @@ def write_index_file(*, ckpt_dir: str, index: Any):
         f.write(json.dumps(index))
 
 
+def read_index_file(ckpt_dir: str):
+    """Reads index files written with `write_index_file`."""
+    with tf.io.gfile.GFile(os.path.join(ckpt_dir, "index"), "r") as f:
+        return json.loads(f.read())
+
+
 def _parse_tensor_spec(spec_dict: Dict[str, str]) -> TensorSpec:
     # The shape string is of format `(dim...)`. [1:-1] removes the parentheses.
     shape = [int(x) for x in spec_dict["shape"][1:-1].split(",") if x]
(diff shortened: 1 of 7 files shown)


jiya-zhang:
Thanks Mark! Adding @cpgaffney1 to take a look as well (this will be the PR that gets merged for Orbax fyi)

axlearn/common/checkpointer.py
```diff
@@ -922,3 +971,216 @@ def validate_and_restore(*, step: int, ckpt_dir: str):
...
+                spec["index"].append((path, value))
+        return spec
+
+    # pylint: disable-next=redefined-builtin
+    def ckpt_dir(self, step: int, dir: Optional[str] = None) -> str:
+        """Obtains the checkpoint dir for the given step."""
+        if dir is None:
+            dir = self._manager.directory
+        # pylint: disable-next=protected-access
+        return str(self._manager._get_save_directory(step, dir))
```
cpgaffney1, line 1084: Use 
```
name_format = ocp.path.step.standard_name_format(
            step_prefix=_STEP_PREFIX,
            step_format_fixed_length=_STEP_NUM_DIGITS,
        )
ocp.path.step.build_step_path(directory, name_format, step)
```

axlearn/common/checkpointer.py
```diff
@@ -922,3 +971,216 @@ def validate_and_restore(*, step: int, ckpt_dir: str):
...
+
+class OrbaxCheckpointer(BaseCheckpointer):
+    """A checkpointer that uses orbax CheckpointManager.
+
+    NOTE: While this class uses index files to do additional validation on checkpoint state, the
+    index file is not used for committing checkpoints (i.e., it is not reliable to check for the
+    presence of 'index' file). Instead, use `checkpoint_paths` to identify committed checkpoints.
+    """
+
+    # Disable lazy imports within the class, to avoid requiring orbax dependency globally.
```
cpgaffney1, line 984: What is the concern with doing this? Orbax doesn't loop in that many additional dependencies.

axlearn/common/checkpointer.py
```diff
@@ -922,3 +971,216 @@ def validate_and_restore(*, step: int, ckpt_dir: str):
...
+            )
+            restored_state = composite_state["state"]
+        except Exception as e:  # pylint: disable=broad-except
+            if step is not None:
+                raise ValueError(f"Failed to restore at step {step}.") from e
+            logging.info("Could not find any completed checkpoints under %s: %s", cfg.dir, e)
+            restored_state = state
+
+        # If we successfully restored from step=None, use the latest step.
+        if step is None:
```
cpgaffney1, line 1145: My preference would be for doing this check first and raising an error earlier if no steps are found, followed by the try/except. I think it's a bit more readable but up to you.

cpgaffney1 (approved):

axlearn/common/checkpointer.py
markblee: Thanks -- the existing behavior is to use the input state when restoring from a checkpoint dir without checkpoints, but raise if step is provided explicitly (so in the except we want to distinguish None vs not-None).

axlearn/common/checkpointer.py
markblee: Mainly to avoid assuming that all users have orbax installed -- once this is stable I can move orbax into the core deps. (It's likely already implied by some other packages, so maybe this is less of an issue.)

axlearn/common/checkpointer.py
markblee: Done.

axlearn/common/checkpointer.py
```diff
@@ -922,3 +953,220 @@ def validate_and_restore(*, step: int, ckpt_dir: str):
...
+            restore_concurrent_gb: Max concurrent GB during write. Defaults to 96 (orbax default).
+        """
+
+        keep_last_n: int = 1
+        validation_type: CheckpointValidationType = CheckpointValidationType.EXACT
+        save_concurrent_gb: int = 96
+        restore_concurrent_gb: int = 96
+
+    @classmethod
+    def checkpoint_paths(cls, base_dir: str) -> List[str]:
```
jiya-zhang, line 986: Is this method ever called?

axlearn/common/checkpointer.py
```diff
@@ -922,3 +953,220 @@ def validate_and_restore(*, step: int, ckpt_dir: str):
...
+                )
+            elif isinstance(value, tf.data.Iterator):
+                spec["index"].append((path, str(type(value))))
+                spec["tf_ckpt_map"][path] = value
+            else:
+                spec["index"].append((path, value))
+        return spec
+
+    # pylint: disable-next=redefined-builtin
+    def ckpt_dir(self, step: int, dir: Optional[str] = None) -> str:
```
jiya-zhang, line 1064: Is this method ever called?

axlearn/common/checkpointer.py
```diff
@@ -922,3 +953,220 @@ def validate_and_restore(*, step: int, ckpt_dir: str):
...
+        spec = self._get_spec(step=step, state=state)
+        self._eval_summaries_sem.acquire()  # pylint: disable=consider-using-with
+        assert not self._eval_summaries
+        self._eval_summaries.update(evaler_summaries or {})
+        self._manager.save(
+            step=step,
+            # The input iterator is saved as part of `save_tf_savables`.
+            args=ocp.args.Composite(
+                index=self._index_args(spec["index"]),
+                tf_ckpt_map=self._tf_ckpt_args(spec["tf_ckpt_map"]),
```
jiya-zhang, line 1090: What are `index` and `tf_ckpt_args`? Does orbax provide similar functionalities, eg. checking tree structure (which I assume is what the index file is for)?

jiya-zhang (commented):
Also, @markblee do you think the amount of logs is acceptable during your testing? Personally, the logs are getting a little too much and hard to debug for me

axlearn/common/checkpointer.py
markblee: Not in this module (but by other code as part of its API).

axlearn/common/checkpointer.py
markblee: Same as other comment.

axlearn/common/checkpointer.py
markblee: Good question! `index` is not just for validating structure but also restoring from tensor specs. `tf_ckpt_args` is for tf saveables like tf data iterators. LMK if orbax has built-in support for these.

jiya-zhang: @cpgaffney1 Do you mind answering if orbax has built-in support for either `index` or `tf_ckpt_args` functionalities here?

markblee: Sorry for the delay -- I've changed the internal impl slightly to use the json handler. Will push an update soon.

ruomingp (approved):

zhzhyi (approved):

PR #652, opened 2024-08-15
Fix precision_recall_curve to correctly handle masked examples

dlin28:
The current precision_recall_curve computation is incorrect for weight=0 or masked examples.
The fix is to ignoring masked examples when computing tp, fp, and thresholds in binary_clf_curve.
Added more tests for checking masked examples are handled correctly.

In addition, fix precision_recall_curve to correctly handle weights and added tests as well.

### DIFF
--- a/axlearn/common/metrics_classification.py
+++ b/axlearn/common/metrics_classification.py
@@ -141,9 +141,8 @@ def binary_clf_curve(
         A dict with keys "fps", "tps", and "thresholds".
             Each is a scalar Tensor with values of shape [num_samples].
             fps, tps have values in [0, inf) and thresholds have values in (-inf, inf).
-            The order is based on descending order of thresholds. Because samples of weight 0 are
-            ignored during calculation, returned thresholds only contain valid thresholds that are
-            used for tps and fps calculation.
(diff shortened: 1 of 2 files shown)


axlearn/common/metrics_classification.py
```diff
@@ -172,6 +177,9 @@ def binary_clf_curve(
     fps = jnp.where(y_pred_diff, fps, jnp.iinfo(jnp.int32).max)
     fps = jax.lax.cummin(fps, reverse=True)
 
+    # Masked entries at the end of thresholds are set to jnp.finfo(jnp.float32).max.
+    # Reset them to the rightmost unmasked threshold value.
+    thresholds = jnp.where(weight, thresholds, thresholds[jnp.argsort(weight)[-1]])
```
markblee, line 182: Thanks, as discussed offline maybe we should remove this to simplify, since the masked values should not be used anyway, and I'm not sure that `jnp.argsort(weight)[-1]` gives the rightmost unmasked value.

dlin28: Done

markblee (approved):

PR #656, opened 2024-08-19
Add litepod config for 70B model

kelvin-zou:
Adding a litepod config to avoid remat on GPU or v5p, also fixing a nit issue on flashattention remat_spec.


axlearn/experiments/text/gpt/fuji.py
```diff
@@ -403,4 +401,24 @@ def make_grad_accum_config(
...
             )
 
+        if model_size == "70B":
+
+            def make_config_with_act_offload(base_config_name: str) -> SpmdTrainer.Config:
+                """Make configs for the v5e/v6e tpu with low HBM."""
+                # pytype: disable=annotation-type-mismatch
+                cfg: SpmdTrainer.Config = config_map[base_config_name]().clone()
+                # pytype: enable=annotation-type-mismatch
+                update_model_remat_config(
```
samos123, line 411: Would it call update_model_remat_config twice this way?

kelvin-zou: unfortunately yes, but those are light weight config changes.

kelvin-zou:
@samos123 Yes, I am working on some more general way to support various HW specific configs, will share it once I have it approved internally.

axlearn/experiments/text/gpt/fuji.py
samos123: hmm as long as it doesn't cause issues, it may be fine?

kelvin-zou:
Close it for a more thorough PR later. No need for it after the other change.

PR #659, opened 2024-08-20
Asynchronous save_tf_savables

hanzhi713:
Save 5~9s during a checkpoint save on TPU. Achieved by first saving to a temporary local directory and then async upload to remote using tf.io.gfile.

### DIFF
--- a/axlearn/common/checkpointer.py
+++ b/axlearn/common/checkpointer.py
@@ -7,6 +7,7 @@ import difflib
 import enum
 import json
 import os.path
+import tempfile
 import threading
 import time
 from concurrent import futures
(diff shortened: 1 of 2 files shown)


axlearn/common/checkpointer.py
```diff
@@ -121,22 +122,47 @@ def filter_for_validation(structure):
...
-# pylint: disable-next=redefined-builtin
-def save_tf_savables(value_map: Dict[str, Any], *, dir: str):
-    """Saves TF savables from `value_map` into `dir`."""
+def _upload_dir(src_dir_handle: tempfile.TemporaryDirectory, *, dst_dir: str):
+    """Upload a directory (non-recursively) from a temporary dir to dst_dir.
+
+    Temporary dir will be deleted after the upload is complete.
+    """
+    src_dir = src_dir_handle.name
+    if not tf.io.gfile.exists(dst_dir):
```
markblee, line 131: nit -- `makedirs` should work even if it exists.

axlearn/common/checkpointer.py
```diff
@@ -272,6 +298,7 @@ def __init__(self, cfg: Config):
             self._manager = array_serialization.GlobalAsyncCheckpointManager(
                 timeout_secs=cfg.timeout_secs
             )
+        self._executor = futures.ThreadPoolExecutor()
```
markblee, line 301: Since we're no longer using a context manager, should we invoke `shutdown` explicitly?

hanzhi713: Technically we don't because when executor is garbage collected, it will signal its worker threads to stop.

axlearn/common/checkpointer.py
hanzhi713: Done.

axlearn/common/checkpointer.py
markblee: We generally should avoid relying on implicitly cleanup at the mercy of the garbage collector. Let's add the `shutdown` when the checkpointer exits the context manager.

markblee: A `stop` method makes sense to me. Thanks!

hanzhi713: Added.

markblee (approved):

hanzhi713:
@markblee Do you know why CI is on hold?

PR #661, opened 2024-08-21
rename positions to input_positions

zhiyun:


markblee (approved):
(Let's hold off on merging until everything is passing..)

PR #664, opened 2024-08-22
Style changes with py39+ as target.

miaojingang:


markblee (commented):
Thanks @miaojingang , please see internal discussion first.

markblee (approved):

PR #671, opened 2024-08-22
Data-sharded and zero-copy async checkpoints

hanzhi713:
Similar to GlobalAsyncCheckpointManager but with few improvements:

1. Writing to tensorstore requires no host-to-host copy most of the time. This reduces host
memory usage while also reduces blocking time of the checkpointing process.
2. Tensorstore calls now run in a background event loop, hiding the cost of `ts.open` and `ts.copy`.
Now, only D2H blocks training while serialization is fully asynchronous.
3. Added additional sharding along data-parallel axis during save to further reduce host memory
overhead and improves D2H time. It's achieved by sharding the first dim that's divisible by the
data-parallel dim. We manipulate shard.index to match the sliced shard, so to tensorstore it
behaves as if we're sharding along the data-parallel axis. If no such dim is found, we use the
old way to save-restore, i.e. using the first (0th) replica to do the save only.
4. Optionally one can specify max_concurrent_gb to limit in-flight host memory during
device-to-host transfers and tensorstore writes.

Overall, this PR reduces the blocking time during checkpoint save by 10~60s (depending on how the model is sharded) and can reduce the host memory usage by N-fold where N is the replication count (=data parallel). Experiments show that this PR can solve OOM issues when training some large models.

### DIFF
--- a/ACKNOWLEDGEMENTS.md
+++ b/ACKNOWLEDGEMENTS.md
@@ -65,6 +65,7 @@ Summary:
 * [tylin coco-caption](#tylin-coco-caption)
 * [Yet-Another-EfficientDet-Pytorch](#yet-another-efficientdet-pytorch)
 * [zihangdai xlnet](#zihangdai-xlnet)
+* [Orbax] (#orbax)
 
 ---
 ### [AI21Labs Jamba](https://huggingface.co/ai21labs/Jamba-v0.1)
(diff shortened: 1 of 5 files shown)


markblee (approved):

BlackBoyZeus:
This is some surgical tensor slicing 🎯

PR #690, opened 2024-09-09
Adding support for Pathways proxy

jesus-orozco:
Enable JAX+Pathways single-controller architecture for coordination of accelerators.
- Set jax_backend to "proxy"
- Additional logic to handle new jax_backend type
- Import previewutilities library to identify pathways-enabled workloads

### DIFF
--- a/axlearn/common/launch_trainer_main.py
+++ b/axlearn/common/launch_trainer_main.py
@@ -7,6 +7,7 @@ from absl import app, flags
 from axlearn.common import launch, launch_trainer, measurement
 from axlearn.common.config import config_for_function
 
+import previewutilities
 
 def main(_):
     measurement.initialize(flags.FLAGS)
--- a/axlearn/common/utils_spmd.py
+++ b/axlearn/common/utils_spmd.py
@@ -53,7 +53,7 @@ def setup(
         if initialization_timeout is not None:
             init_kwargs["initialization_timeout"] = initialization_timeout
 
-        if jax_backend == "tpu":
+        if jax_backend == "tpu" or jax_backend == "proxy":
             if not (
                 distributed_coordinator is None and num_processes is None and process_id is None
             ):
@@ -115,5 +115,6 @@ def setup(
                     f"({initialization_timeout} seconds)."
                 )
         else:
-            jax.distributed.initialize(**init_kwargs)
-            _jax_distributed_initialized = True
+            if jax_backend != "proxy":
+                jax.distributed.initialize(**init_kwargs)
+                _jax_distributed_initialized = True
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,8 @@ ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 # Install dependencies.
 RUN pip install flit
 RUN pip install --upgrade pip
+COPY previewutilities-0.0.5-py3-none-any.whl previewutilities-0.0.5-py3-none-any.whl
+RUN pip install previewutilities-0.0.5-py3-none-any.whl
 
 ################################################################################
 # CI container spec.                                                           #
--- a/axlearn/common/launch.py
+++ b/axlearn/common/launch.py
@@ -113,7 +113,7 @@ def setup():
     logging.info("Devices: %s", devices)
     local_devices = jax.local_devices()
     logging.info("Local Devices: %s", local_devices)
-    if not devices or not all(device.platform == FLAGS.jax_backend for device in devices):
+    if FLAGS.jax_backend != "proxy" and (not devices or not all(device.platform == FLAGS.jax_backend for device in devices) ):
         raise RuntimeError(f"Expected backend {FLAGS.jax_backend}. Got {devices}.")
     if FLAGS.data_dir:
         # TODO(ruoming): Get rid of --data_dir and use only env var DATA_DIR.


pyproject.toml
```diff
@@ -92,6 +92,7 @@ gcp = [
     "google-cloud-build==3.24.1",
     "ml_goodput_measurement==0.0.2",
     "pyOpenSSL>=22.1.0",  # compat with cryptography version.
+    "pathwaysutils@git+https://github.com/google/pathways-utils", # for JAX+Pathways single-controller accelerator coordinator
```
ruomingp, line 95: I'm concerned about depending a github repo directly without specifying a specific version, as any changes to the repo will be automatically reflected. Does `pathways-utils` have versioned releases that can be installed via `pip install`?

ruomingp, line 95: Let's put pathways-utils in a separate section, since it's not a critical dependency for regular GCP usage.

axlearn/common/launch_trainer_main.py
```diff
@@ -2,6 +2,7 @@
 
 """Main function for launching the trainer."""
 
+import pathwaysutils  # pylint: disable=unused-import
```
ruomingp, line 5: Can we do best-effort import of `pathwaysutils`? If the import is not found, we can throw an exception only if jax_backend == "proxy".

pyproject.toml
jesus-orozco: Thanks for the feedback @ruomingp ! The short-term roadmap for the pathways-utils library doesn't account for a Pypi release. Adding a tag to the dependency so we can control which version gets installed with Axlearn.

jesus-orozco: This makes sense. Moving pathways to its own section.

axlearn/common/launch_trainer_main.py
ruomingp: Please address this comment so that users can continue to use launch_trainer_main.py without depending on pathways.

ruomingp (commented):
Thanks.

pyproject.toml
```diff
@@ -143,6 +143,11 @@ orbax = [
     "orbax-checkpoint==0.5.23",
 ]
```
markblee, line 145: Nit:
```suggestion
```

axlearn/common/utils_spmd.py
```diff
@@ -53,7 +53,7 @@ def setup(
         if initialization_timeout is not None:
             init_kwargs["initialization_timeout"] = initialization_timeout
 
-        if jax_backend == "tpu":
+        if jax_backend in ("tpu", "proxy"):
```
markblee, line 56: Add a comment that "proxy" is for the pathways codepath?

pyproject.toml
markblee: Hi @jesus-orozco , is there an ETA for a pypi release? IIRC, introducing git hashes to pyproject causes issues for our own pypi release. We prefer to avoid this if possible.

jesus-orozco: I'm following up with the product team to get a better picture of the roadmap. At the moment Github is the only path forward to install the package, unless we bake the whl directly into the docker image.

axlearn/common/launch_trainer_main.py
jesus-orozco: Moved the import statement of pathwaysutils to the GKE-TPU Job using best effort import, since pathways will only be applied to TPUs on GKE via custom Jobset. 
Hopefully this makes more sense. I have a working example of the jobset creation w/ pathways under a draft [workload PR](https://github.com/apple/axlearn/pull/746).

pyproject.toml
jiya-zhang: @markblee Pypi release ETA is Oct 18

axlearn/common/utils_spmd.py
jesus-orozco: ty! Added a comment

pyproject.toml
jesus-orozco: @markblee Pathwaysutils was released on Pypi this week. We have tested and integrated its latest version into this PR.

axlearn/cloud/gcp/job.py
```diff
@@ -450,6 +452,9 @@ def define_flags(cls, fv: flags.FlagValues):
             "not all TPU types support this flag.",
             **common_kwargs,
         )
+        flags.DEFINE_boolean(
+            "use_pathways", False, "Wether the workload is pathways-enabled.", **common_kwargs
+        )
```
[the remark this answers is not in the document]
jesus-orozco: Thanks this makes sense. Refactored the flag and using dynamic import instead.

axlearn/cloud/gcp/job.py
```diff
@@ -443,6 +444,7 @@ class Config(GKEJob.Config):
         enable_tpu_ici_resiliency: Optional[bool] = None
         location_hint: Optional[str] = None
         enable_tpu_smart_repair: bool = False
+        import_pathways: Optional[list[str]] = []
```
ruomingp, line 447: The dynamic imports do not seem limited to pathways:

```suggestion
        import_modules: list[str] = []
```

axlearn/cloud/gcp/job.py
```diff
@@ -457,6 +459,9 @@ def define_flags(cls, fv: flags.FlagValues):
             "not all TPU types support this flag.",
             **common_kwargs,
         )
+        flags.DEFINE_list(
+            "import_pathways", [], "Modules to enable pathways proxy.", **common_kwargs
```
ruomingp, line 463: Same here and below.

Ethanlm:
I closed this PR since https://github.com/apple/axlearn/pull/1140 has been merged.

PR #693, opened 2024-09-11
Inherit dtype from parent to avoid defaulting to float32

a-metz:


axlearn/common/attention.py
```diff
@@ -659,7 +659,7 @@ class Config(BaseLayer.Config):
         # Dimension of each attention head.
         per_head_dim: Required[int] = REQUIRED
         # Autoregressive cache dtype. Should match the step dtype.
-        # Needs to match the forward dtype for Repeated layers. If None, infer as config.dtype.
+        # Needs to match the forward dtype for Repeated layers. If None, infer as BaseLayer.dtype().
```
markblee, line 662:
```suggestion
        # Needs to match the forward dtype for Repeated layers. If None, infer as `self.dtype()`.
```

markblee (approved):

PR #695, opened 2024-09-11
segment_ids for text_to_lm_training_input outputs

changlan:
We add two optional tensors `input_segment_ids` and `input_positions` to the output of `text_to_lm_training_input`, to denote the segments in the sequence and the positions within each segment.

### DIFF
--- a/axlearn/common/input_lm.py
+++ b/axlearn/common/input_lm.py
@@ -39,6 +39,20 @@ class InputDataType(Enum):
     SEQ2SEQ_NO_MASK = "seq2seq_no_mask"
 
 
+class PackingMethodType(Enum):
+    """Represents methods of packing multiple documents into a single sequence.
+
+    Values:
+        EOS_DELIM_NO_MASK: Documents within a sequence are delimited with an EOS token. No
+            self-attention masks between documents.
+        EOS_DELIM_MASK: Documents within a sequence are delimited with an EOS token. Inject
+            `input_segment_ids` and `input_positions` for self-attention masks in a causal LM.
+    """
+
+    EOS_DELIM_NO_MASK = "eos_delim_no_mask"
+    EOS_DELIM_MASK = "eos_delim_mask"
+
+
 def text_to_lm_training_input(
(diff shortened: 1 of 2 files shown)


axlearn/common/input_lm_test.py
```diff
@@ -185,78 +134,169 @@ def test_training_lm_processor_single_example(self, text: str):
...
+        #      2,     29,  3155,     1, 21820,   296,     2,    29,  3155]
+        window_size = 3
+
+        # Pad the concatenated sequence to 20 tokens.
+        # [    1,  21820,   296,     2,    29,  3155,     1, 21820,  8114,    2
+        #     29,  3155,     1, 21820,   296,     2,    29,  3155,      0,    0]
+        batch_size, max_len = 2, 5
+
+        # Always pad.
+        max_padding_fraction = 1.0
```
ruomingp, line 231: With this change, do we still test the truncation behavior?

axlearn/common/input_lm.py
```diff
@@ -142,12 +163,45 @@ def process_batched(inputs: dict[str, Any]) -> dict[str, Any]:
...
+            segment_ids = tf.where(tf.equal(batched_input_ids, vocab.pad_id), 0, segment_ids)
+
+            assert segment_ids.shape[1] == max_len
+
+            def cummax_on_last_axis(x):
+                """tf.scan implementation of cummax."""
+                assert len(x.shape) == 2
+                # transpose to [seq, batch] for tf.scan, which iterates on dimension 0.
+                x = tf.transpose(x)
+                x = tf.scan(tf.maximum, x, initializer=tf.reduce_min(x, axis=0))
```
ruomingp, line 186: Neat use of tf.scan!

ruomingp (commented):
Can you also add the PR description? Thanks!

axlearn/common/input_lm_test.py
changlan: Great catch! I added a test case of the truncation behavior. PTAL

ruomingp (approved):

PR #701, opened 2024-09-16
Fix a few issues for AOT compilation

changlan:
- Fix the logging in SpmdTrainer when devices list is present.
- In flash attention, obtain the backend type from physical_mesh instead of jax.default_backend().
- Allow custom data_dir for AOT compilation.

### DIFF
--- a/axlearn/common/flash_attention/layer.py
+++ b/axlearn/common/flash_attention/layer.py
@@ -6,8 +6,8 @@ from typing import Optional
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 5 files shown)


changlan:
Oops I didn't have the chance to address your comments before it got auto-merged. I can create another PR for the stylistic changes or leave it as is. WDYT? @markblee

markblee:
No worries, they were nitpicks -- feel free to ignore/address in the future when you have spare cycles.

PR #710, opened 2024-09-23
Don't inherit from protocol.

markblee:
Avoids "attribute error `__non_callable_proto_members__`."


gyin94 (approved):
lgtm. thanks. this unblocks us from upgrading a few more packages that need a higher version of typing-extensions

ruomingp (approved):

PR #712, opened 2024-09-24
Set TF_FORCE_GPU_ALLOW_GROWTH=true by default

samos123:
This is needed to be able to run Fuji v2 70B on GPU without GPU memory OOMs.


kelvin-zou (commented):
Thanks! 
Can we move it to somewhere since it is GPU specific? This launch cmd is shared across GPU and TPU.

samos123:
Hmm but we also set a lot of TPU environment variables in launch.py without any if statements. I don't think there is a better place since it needs to happen before jax is started?

ruomingp (approved):

PR #713, opened 2024-09-26
fix tools call issue

gyin94:
this would solve 2 issues:

- o1 doesn't support tools input
- vLLM latest update would throw an error if tools field are parsed while enable-tool is not passed

### DIFF
--- a/axlearn/open_api/openai.py
+++ b/axlearn/open_api/openai.py
@@ -5,6 +5,7 @@ https://github.com/openai/openai-python
 
 This would work for both ChatGPT and vLLM based open source models.
 """
+import copy
 import json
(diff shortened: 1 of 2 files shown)


axlearn/open_api/openai.py
```diff
@@ -77,11 +78,13 @@ async def async_generate(
                     **kwargs,
                 )
             else:
+                req_kwargs = copy.deepcopy(kwargs)
```
gyin94, line 81: we do a full copy here since kwargs would be reused for retry. Hence we do not want to modify it here just due to `tools` field.

markblee (approved):

PR #720, opened 2024-09-30
Configurable concurrent restore gb

hanzhi713:
The default value of 32GB is too small when restoring some large MOE model. Currently, there's no way to specify the `concurrent_gb` arg of `restore_from_dir` when resuming training from a checkpoint. 

Configs here are quite messed-up because there's a `concurrent_gb` in `TensorStoreStateStorageBuilder`, which is only used for inference but not for restoring training checkpoints. Thus, a separate config entry is needed in `TensorStoreStateStorage`.

### DIFF
--- a/axlearn/common/checkpointer.py
+++ b/axlearn/common/checkpointer.py
@@ -336,9 +336,7 @@ class TensorStoreStateStorage(StateStorage):
                 max_data_shard_degree=cfg.max_data_shard_degree,
             )
         else:
-            self._manager = array_serialization.GlobalAsyncCheckpointManager(
-                timeout_secs=cfg.timeout_secs
-            )
+            self._manager = GlobalAsyncCheckpointManager(timeout_secs=cfg.timeout_secs)
         if cfg.max_concurrent_restore_gb is not None and cfg.max_concurrent_restore_gb <= 0:
             raise ValueError(
                 f"max_concurrent_restore_gb must be strictly positive. "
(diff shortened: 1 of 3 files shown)


axlearn/common/checkpointer.py
```diff
@@ -315,11 +315,14 @@ class Config(StateStorage.Config):
                 `None` and `1` means no sharding. `-1` means fully shard along data-parallel
                 replicas. `>1` means custom sharding degree (currently not implemented).
             max_concurrent_gb: Max concurrent shards (in GB) to write.
```
ruomingp, line 317: Should we rename this `max_current_save_gb` to be symmetric?

axlearn/common/checkpointer.py
```diff
@@ -315,11 +315,14 @@ class Config(StateStorage.Config):
...
+            max_concurrent_restore_gb: Max concurrent shards (in GB) to read during checkpoint
+                restore. `None` or `0` means using a default value of 32GB.
```
ruomingp, line 319: Why do we use 32GB as the default? Is it for backwards compatibility?

hanzhi713: Yeah for backwards compatibility.

axlearn/common/checkpointer.py
hanzhi713: I think we should avoid changing existing config names?

ruomingp: It's actually ok to change existing config names. We can do so in a separate PR though.

hanzhi713: Sure. We can leave it for a future PR.

hanzhi713: Done.

ruomingp (approved):

markblee (approved):

PR #714, opened 2024-09-26
Expose segment_ids to _compute_attention in FlashAttention

changlan:
In axlearn's current implementation, `segment_ids` is converted into `attention_logit_biases` in Decoder, and only `attention_logit_biases` is exposed to the underlying attention layers. However, although `attention_logit_biases` is more general than `segment_ids`, the materialized [bnts] tensor is often too large for most use cases (e.g. causal masking, segment masking, etc). Since flash attention supports segment_ids natively, we can provide `segment_ids` along with `attention_logit_biases` and use `segment_ids` whenever it is possible.

This change significantly reduces HBM usage of segment masking. `segment_ids` is also useful for implementing segment masking with SplashAttention as well.

I also updated the Triton kernel implementation so that it can take `bias` and `segment_ids` simultaneously.

PR #721, opened 2024-10-01
Fix the signature of TransformerLayer

changlan:
#714 added an optional `target_segment_ids` argument to TransformerLayer. This PR fixes the remaining subclasses of `BaseTransformerLayer`.

### DIFF
--- a/axlearn/common/attention.py
+++ b/axlearn/common/attention.py
@@ -3823,6 +3823,8 @@ class _TransformerPipeline(Pipeline):
         carry_in = dict(data=data)
         return_aux = return_aux or set()
 
+        # Even though attention logit biases do not change across layers, we
+        # include them in the carry so that they are aligned with the microbatches.
         carry_in.update(kwargs)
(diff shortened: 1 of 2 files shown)


axlearn/common/attention.py
```diff
@@ -3816,23 +3817,13 @@ def forward(
...
-        self_attention_logit_biases: Optional[Tensor] = None,
-        cross_attention_data: Optional[Tensor] = None,
-        cross_attention_logit_biases: Optional[Tensor] = None,
         return_aux: Optional[set[str]] = None,
+        **kwargs,
     ) -> TransformerLayer.Output:
         carry_in = dict(data=data)
         return_aux = return_aux or set()
 
-        # Even though attention logit biases do not change across layers, we
```
markblee, line 3827: Should we retain the comment?

changlan: Done

markblee (approved):

PR #738, opened 2024-10-07
Reduce intermediate stack frames by ~50%.

apghml:


axlearn/common/base_layer.py
```diff
@@ -267,84 +269,101 @@ def _methods_to_wrap_for_auto_child_context(self) -> dict[str, Callable]:
...
+    ) -> dict[str, Callable]:
+        cfg: Module.Config = self.config
 
-    def param_init(self) -> param_init.Initializer:
-        init = getattr(self, "_param_init", None)
-        if init is not None:
-            return init
-        return self.parent.param_init()
+        if cfg.remat_spec is not None:
+            for method_name, method_fn in dict(methods).items():
```
markblee, line 278:
```suggestion
            for method_name, method_fn in methods.items():
```
Did we need the `dict`?

axlearn/common/base_layer.py
```diff
@@ -267,84 +269,101 @@ def _methods_to_wrap_for_auto_child_context(self) -> dict[str, Callable]:
...
-        return {}
+        return super()._wrap_methods_with_auto_child_context(methods)
 
-    def _call_thunk(self, *args, method_fn, **kwargs) -> Callable[[], Any]:
-        cfg: Module.Config = self.config
-        self.vlog(
-            3, "BaseLayer._call_thunk %s.%s: remat=%s", self.path(), method_fn, cfg.remat_spec
-        )
+    def _maybe_wrap_with_remat(self, method_fn: Callable) -> Callable:
+        """Maybe wrap `method_fn` with jax remat.
```
markblee, line 284: I may have missed it, but where is the comment about when wrapping happens?

axlearn/common/module.py
```diff
@@ -520,10 +536,29 @@ def _wrap_method_with_auto_child_context(*, method_fn: Callable, method_name: st
...
+        # TypeErrors to make it easier to see issues where a wrapped method is called
+        # by the user with the wrong signature.
+        try:
+            return _call_method_in_context(
+                self, *args, method_fn=method_fn, method_name=method_name, **kwargs
+            )
+        except TypeError as e:
+            # Make it easier to see what call triggered the error in CI.
+            # When running in an environment like TPUs where stack summaries are available,
+            # this is unecessary and we would have slightly cleaner summaries without it.
```
markblee, line 551: Do you have an example with/without this? Since CI logs are usually easy to search I wonder if we need this?

markblee (approved):
Thanks!

axlearn/common/base_layer.py
apghml: It's in the docstring on the next line:

```
This is called from `BaseLayer._wrap_methods_with_auto_child_context`.
```

apghml: (In the past ruoming has asked me to add a copy when doing this kind of iteration.)

axlearn/common/base_layer.py
apghml: (In the past Ruoming has asked me to do a copy when doing this kind of iteration.)

axlearn/common/module.py
apghml: It's mainly copied from the already existing code here which would otherwise be eliminated by this PR:
https://github.com/apple/axlearn/blob/2e98fb64a6b33dcca7fdbc8fe3c5243127839b95/axlearn/common/module.py#L875

My understanding is that it was originally added because without it, it can be hard to tell when a module method is called with an incorrect signature, what method was being called and what arguments it was called with because of all the wrapping.

Example with:
```
[...]
E           TypeError: Type error when calling <class 'axlearn.common.base_layer_test.TestLayer'>@test.child.<function TestLayer.forward at 0xa87d8ef89> with args=[<class 'jax._src.interpreters.partial_eval.DynamicJaxprTracer'>, <class 'int'>] and kwargs={}

module.py:561: TypeError
```

Example without
```
[...]
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
module.py:513: in _call_method_in_context
    return method_fn(module, *args, **kwargs)
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
base_layer_test.py:81: in forward
    return self.child(x, 5)
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
module.py:919: in __call__
    return self.forward(*args, **kwargs)
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
module.py:545: in wrap_method_fn
    return _call_method_in_context(
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
traceback_util.py:146: in in_context_exception_wrapper
    return fn(*args, **kwargs)
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
module.py:513: in _call_method_in_context
    return method_fn(module, *args, **kwargs)
[...]
E       TypeError: TestLayer.forward() takes 2 positional arguments but 3 were given

traceback_util.py:268: TypeError
```

FWIW, I also have a separate PR that eliminates this and surfaces the TypeError directly below the original call site in a cleaner way, but that is a somewhat higher risk change I think.

apghml: It's mainly copied from the already existing code here which would otherwise be eliminated by this PR:
https://github.com/apple/axlearn/blob/2e98fb64a6b33dcca7fdbc8fe3c5243127839b95/axlearn/common/module.py#L875

My understanding is that it was originally added because without it, it can be hard to tell when a module method is called with an incorrect signature, where the call happened and what arguments it happened with.

Example with:
```
[...]
E           TypeError: Type error when calling <class 'axlearn.common.base_layer_test.TestLayer'>@test.child.<function TestLayer.forward at 0xa87d8ef89> with args=[<class 'jax._src.interpreters.partial_eval.DynamicJaxprTracer'>, <class 'int'>] and kwargs={}

module.py:561: TypeError
```

Example without
```
[...]
E       TypeError: TestLayer.forward() takes 2 positional arguments but 3 were given

traceback_util.py:268: TypeError
```
Both stack traces in the part I elided with [...] do contain the line
```
base_layer_test.py:81: in forward
    return self.child(x, 5)
```
but it's easy to miss as it's buried in something that looks like:
```
[...]
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
module.py:545: in wrap_method_fn
    return _call_method_in_context(
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
traceback_util.py:146: in in_context_exception_wrapper
    return fn(*args, **kwargs)
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
module.py:513: in _call_method_in_context
    return method_fn(module, *args, **kwargs)
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
base_layer_test.py:81: in forward
    return self.child(x, 5)
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
module.py:919: in __call__
    return self.forward(*args, **kwargs)
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
module.py:545: in wrap_method_fn
    return _call_method_in_context(
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
traceback_util.py:146: in in_context_exception_wrapper
    return fn(*args, **kwargs)
traceback_util.py:268: in stack_annotation_wrapper
    return fn(*args, **kwargs)
module.py:513: in _call_method_in_context
    return method_fn(module, *args, **kwargs)
```

FWIW, I also have a separate PR that eliminates this and surfaces the TypeError directly below the original call site in a cleaner way, but that is a somewhat higher risk change I think.

markblee: I see, thanks for the clarification. Improving it in a separate PR SGTM.

PR #739, opened 2024-10-07
Use regex for parsing step_dir

nlusskin:
Instead of assuming the end of a step directory will be a series of digits to be converted to a step number, uses regex to find the last group of `STEP_NUM_DIGITS` from the directory passed into the function.

### DIFF
--- a/axlearn/common/checkpointer.py
+++ b/axlearn/common/checkpointer.py
@@ -7,6 +7,7 @@ import difflib
 import enum
 import json
 import os.path
+import re
 import tempfile
 import threading
 import time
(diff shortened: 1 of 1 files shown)


axlearn/common/checkpointer.py
```diff
@@ -81,8 +82,10 @@ class CheckpointValidationType(str, enum.Enum):
 
 
 def parse_step_from_dir(step_dir: str) -> int:
-    # TODO(markblee): use regex.
-    return int(step_dir[-STEP_NUM_DIGITS:])
+    step = re.findall(r'(\d{' + str(STEP_NUM_DIGITS) + r'})', step_dir)
```
markblee, line 85: Thanks! I do think we should be restrictive about the patterns that we accept, specifically that it ends with `STEP_NUM_DIGITS` digits. Should we update to a pattern that matches that and also add a test case?

nlusskin: We run into an issue quite often where we'll have some checkpoint suffixed (i.e. `checkpoints/step_00001000-ext`) where the only way to run this with axlearn is to copy this directory to `step_00001000`. How would you suggest we incorporate a pattern like this? Happy to add test cases to this PR as well.

ruomingp (commented):

axlearn/common/checkpointer.py
markblee: I'll probably need to understand more about your use-case and why we need to suffix checkpoints. Can you open an internal PR for discussion?

apghml: Usually, the way to workaround this issue is to do something like `xyz/step_00001000` instead of `step_00001000-xyz`. But +1 to Mark's suggestion of opening an internal PR to discuss further.

PR #741, opened 2024-10-08
Fix garbage collector not garbage collecting for gcs dirs

hanzhi713:
Fixed garbage collection bugs introduced by #703.

The cause is that `tf.io.gfile.listdir` will append trailing slash to gcs dirs. However, `checkpoint_paths` will return path without the trailing slash due to the use of `os.path.dirname`. This causes the `in` operator to fail.

This is not caught by unit tests since `tf.io.gfile.listdir` has inconsistent behavior. When listing local dirs, it has no trailing slash.

Before #703, we use `tf.io.gfile.glob` consistently and therefore there're no mismatching slashes.

We fixed this bug by stripping the trailing slash if present. We also added a unit tests to catch this.

### DIFF
--- a/axlearn/common/checkpointer.py
+++ b/axlearn/common/checkpointer.py
@@ -915,7 +915,9 @@ class Checkpointer(BaseCheckpointer):
         remaining_dirs, gc_dirs = [], []
 
         try:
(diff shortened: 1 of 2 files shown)


markblee (approved):
Thanks!

PR #743, opened 2024-10-09
Sliding window local attention

changlan:
* Added sliding window local attention feature to MultiheadAttention and FlashAttention.
* Added or_masks and and_masks which would be useful to composite different mask functions.
* Fixed a bug that causes TPU decoding always to fail.

### DIFF
--- a/axlearn/common/attention.py
+++ b/axlearn/common/attention.py
@@ -314,6 +314,7 @@ class BaseTransformerLayer(BaseLayer):
         raise NotImplementedError(type(self))
 
 
+# TODO(c_lan): Rename `..._mask` functions to "..._biases".
 def make_causal_mask(seq_len: int) -> Tensor:
     """Generates attention logit biases for causal masking.
 
@@ -326,6 +327,20 @@ def make_causal_mask(seq_len: int) -> Tensor:
     """
     # TODO(sneha): support batching
     return _bool_to_bias(causal_mask(jnp.arange(seq_len)[:, None], jnp.arange(seq_len)[None, :]))
+
+
+def make_sliding_window_causal_mask(seq_len: int, sliding_window_size: int) -> Tensor:
+    """Generates attention logit biases for sliding window attention.
+
+    Args:
+        seq_len: Sequence length.
+
+    Returns:
+        A float tensor of shape [seq_len, seq_len] where the value at [i, j] = -inf
+        if i - j > sliding_window_size or i < j, 0 otherwise.
+    """
+    mask_fn = sliding_window_causal_mask(sliding_window_size)
+    return _bool_to_bias(mask_fn(jnp.arange(seq_len)[:, None], jnp.arange(seq_len)[None, :]))
 
 
 def _bool_to_bias(mask: Tensor) -> Tensor:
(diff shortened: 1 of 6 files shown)


axlearn/common/attention.py
```diff
@@ -1781,8 +1832,6 @@ def _forward_for_mode(
                 "key and value must be both None or both set, "
                 f"key:{type(key)}, value:{type(value)}"
             )
-        if self._mask_fn and (key is not None or value is not None):
-            raise ValueError("key and value are not expected when using an attention mask.")
```
ruomingp, line 1785: Can you remind me why this logic is removed?

ruomingp (approved):

axlearn/common/attention.py
changlan: It breaks [this test](https://github.com/apple/axlearn/blob/9ba319391c9019df2ff56e192320c45c8808c166/axlearn/common/flash_attention/layer_test.py#L498) where `key` and `value` are specified. It doesn't seem to me that this restriction is necessary, as `MaskFn` does expect `key_position` as well. WDYT?

markblee (approved):

axlearn/common/attention.py
ruomingp: Got it. Thanks!

axlearn/common/flash_attention/tpu_attention_test.py
```diff
@@ -49,6 +50,19 @@ class TestFlashAttention(TestCase):
         ),
     ]
 
+    @parameterized.product(seq_len=[8, 16, 32, 128], sliding_window_size=[4, 8, 16])
+    def test_sliding_window_mask(self, seq_len, sliding_window_size):
+        shape = (seq_len, seq_len)
+        ref_mask = splash_attention_mask.LocalMask(
+            shape=shape, window_size=(sliding_window_size, None), offset=0
```
qdavid1, line 57: According to the documentation of splash_attention_mask.LocalMask(), setting window_size=(sliding_window_size, None) would make the mask unbounded on the right, which would not be the same behavior as sliding_window_causal_mask().

changlan: It is fixed in https://github.com/apple/axlearn/pull/758

qdavid1: Thanks for fixing!

axlearn/common/flash_attention/layer.py
```diff
@@ -190,11 +205,14 @@ def _compute_attention(
                 )
             attention_logit_biases = attention_logit_biases.astype(q_proj.dtype)
 
-        mask_fn = self._mask_fn
+        if attention_logit_biases is None or self._mask_fn is causal_mask:
```
berlino, line 208: @changlan  I guess this condition should be 

if attention_logit_bias is None or not self._is_mask_fn_used()

changlan: With today's `_logit_biases_for_mask` impl, it should be safe to just `if attention_logit_bias is None`.

PR #745, opened 2024-10-09
Adds axlearn gcp logs command.

markblee:


axlearn/cloud/gcp/jobs/logs.py
```diff
@@ -0,0 +1,123 @@
...
+from axlearn.cloud.gcp.config import default_project
+from axlearn.cloud.gcp.utils import catch_auth
+
+FLAGS = flags.FLAGS
+
+
+def _private_flags(flag_values: flags.FlagValues = FLAGS):
+    flags.DEFINE_string("name", None, "Job name.", flag_values=flag_values)
+    flags.DEFINE_alias("job_name", "name", flag_values=flag_values)
+    flags.DEFINE_integer("worker", 0, "Worker ID.", flag_values=flag_values)
```
ruomingp, line 25: Does the worker ID correspond to JAX `process_index`? If not, is there a way to find the logs of `process_index=0`?

ruomingp (approved):

axlearn/cloud/gcp/jobs/logs.py
hanzhi713: worker_id 0 isn't always process_index 0. To find the log of process_index 0, you have to download logs from all workers and search for the one that prints out it's process 0 in the step log.... and even worse, if there's a pod restart, the process <-> container order will be changed again.

markblee: Yeah, the current workaround goes through logged process ID (which can be filtered in cloud logging). Will sync with the infra team to track this item.

zhzhyi (approved):

axlearn/cloud/gcp/jobs/logs.py
zhzhyi: `jax.distributed.initialize` takes a process id parameter. Maybe we can manually set it to enforce the mapping between worker id and process id?

hanzhi713: ^^ This would be great.

ruomingp: Another possibility is for process 0 to export its worker ID somehow, for example, writing to GCS.

PR #748, opened 2024-10-11
Add new env variable to disable axlearn wrap.

ds-hwang:
AXLEARN_ENABLE_STACK_SUMMARY env variable controls the wrapping mechanism. By default, it's enabled as-is. Wraps consists of traceback, remat, and exception handling.

For example, AXLEARN_ENABLE_WRAP=false makes it to use axlearn with the VSCode debugger GUI, as the number of intermediate stack frames in wrapped module methods is gone. 
    "configurations": [
        {
	    ...
            "env": {
                "AXLEARN_ENABLE_STACK_SUMMARY": "false",
            },
        }
    ]

or create .env in ${workspaceFolder}
AXLEARN_ENABLE_STACK_SUMMARY=false

### DIFF
--- a/axlearn/common/module.py
+++ b/axlearn/common/module.py
@@ -534,6 +534,12 @@ def _wrap_method_with_auto_child_context(*, method_fn: Callable, method_name: st
     Callers of this function should either bind the returned function to an instance, e.g. using
     `partial(method_fn, instance)`, or supply an instance explicitly as the first arg.
     """
+    method_fn_in_context = functools.partial(
+        _call_method_in_context, method_fn=method_fn, method_name=method_name
+    )
+    if not traceback_util.is_stack_summary_enabled():
+        method_fn = functools.wraps(method_fn)(method_fn_in_context)
+        return method_fn
 
     @no_stack_summary
     @functools.wraps(method_fn)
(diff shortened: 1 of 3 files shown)


apghml (approved):

axlearn/common/module.py
```diff
@@ -535,6 +535,12 @@ def _wrap_method_with_auto_child_context(*, method_fn: Callable, method_name: st
     `partial(method_fn, instance)`, or supply an instance explicitly as the first arg.
     """
 
+    if not traceback_util.is_stack_summary_enabled():
+        method_fn = functools.wraps(method_fn)(
+            functools.partial(_call_method_in_context, method_fn=method_fn, method_name=method_name)
+        )
```
ruomingp, line 541: Nit: since we always `_call_method_in_context`, maybe we can move it outside the `if`:

```
method_fn_in_context = functools.partial(
    _call_method_in_context, method_fn=method_fn, method_name=method_name)
if not traceback_util.is_stack_summary_enabled():
    return functools.wraps(method_fn)(method_fn_in_context)
```

ruomingp (approved):

ds-hwang (commented):
Thank you for review! Could you take a look again?

ruomingp (approved):

markblee (approved):

PR #756, opened 2024-10-17
Alternative method of submitting jobs to DF Runner

remylouisew:
Running a dataflow job via the axlearn gcp vm start command is not necessary or intuative for non-apple users. Additionally, if you are already running your commands from a VM (e.g from a remote desktop), this process does not work. I recognize that there are still scenarios where you would want to launch your dataflow jobs from a VM, so rather than replacing that ability, I am adding an alternative. 

In order to submit jobs to the Dataflow runner without using ‘axlearn gcp vm start’, changes to the quoting behavior of [dataflow.py](http://dataflow.py/) are necessary. Unfortunately, there’s not an obviously elegant way to provide two versions of [dataflow.py](http://dataflow.py/), so if you can think of a better option, please let me know.

What I’ve done is this: [dataflow.py](http://dataflow.py/) remains as it was, and I’m adding [dataflow.alt.py](http://dataflow.alt.py/). In the directions, I’ve added instructions to replace the original module if the user wants to submit jobs to the dataflow runner without needing to create a VM.

Additional note: PR #711 makes a change to the quoting behavior that allows the submission of dataflow jobs without ‘axlearn gcp vm start’, however this fix will not work for any commands that include parameters that require quotes, e.g --dataflow_service_options='worker_accelerator=type:nvidia-tesla-t4;count:1;install-nvidia-driver'
The [dataflow.alt.py](http://dataflow.alt.py/) module DOES work these kinds of parameters.

### DIFF
--- a/axlearn/cloud/gcp/jobs/dataflow.alt.py
+++ b/axlearn/cloud/gcp/jobs/dataflow.alt.py
@@ -1,68 +1,57 @@
 # Copyright © 2023 Apple Inc.
 
-"""Runs Dataflow jobs locally or in GCP.
+"""Runs Dataflow jobs locally or in GCP ** without needing to submit the job from a VM **
+Users may encounter errors when they are trying to submit a job to the Dataflow runner using the primary dataflow.py script.
 
-If you're having issues when submitting jobs to the DataflowRunner, use dataflow.alt.py instead
+
+TO USE THIS MODULE, YOU MUST RENAME IT TO DATAFLOW.PY (and replace/rename the existing dataflow.py)
+
 
 The flow is:
 1. Builds the dataflow worker image.
 2. Runs the dataflow job either on dataflow (if runner is 'DataflowRunner', the default) or locally
-    (if runner is 'DirectRunner').
+   (if runner is 'DirectRunner').
+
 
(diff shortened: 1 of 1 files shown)


markblee (commented):
Thanks @remylouisew -- IIUC, the main issue is with flag escaping. Aside from finding a more generic fix for flag parsing, maybe we can add support for loading from flagfiles, which would avoid the duplicate code and manual renaming step. (It seems that flag escaping is painful for users anyway.) WDYT?

remylouisew:
The flag escaping has indeed been painful! I am not familiar with the process of loading from flagfiles, could you elaborate on how it would be implemented?

markblee:
> @markblee The flag escaping has indeed been painful! I am not familiar with the process of loading from flagfiles, could you elaborate on how it would be implemented?

`absl` has builtin support for flagfiles: https://abseil.io/docs/python/guides/flags#a-note-about---flagfile
So either we can accept a flagfile directly at the top-level `axlearn gcp dataflow` command, which is read and then flags forwarded to the user command; or the user script can use flagfiles directly. Let me know if additional clarifications are helpful.

ruomingp (commented):

PR #764, opened 2024-10-21
audio pillar starts to use einops

ds-hwang:
audio pillar starts to use einops


ruomingp (approved):

pyproject.toml
```diff
@@ -151,6 +151,7 @@ grain = [
 ]
 # Audio dependencies.
 audio = [
+    "einops",
```
markblee, line 154: Should we pin these deps?

ds-hwang: Done.

markblee (approved):

PR #776, opened 2024-10-23
Fix inconsistent paddings in conv layer.

ds-hwang:
The output paddings value is determined by the padding value at the anchor point in the window. If anchor is None, the default anchor point is the left time padding from conv padding config.

For examples with window=5,
1. "SAME" padding case,
    * padding=(2,2): (0 0 0 0 0)
    * anchor index is 2: (0 0 |0| 0 0)
```
                pad  |           | pad
    paddings:     0 0|0 0 0 1 1 1|1 1
                  |___0___|
                    |___0___|
                      |___0___|
                        |___1___|
                          |___1___|
                            |___1___|
```

2. "VALID" padding case,
    * padding=(0,0): (0 0 0 0 0)
    * anchor index is 0:  (|0| 0 0 0 0)
```
            pad |           | pad
    paddings:   |0 0 0 1 1 1|
                |0_______|
                  |0_______|
```

3. The legacy "VALID" padding case,
    * padding=(0,0) and anchor=4: (0 0 0 0 0)
    * anchor index is 4:  (0 0 0 0 |0|)
```
            pad |           | pad
    paddings:   |0 0 0 1 1 1|
                |________1|
                  |________1|
```

4. "CAUSAL" padding case,
    * padding=(4,0): (0 0 0 0 0)
    * anchor index is 4:  (0 0 0 0 |0|)
```
                pad      |           | pad
    paddings:     0 0 0 0|0 0 0 1 1 1|
                  |_______0|
                    |_______0|
                      |_______0|
                        |_______1|
                          |_______1|
                            |_______1|
```

5. "CAUSAL" with lookahead=1,
    * padding=(3, 1): (0 0 0 0 0)
    * anchor index is 3:  (0 0 0 |0| 0)
```
                pad    |           | pad
    paddings:     0 0 0|0 0 0 1 1 1|1
                  |_____0_|
                    |_____0_|
                      |_____0_|
                        |_____1_|
                          |_____1_|
                            |_____1_|
```

6. Arbitrary padding case,
    * padding=(2,1): (0 0 0 0 0)
    * anchor index is 2:  (0 0 |0| 0 0)
```
                pad  |           | pad
    paddings:     0 0|0 0 0 1 1 1|1
                  |___0___|
                    |___0___|
                      |___0___|
                        |___1___|
                          |___1___|
```

Noteworthy, this PR changes partial frame is valid frame. Otherwise, Conformer
loses some information at the end of audio. Especially, this fixes the issue
where the last token of an audio segment is marked as padding in situations
where text and audio are interleaved. So axlearn/audio/model_asr_test.py is
changed accordingly.

### DIFF
--- a/axlearn/audio/frontend.py
+++ b/axlearn/audio/frontend.py
@@ -219,6 +219,7 @@ class LogMelFrontend(BaseFrontend):
         # To identify padding frames, apply the framer to the input padding.
         # Consider a frame padded if it contains at least one padding sample.
         paddings = sliding_window(paddings, window_size=self._frame_size, stride=self._hop_size)
+        # TODO(dhwang2): Logically, a partial frame is a valid frame. Explore it later.
         paddings = jnp.max(paddings, axis=-1, keepdims=True)
         outputs = outputs * (1 - paddings)
         return dict(outputs=outputs[..., None], paddings=jnp.squeeze(paddings, axis=-1))
--- a/axlearn/audio/model_asr_test.py
+++ b/axlearn/audio/model_asr_test.py
@@ -130,15 +130,15 @@ class ASRModelTest(TestCase):
     """Tests ASRModel."""
 
     @parameterized.parameters(
-        (True, "forward", "ctc", 14.136186),
-        (False, "forward", "ctc", 14.811403),
+        (True, "forward", "ctc", 15.256454),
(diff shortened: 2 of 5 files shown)


ruomingp (approved):

ds-hwang:
Thank you for review!

markblee (approved):

PR #779, opened 2024-10-24
Disambiguate source_len / target_len

changlan:
Current TPU attention implementation use ambiguous naming source_len and target_len for query and kv lengths, but we typically follow the convention in which query length is target_len and vice versa. We rename them to q_seq_len and kv_seq_len to disambiguate.

### DIFF
--- a/axlearn/common/flash_attention/tpu_attention.py
+++ b/axlearn/common/flash_attention/tpu_attention.py
@@ -36,11 +36,11 @@ from axlearn.common.utils import Tensor
 
 
 def tpu_flash_attention(
-    query: Tensor,  # [batch_size, source_len, num_heads, head_dim]
-    key: Tensor,  # [batch_size, target_len, num_heads, head_dim]
-    value: Tensor,  # [batch_size, target_len, num_heads, head_dim]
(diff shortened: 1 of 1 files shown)


apghml:
Thanks, I think I've addressed this in https://github.pie.apple.com/foundation-models/axlearn/pull/847/files#diff-22479d8930dc7b23057c6517b633f9c03c00e805e1099f9c6f30fd9da2109f05

(Swapping `source_len` and `target_len` so they match the convention used elsewhere in the project.)

If we want to rename everything to `q_seq_len` and `kv_seq_len`, it might make sense to do that project-wide as a followup pr.

apghml (changes requested):
Thanks, I have a separate PR that includes a fix for this (swapping `source_len` and `target_len` so they match the convention elsewhere). If it's alright with you, could we do that?

If we want to standardize on `q_seq_len` and `kv_seq_len`, then maybe we can do that for the entire project at once in a followup PR?

apghml:
(Personally I do think the names `q_seq_len` and `kv_seq_len` make more sense longer term if we can do it project-wide.)

PR #780, opened 2024-10-25
Support Causal Convolution.

ds-hwang:
jax.lax.conv_general_dilated doesn’t natively support "CAUSAL" mode, so it implement causal padding convolution.

For example, with a window size of 3 and a stride of 2, padding="SAME" is applied as padding=((1,1),) as shown below, as keeping input and output length.
```
          pad|         |pad
paddings:   0|0 0 0 0 0|0
            |___|
                |___|
                    |___|
```

To implement padding="CAUSAL", we use padding=((2,0),) as follows.
```
           pad|         |pad
paddings:  0 0|0 0 0 0 0|
           |___|
               |___|
                   |___|
```

The first component is time and treated as "CAUSAL", while the remaining components are handled with "SAME" padding.

### DIFF
--- a/axlearn/common/layers.py
+++ b/axlearn/common/layers.py
@@ -56,9 +56,11 @@ from axlearn.common.utils import (
 )
 
 # The padding type for jax.lax.conv_general_dilated API. Either the strings ‘SAME’, or ‘VALID’, or
-# a sequence of n (low, high) integer pairs that give the padding to apply before and after each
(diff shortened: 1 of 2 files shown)


ds-hwang:
@ruomingp, @markblee Could you take a look?

ruomingp (approved):

markblee (approved):

ds-hwang:
Thank you for review!

PR #783, opened 2024-10-25
Add Goodput & Badput recording and monitoring support.

dipannita08:
This change adds the following:
 - [x] Upgrades to the latest [ml-goodput-measurement](https://github.com/AI-Hypercomputer/ml-goodput-measurement) library
 - [x] Integrates badput recording into the GoodputRecorder in AxLearn
 - [x]  Builds a Goodput monitor to configure and visualize Goodput and Badput using Tensorboard

Tested:
- [x] Functional end to end testing using fuji-test and fuji-7b ([example tensorboard instance](https://screenshot.googleplex.com/57zZoPvRB8WvFuY))
- [x]  Unit tests

image


axlearn/common/launch_trainer_main.py
```diff
@@ -4,18 +4,21 @@
...
 from absl import app, flags
 
-from axlearn.common import launch, launch_trainer, measurement
+from axlearn.common import launch, launch_trainer, measurement, monitoring
 from axlearn.common.config import config_for_function
 
 
 def main(_):
     measurement.initialize(flags.FLAGS)
+    monitoring.initialize(flags.FLAGS)
```
markblee, line 13: Do we need a separate module for this, or can we consolidate into the `measurement` interface?

markblee (commented):
Thanks!

axlearn/common/launch_trainer_main.py
jiya-zhang: I think it's a design choice. `measurement` part calls the GoodPut Recorder and uploads timestamps to Cloud Logging. `monitoring` part calls the GoodPut Monitor, which calculates goodput for a customized time interval and regularly updates Tensorboard file in GCS. These seem to have separate functionalities and sometimes users might need only `measurement` and not `monitoring`

markblee: Can the decision of whether to upload be handled via a config field instead of a separate module?

dipannita08: The config is useful either way to decide whether/how to upload the data. The intention to separate the monitoring module is to allow for the flexibility of using other types of monitors and allow for growing functionality of monitoring which is likely to depend on different things than recording. I think its useful to be able to configure these two entities separately because of the dataflow and the underlying systems that support the recording vs the calculating/monitoring, adding future flexibility and overall modularity. Having said that, its definitely possible to couple them in the same module, but I am not sure if its a better overall design. Let me know your thoughts though! :)

Ethanlm:
>Functional end to end testing using fuji-test and fuji-7b ([example tensorboard instance](https://screenshot.googleplex.com/57zZoPvRB8WvFuY))
I don't have access to this link. Can you provide an example that we can take a look>

axlearn/common/launch_trainer_main.py
markblee: Thanks for the discussion!

> allow for the flexibility of using other types of monitors and allow for growing functionality of monitoring which is likely to depend on different things than recording

Our config system supports composition of different configurable objects. For example, you can imagine that in the measurement module, we have a configurable object that is composed of a recorder and a monitor. A simple design might look like:

```
class GoodputMeasurement(...):
    @config_class
    class Config(...):
        recorder: Recorder.Config  # Configures recorder.
        monitor: Monitor.Config  # Configures monitor.
```

This allows us to interchange monitor implementations while keeping the same recorder, or vice versa, without introducing a separate Python module. (As it is, it seems tolerable to further simplify by extending `GoodputRecorder` with the `monitor` config directly, which may avoid additional pass-through APIs for `record` and `start_monitoring`. If the `Recorder` name is confusing, we can rename it.)

That said, I wonder what other monitors we plan to support for goodput. Are there alternatives to the `goodput_monitoring` that we plan to add that demands flag-level configurability? Or can we start with the simpler option of using the `goodput_monitoring` as the default option for now?

> because of the dataflow and the underlying systems that support the recording vs the calculating/monitoring

As it is currently implemented, we seem to want to instantiate and run both at the same time anyway. We can always add another Python module later if it grows too complex.

WDYT?

axlearn/common/trainer.py
```diff
@@ -800,13 +805,15 @@ def restore_checkpoint(self, restore_step: Optional[int] = None) -> Optional[int
                         step,
                         restore_input_iter,
                     )
+                self._maybe_record_event(measurement.Event.END_DATA_LOADING)
```
yiping-ma, line 808: If this line got skipped due to self.checkpointer.restore() failure, the data loading end time will be missing for this event, is it expected?

axlearn/common/launch_trainer_main.py
dipannita08: Thanks! I agree and apologize for the delay in response, I will make the changes accordingly!

axlearn/common/trainer.py
dipannita08: In that case, the badput would go into the "Unknown" bucket and not the "Data Loading" bucket. This is the expected behavior today, as there is currently no way to distinguish lost productive due to data loading if checkpointer.restore() failed since we don't really know if data loading even happened or not. We might however be able to put this into a "program runtime badput" if we can determine a better signal for data loading to isolate from infrastructure or hardware issues that could lead to "unknown" source of Badput as it stands today.

axlearn/cloud/gcp/measurement.py
```diff
@@ -49,10 +55,48 @@ def record(self, event: measurement.Event, *args, **kwargs):
...
+
+        if self._monitor:
+            self._monitor.start_goodput_uploader(*args, **kwargs)
+            logging.info("Started Goodput upload to Tensorboard in the background!")
+        else:
+            logging.log_first_n(
+                logging.WARNING,
+                "Goodput upload could not be started. Please check GoodputMonitor logs.",
+                1,
+            )
```
markblee, line 102: When do we expect reach this case?

axlearn/cloud/gcp/measurement_test.py
```diff
@@ -34,13 +36,46 @@ def test_from_flags(self, spec):
...
+        fv = flags.FlagValues()
+        measurement.define_flags(flag_values=fv)
+        fv.set_default(
+            "recorder_spec",
+            ["name=test-name", "upload_dir=/test/path/to/upload", "upload_interval=15"],
+        )
+        fv.mark_as_parsed()
+
+        recorder = GoodputRecorder.from_flags(fv)
+        recorder._monitor = None  # Ensure _monitor is initially None
```
markblee, line 63:
```suggestion
        self.assertIsNone(recorder._monitor)  # Ensure _monitor is initially None
```

axlearn/cloud/gcp/measurement.py
```diff
@@ -22,7 +23,11 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
         """Converts flags to a recorder.
 
         `fv.recorder_spec` will be interpreted as a list of `key=value` pairs; config names
-        corresponding to keys will be set to the corresponding values.
+        corresponding to keys will be set to the corresponding values. A GoodputRecorder can
+        additionally take in following Tensorboard configs in the recorder_spec:
+         - upload_dir: The directory to write Tensorboard data to.
+         - upload_interval: The time interval in seconds at which to query and upload data
```
markblee, line 29: Should we move the configs to this class for now? I.e., on L19:

```
@config_class
class Config(measurement.Recorder.Config):
    upload_interval: Required[str] = REQUIRED
```

axlearn/common/launch_trainer_main.py
```diff
@@ -13,6 +13,7 @@ def main(_):
     launch.setup()
     trainer_config = launch_trainer.get_trainer_config()
     trainer_config.set(recorder=config_for_function(lambda: measurement.global_recorder))
+    measurement.start_monitoring()
```
markblee, line 16: I suppose we must do this after `launch.setup` due to `jax.process_index`? If so, maybe we should add a comment to the goodput `start_monitoring` docstring that it assumes jax distributed init.

markblee:
Please feel free to "re-request review" when ready. Thanks!

axlearn/cloud/gcp/measurement.py
dipannita08: This could happen if there are internal errors (such as access errors) from GCP services such as Cloud Logging or Cloud Storage.

axlearn/cloud/gcp/measurement.py
```diff
@@ -49,10 +65,50 @@ def record(self, event: measurement.Event, *args, **kwargs):
...
             logging.log_first_n(
                 logging.WARNING,
                 "Ignoring unknown event %s",
                 1,
                 event,
             )
+
+    def start_monitoring(self, *args, **kwargs):
+        """
+        Instantiate ml-goodput-measurement's GoodputMonitor to asynchronously calculate
```
markblee, line 90: We actually follow the Google style docstrings 🙂 
```suggestion
        """Instantiate ml-goodput-measurement's GoodputMonitor to asynchronously calculate
```

axlearn/cloud/gcp/measurement.py
markblee: Got it. Since these cases are usually not expected, would it be worth `raise RuntimeError(...)` instead?

axlearn/common/trainer.py
```diff
@@ -324,6 +325,7 @@ def __init__(
                     model=self.model,
                     model_param_partition_specs=model_param_partition_specs,
                 )
+        self._maybe_record_event(measurement.Event.END_ACCELERATOR_INIT)
```
markblee, line 328: Can you clarify what accelerator init is supposed to capture? E.g., would `utils_spmd` where we call jax distributed init be more appropriate?

axlearn/common/trainer.py
```diff
@@ -847,6 +850,7 @@ def _prepare_training(self, prng_key: Tensor) -> bool:
             with fs.open(os.path.join(cfg.dir, "model_analysis.txt"), "w") as f:
                 f.write(model_analysis)
 
+        self._maybe_record_event(measurement.Event.END_TRAINING_PREPARATION)
```
markblee, line 853: What's usually considered part of "training preparation"? Should we count the jit compilation below as a potentially substantial part of it? What about the checkpoint restoration above?

axlearn/common/trainer.py
```diff
@@ -883,6 +887,7 @@ def restore_checkpoint(self, restore_step: Optional[int] = None) -> Optional[int
             restore_input_iter = cfg.save_input_iterator
             try:
                 # Try to restore with `input_iter`.
+                self._maybe_record_event(measurement.Event.START_DATA_LOADING)
```
markblee, line 890: Hm, is data loading be in relation to the input loading or the checkpoint or both? Here it seems only capturing the checkpoint restoration?

axlearn/cloud/gcp/measurement.py
dipannita08: My general thought around this, is to cause no disruptions to the workload due to Goodput measurement/monitoring (even if someone didn't setup GCP services for Goodput correctly, they could still have their workload running if all else goes well), we would still see fatal errors on the Goodput side and logs to figure out what happened if required. I'll let you decide if we want to make this fatal here, happy to change it per your suggestion.

axlearn/common/trainer.py
dipannita08: This is supposed to capture device related initialization such as device scanning, mesh initialization, device reinit/reset, security setup, initialization of pre-mapped buffers etc. You are right, jax distributed init should be included here.

I would lean on your team to update/re-position the record calls to locations that seems best fit for this codebase.

axlearn/common/trainer.py
dipannita08: This would include the time spent on the creation of checkpoint managers, checkpoint loading, running mesh and model optimizers etc.

The JIT compilation is currently computed algorithmically based on the entire timeline of events (other recorded logs) and is included in the "program startup" badput that is meant to measure the time spent on framework specific function transformations (such as JAX tracing), compilation tasks, runtime initialization etc.

For now, checkpoint restoration is not included in this bucket - the expectation is the next version of the library (v0.0.5) has more definitive recorder and calculator APIs for checkpoint save and restore badput.

axlearn/common/trainer.py
dipannita08: Data loading is in relation to the input data only. Checkpoint restore would be recorded and badput would be computed separately with changes coming in the next version of the Goodput package. Please feel free to update the location of the record calls specifically, for data loading, I wasn't sure where is the most appropriate place to put it.

Thanks for your help!

axlearn/cloud/gcp/measurement.py
dipannita08: Done!

axlearn/cloud/gcp/measurement.py
markblee: That sounds reasonable to me.

axlearn/common/trainer.py
markblee: Thanks @dipannita08 ! WDYT about removing the `trainer.py` changes for now, so that we can add them in a follow-up PR? This PR can focus on adding the scaffolding in `measurement.py`.

markblee (commented):
Ack.

axlearn/common/trainer.py
dipannita08: Sounds good, I'll circle back with you on those changes in a follow up PR.

axlearn/cloud/gcp/measurement.py
```diff
@@ -49,10 +65,51 @@ def record(self, event: measurement.Event, *args, **kwargs):
...
+
+        if self._monitor:
+            self._monitor.start_goodput_uploader(*args, **kwargs)
+            logging.info("Started Goodput upload to Tensorboard in the background!")
+        else:
+            logging.log_first_n(
+                logging.WARNING,
+                "Goodput upload could not be started. Please check GoodputMonitor logs.",
+                1,
+            )
```
ruomingp, line 115:
```suggestion
            )
            if not self._monitor:
                # This could happen if there are internal errors (such as access errors) from GCP services such as Cloud Logging or Cloud Storage.
                logging.log_first_n(
                    logging.WARNING,
                    "Goodput upload could not be started. Please check GoodputMonitor logs.",
                    1,
                )
        self._monitor.start_goodput_uploader(*args, **kwargs)
        logging.info("Started Goodput upload to Tensorboard in the background!")
```

axlearn/common/measurement.py
```diff
@@ -47,6 +59,10 @@ def record(self, event: Event, *args, **kwargs):
         """Records an event with the given name."""
         raise NotImplementedError(type(self))
 
+    def start_monitoring(self, **kwargs):
+        """Starts computing and uploading metrics at some configured interval in the background."""
+        raise NotImplementedError(type(self))
```
ruomingp, line 64: To avoid breaking other subclasses of `Recorder`.

```suggestion
        pass
```

axlearn/cloud/gcp/measurement.py
```diff
@@ -49,10 +65,50 @@ def record(self, event: measurement.Event, *args, **kwargs):
...
+            if not self._monitor:
+                # This could happen if there are internal errors (such as access errors) from GCP services such as Cloud Logging or Cloud Storage.
+                logging.log_first_n(
+                    logging.WARNING,
+                    "Goodput upload could not be started. Please check GoodputMonitor logs.",
+                    1,
+                )
+        self._monitor.start_goodput_uploader(*args, **kwargs)
+        logging.info("Started Goodput upload to Tensorboard in the background!")
```
markblee, line 114:
```suggestion
        if self._monitor:
            self._monitor.start_goodput_uploader(*args, **kwargs)
            logging.info("Started Goodput upload to Tensorboard in the background!")
        else:
            # This could happen if there are internal errors (such as access errors) from GCP services such as Cloud Logging or Cloud Storage.
            logging.log_first_n(
                logging.WARNING,
                "Goodput upload could not be started. Please check GoodputMonitor logs.",
                1,
            )
```
So that we check that `self._monitor` is valid before invoking `start_goodput_uploader`.

BTW, it's still unclear to me how `self._monitor` can be None after we construct the instance of `GoodputMonitor`. Are we missing a try/catch somewhere? Does the `__init__` method of `GoodputMonitor` raise an exception (that seems a bit unexpected)?

axlearn/cloud/gcp/measurement_test.py
```diff
@@ -34,13 +36,46 @@ def test_from_flags(self, spec):
...
+            ["name=test-name", "upload_dir=/test/path/to/upload", "upload_interval=15"],
+        )
         fv.mark_as_parsed()
 
         recorder = GoodputRecorder.from_flags(fv)
         recorder._recorder = mock.MagicMock()
         recorder.record(measurement.Event.START_JOB)
         self.assertTrue(recorder._recorder.record_job_start_time.called)
+
+    def test_start_monitoring(self):
```
markblee, line 53: Does this test the failure scenario?

axlearn/common/measurement.py
```diff
@@ -47,6 +59,10 @@ def record(self, event: Event, *args, **kwargs):
         """Records an event with the given name."""
         raise NotImplementedError(type(self))
 
+    def start_monitoring(self, **kwargs):
+        """Starts computing and uploading metrics at some configured interval in the background."""
+        pass
```
markblee, line 64: Let's `raise NotImplementedError(type(self))` and let subclasses decide whether to implement -- it should be fairly straightforward for a subclass to decide to not monitor, but we want the decision to be explicit.

axlearn/cloud/gcp/measurement.py
```diff
@@ -49,10 +65,50 @@ def record(self, event: measurement.Event, *args, **kwargs):
...
+            self._monitor = goodput_monitoring.GoodputMonitor(
+                job_name=cfg.name,
+                logger_name=f"goodput_logger_{cfg.name}",
+                tensorboard_dir=cfg.upload_dir,
+                upload_interval=int(cfg.upload_interval),
+                monitoring_enabled=(jax.process_index() == 0),
+                include_badput_breakdown=True,
+            )
+            if not self._monitor:
+                # This could happen if there are internal errors (such as access errors) from GCP services such as Cloud Logging or Cloud Storage.
```
markblee, line 107: BTW, I just triggered the CI, sorry for not doing so early. (I suspect lines like this will fail pylint for being too long.)

axlearn/common/measurement.py
ruomingp: Hmmm, it was originally raising `NotImplementedError(type(self))`, but I suggested to change to `pass` to avoid breaking subclasses upon axlearn bump.

Do we have tests to catch such breakage?

axlearn/cloud/gcp/measurement.py
dipannita08: I will re-run precomit checks.

axlearn/cloud/gcp/measurement_test.py
ruomingp: Looks like this is not addressed yet.

ruomingp (commented):

axlearn/common/measurement.py
markblee: I see, I think we only have the GoodputRecorder at the moment, so I think @dipannita08 's tests might cover it. (But let me know if I missed something.)

axlearn/cloud/gcp/measurement.py
dipannita08: PTAL, I've updated this code. GoodputMonitor's __init__ does not raise exceptions. We check if the monitor is not yet initialized, if so we do that, and then we can safely call start_goodput_uploader. The potential GCP errors we spoke about earlier can only happen after start_goodput_uploader is called. These only log the error and do not have any impact on the workload.

axlearn/common/measurement.py
dipannita08: Done, and yes.

axlearn/cloud/gcp/measurement_test.py
dipannita08: Done.

dipannita08:
Thanks! Can we re-trigger the CI tests?

dipannita08:
A test failure in axlearn/common/flash_attention/layer_test.py, does not seem related changes in this PR, could you help confirm? @markblee @ruomingp

markblee:
@dipannita08 This is the error I saw:
```
#22 1339.0 =================================== FAILURES ===================================
#22 1339.0 __________________________ UtilsTest.test_initialize2 __________________________
#22 1339.0 [gw0] linux -- Python 3.10.16 /opt/venv/bin/python
#22 1339.0 
#22 1339.0 self = <axlearn.common.measurement_test.UtilsTest testMethod=test_initialize2>
#22 1339.0 recorder_type = 'axlearn.experiments.testdata.axlearn_common_measurement_test.dummy_recorder:dummy_recorder'
#22 1339.0 expected = 'DummyRecorder'
#22 1339.0 
#22 1339.0     @parameterized.parameters(
#22 1339.0         # No-op if no recorder_type provided.
#22 1339.0         dict(
#22 1339.0             recorder_type=None,
#22 1339.0             expected=None,
#22 1339.0         ),
#22 1339.0         dict(
#22 1339.0             recorder_type="test",
#22 1339.0             expected="Mock",
#22 1339.0         ),
#22 1339.0         # Try initializing from another module.
#22 1339.0         dict(
#22 1339.0             recorder_type=(
#22 1339.0                 f"axlearn.experiments.testdata.{__name__.replace('.', '_')}.dummy_recorder:"
#22 1339.0                 "dummy_recorder"
#22 1339.0             ),
#22 1339.0             expected="DummyRecorder",
#22 1339.0         ),
#22 1339.0     )
#22 1339.0     def test_initialize(self, recorder_type, expected):
#22 1339.0         mock_recorder = mock.MagicMock()
#22 1339.0         measurement.register_recorder("test")(mock_recorder)
#22 1339.0     
#22 1339.0         fv = flags.FlagValues()
#22 1339.0         measurement.define_flags(flag_values=fv)
#22 1339.0         fv.set_default("recorder_type", recorder_type)
#22 1339.0         fv.mark_as_parsed()
#22 1339.0     
#22 1339.0         self.assertIsNone(measurement.global_recorder)
#22 1339.0 >       measurement.initialize(fv)
#22 1339.0 
#22 1339.0 axlearn/common/measurement_test.py:69: 
#22 1339.0 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
#22 1339.0 axlearn/common/measurement.py:117: in initialize
#22 1339.0     global_recorder = recorder_class.from_flags(fv=fv)
#22 1339.0 axlearn/experiments/testdata/axlearn_common_measurement_test/dummy_recorder.py:14: in from_flags
#22 1339.0     cls.default_config()
#22 1339.0 axlearn/common/config.py:394: in set
#22 1339.0     setattr(self, k, v)
#22 1339.0 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
#22 1339.0 
#22 1339.0 self = config_class(axlearn.common.measurement.Recorder.Config)(klass=<class 'axlearn.experiments.testdata.axlearn_common_measurement_test.dummy_recorder.DummyRecorder'>, name='dummy_recorder')
#22 1339.0 key = 'upload_dir', value = '/dummy/upload_dir'
#22 1339.0 
#22 1339.0     def wrapped_setattr(self, key: str, value):
#22 1339.0         if key.startswith("__"):
#22 1339.0             self.__dict__[key] = value
#22 1339.0         else:
#22 1339.0             if key not in _attr_fields_dict_cache(type(self)):
#22 1339.0 >               raise UnknownFieldError(self._key_error_string(key))
#22 1339.0 E               axlearn.common.config.UnknownFieldError: upload_dir (keys are ['klass', 'name'])
#22 1339.0 
#22 1339.0 axlearn/common/config.py:625: UnknownFieldError
```

dipannita08:
> @dipannita08 This is the error I saw:
> 
> ```
> #22 1339.0 =================================== FAILURES ===================================
> #22 1339.0 __________________________ UtilsTest.test_initialize2 __________________________
> #22 1339.0 [gw0] linux -- Python 3.10.16 /opt/venv/bin/python
> #22 1339.0 
> #22 1339.0 self = <axlearn.common.measurement_test.UtilsTest testMethod=test_initialize2>
> #22 1339.0 recorder_type = 'axlearn.experiments.testdata.axlearn_common_measurement_test.dummy_recorder:dummy_recorder'
> #22 1339.0 expected = 'DummyRecorder'
> #22 1339.0 
> #22 1339.0     @parameterized.parameters(
> #22 1339.0         # No-op if no recorder_type provided.
> #22 1339.0         dict(
> #22 1339.0             recorder_type=None,
> #22 1339.0             expected=None,
> #22 1339.0         ),
> #22 1339.0         dict(
> #22 1339.0             recorder_type="test",
> #22 1339.0             expected="Mock",
> #22 1339.0         ),
> #22 1339.0         # Try initializing from another module.
> #22 1339.0         dict(
> #22 1339.0             recorder_type=(
> #22 1339.0                 f"axlearn.experiments.testdata.{__name__.replace('.', '_')}.dummy_recorder:"
> #22 1339.0                 "dummy_recorder"
> #22 1339.0             ),
> #22 1339.0             expected="DummyRecorder",
> #22 1339.0         ),
> #22 1339.0     )
> #22 1339.0     def test_initialize(self, recorder_type, expected):
> #22 1339.0         mock_recorder = mock.MagicMock()
> #22 1339.0         measurement.register_recorder("test")(mock_recorder)
> #22 1339.0     
> #22 1339.0         fv = flags.FlagValues()
> #22 1339.0         measurement.define_flags(flag_values=fv)
> #22 1339.0         fv.set_default("recorder_type", recorder_type)
> #22 1339.0         fv.mark_as_parsed()
> #22 1339.0     
> #22 1339.0         self.assertIsNone(measurement.global_recorder)
> #22 1339.0 >       measurement.initialize(fv)
> #22 1339.0 
> #22 1339.0 axlearn/common/measurement_test.py:69: 
> #22 1339.0 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> #22 1339.0 axlearn/common/measurement.py:117: in initialize
> #22 1339.0     global_recorder = recorder_class.from_flags(fv=fv)
> #22 1339.0 axlearn/experiments/testdata/axlearn_common_measurement_test/dummy_recorder.py:14: in from_flags
> #22 1339.0     cls.default_config()
> #22 1339.0 axlearn/common/config.py:394: in set
> #22 1339.0     setattr(self, k, v)
> #22 1339.0 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> #22 1339.0 
> #22 1339.0 self = config_class(axlearn.common.measurement.Recorder.Config)(klass=<class 'axlearn.experiments.testdata.axlearn_common_measurement_test.dummy_recorder.DummyRecorder'>, name='dummy_recorder')
> #22 1339.0 key = 'upload_dir', value = '/dummy/upload_dir'
> #22 1339.0 
> #22 1339.0     def wrapped_setattr(self, key: str, value):
> #22 1339.0         if key.startswith("__"):
> #22 1339.0             self.__dict__[key] = value
> #22 1339.0         else:
> #22 1339.0             if key not in _attr_fields_dict_cache(type(self)):
> #22 1339.0 >               raise UnknownFieldError(self._key_error_string(key))
> #22 1339.0 E               axlearn.common.config.UnknownFieldError: upload_dir (keys are ['klass', 'name'])
> #22 1339.0 
> #22 1339.0 axlearn/common/config.py:625: UnknownFieldError
> ```

Missed one file! Updated and re-ran all the tests, should be good now 🤞

dipannita08:
@markblee, thanks! Could you help with the approval & merge?

axlearn/cloud/gcp/measurement.py
```diff
@@ -49,10 +65,47 @@ def record(self, event: measurement.Event, *args, **kwargs):
             self._recorder.record_job_end_time(*args, **kwargs)
         elif event == measurement.Event.START_STEP:
             self._recorder.record_step_start_time(*args, **kwargs)
+        elif event == measurement.Event.START_ACCELERATOR_INIT:
+            self._recorder.record_tpu_init_start_time(*args, **kwargs)
```
markblee, line 69: OOI, is there anything here specific to TPUs or can we use the same API for GPUs on GCP?

markblee (approved):
Thanks @dipannita08 ! LGTM. Will do some additional testing on my side before merging.

axlearn/cloud/gcp/measurement.py
dipannita08: Its not specific to TPUs, since this badput bucket is computed based on recorded markers. The API name needs to be updated to reflect this and whenever this is done, we would need to refactor this piece of code as well.

axlearn/common/measurement.py
```diff
@@ -120,3 +136,16 @@ def record_event(event: Event):
         logging.log_first_n(logging.INFO, "No recorder configured, ignoring events.", 1)
     else:
         global_recorder.record(event)
+
+
+def start_monitoring():
+    """Begins monitoring events as per global monitor functionality."""
+    if global_recorder is None:
+        logging.log_first_n(
```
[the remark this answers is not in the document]
dipannita08: Since this is a no-op, I'll leave this for now and address it the next PR. Expecting to complete integration with package v5 in the next one as well. [Here](https://b.corp.google.com/issues/393595286) is the tracking bug.

PR #786, opened 2024-10-28
Add TPU Monitoring for faster hang detection

kelvin-zou:
Add a TPU monitoring plugin, to detect the idleness of the TPU host, via querying from TPU device metrics. 
Currently we set a long watchdog timeout (30min+) to avoid false positive in troubleshooting the hang, and with more device info we can have a more proactive detection and thus help catch issues earlier.


axlearn/common/trainer.py
```diff
@@ -376,22 +382,46 @@ def _stop_watchdog(self):
...
+                            "seconds and the host is idle.\n"
+                            "NOTE: this is not an error message, but meant to help debugging "
+                            "in case the trainer is stuck.\n"
+                            "Threads:\n%s",
+                            time_elapsed_in_sec_since_last_check,
+                            "\n".join(
+                                itertools.chain.from_iterable(thread_stack_traces()),
+                            ),
+                        )
+                        # TODO(kelvin-zou): maybe crash the program here?
```
Ethanlm, line 410: +1 on crashing the program to avoid the waste. But also making it configurable so that a job can configure how often the health check should run, and when to crash (like if idle for x hours)

ruomingp (approved):

axlearn/common/trainer.py
nstogner: Crashing the program here would trigger a JobSet restart (all Pods would be recreated). Are there some failure modes that this would address?

kelvin-zou: @nstogner  atm it is only logging faster, so that our log parser can catch it earlier and report to GCP. let's chat offline about how to crash here, the logic is much more complex than an `raise Error`.

Ethanlm: > Crashing the program here would trigger a JobSet restart (all Pods would be recreated). Are there some failure modes that this would address?


When a job gets stuck, restarting the jobset can sometimes help (but not always)

PR #787, opened 2024-10-28
Add multislice XLA flag to improve tf summary metric induced latency.

tgunter:
Improves multi-slice step time in the case of multiple slices and heavy summary logging during the forwards pass.


markblee (approved):

samos123:
Note in certain cases this flag may introduce worse performance (e.g. Trillium with optimizer state host offload and 405B model). I'm following up on this internally and will respond back once I learn more about what this flag actually does.

Sharing early with you in case you see very long all-gathers, the root cause could be this flag.

PR #788, opened 2024-10-28
Remove init workaround

hanzhi713:
As documented in the comment by @markblee, the commit is already included in jax, and we can safely remove this workaround.

### DIFF
--- a/axlearn/common/utils_spmd.py
+++ b/axlearn/common/utils_spmd.py
@@ -3,7 +3,6 @@
 """SPMD related utils."""
 
 import logging
-import time
 from typing import Optional
 
 import jax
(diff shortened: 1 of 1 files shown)


axlearn/common/utils_spmd.py
```diff
@@ -89,31 +88,5 @@ def setup(
                 process_id=process_id,
             )
 
-        # Ensure that coordinator initialization respects initialization_timeout.
-        # The current jax version hardcodes the number of attempts to discover coordinator address:
-        # https://github.com/google/jax/blob/33e1a96204bf88f76b9f8d982cb2fe92ebcf0151/jax/_src/clusters/cloud_tpu_cluster.py#L110-L125
-        # TODO(markblee): Remove this once we update jax to include:
-        # https://github.com/google/jax/commit/c5869feb92a175ce40b4e5a897f505e97bcc610b.
-        if initialization_timeout is not None:
```
kelvinzou, line 97: add a comment in flag, stating it is being deprecated?

markblee: `initialization_timeout` should still be used, but handled by `jax.distributed.initialize` directly.

markblee (approved):

PR #789, opened 2024-10-28
Weight only offload

hanzhi713:
This PR is not meant to be merged. It's for Google folks to debug and optimize optimizer offloading.

### DIFF
--- a/axlearn/experiments/text/gpt/fuji.py
+++ b/axlearn/experiments/text/gpt/fuji.py
@@ -154,7 +154,7 @@ def get_trainer_kwargs(
     elif model_size == "7B":
         trainer_kwargs = dict(
             model_kwargs=dict(
-                num_layers=32,
+                num_layers=16,
                 hidden_dim=128 * 32,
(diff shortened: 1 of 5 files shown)


axlearn/experiments/text/gpt/fuji.py
```diff
@@ -363,7 +364,7 @@ def model_config(
         hidden_dim=hidden_dim,
         num_heads=num_heads,
         vocab_size=vocab_size,
-        stack_cfg=stack_cfg if stack_cfg is not None else RepeatedTransformerLayer.default_config(),
+        stack_cfg=stack_cfg if stack_cfg is not None else StackedTransformerLayer.default_config(),
```
samos123, line 367: does weight only offload work with RepeatedTransformerLayer? @hanzhi713

samos123: it works, using StackedTransformerLayer just takes a lot of time to compile. So for 405b it wasn't feasible to use it.

PR #791, opened 2024-10-29
Make RepeatedConformerLayer's repeat field configurable.

ds-hwang:
Streaming Conformer wants to subclass RepeatedConformerLayer.


ds-hwang:
Could you review? From 853

markblee (approved):

PR #797, opened 2024-10-30
Introduce @nowrap annotaion.

ds-hwang:
It's similar to Flax `@nowarp` annotaion.
https://flax-linen.readthedocs.io/en/latest/api_reference/flax.linen/decorators.html#flax.linen.nowrap

Marks the specified module method as one that doesn't need to be wrapped.

Methods decorated with `@nowrap` are helper methods that don't require wrapping, and
`_methods_to_wrap_for_auto_child_context()` will not return them.

This is especially useful in cases where a public method (i.e., one that is not explicitly
prefixed with `_`) does not need an invocation context, such as methods that do not attempt
to access state or PRNG keys.

For instance::
```
        >>> from axlearn.common import module
        >>> class Foo(module.Module):
        ...   @module.nowrap
        ...   def init_states(self, batch_size: int):
        ...     return dict(time_step=jnp.zeros(batch_size, dtype=jnp.int32))
```

### DIFF
--- a/axlearn/common/layers.py
+++ b/axlearn/common/layers.py
@@ -40,7 +40,7 @@ from axlearn.common.config import (
 from axlearn.common.loss import binary_cross_entropy, categorical_hinge_loss, cross_entropy
 from axlearn.common.metrics import WeightedScalar
 from axlearn.common.metrics_classification import precision_recall_f_score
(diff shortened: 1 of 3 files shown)


ds-hwang:
Could you take a look? from #861

markblee (approved):

PR #799, opened 2024-10-31
Make some convolution functions public.

ds-hwang:
It includes conv_dilate_window, conv_output_shape and compute_conv_paddings, because a downstream project use them.


ds-hwang:
Could you take a look this simple renaming PR? from 867

markblee (approved):

PR #801, opened 2024-11-01
Correct CAUSAL padding in conv when stride > 1.

ds-hwang:
The formula for CAUSAL padding is `(window - stride, stride - 1)`. Currently, CAUSAL padding uses `(window - 1, 0)`, which only works when the stride is 1.

As a result, the streaming subsampler is currently losing the last frame. This approach provides a general logic for CAUSAL cases, and after this PR, the frontend will also reuse this implementation.

For example, with window=15 and stride=6, padding is (9, 5). Below are examples illustrating how input paddings are transformed into output paddings across different scenarios.
```
     left_pad         |         input paddings              -> outputs paddings
 1) |1 1 1|1 1 1|1 1 1|0 0 0|0 0 0|0 0 0|0 0 0|0 0 0|0 0 0| -> 0 0 0
 2) |1 1 1|1 1 1|1 1 1|1 0 0|0 0 0|0 0 0|0 0 0|0 0 0|0 0 0| -> 1 0 0
 3) |1 1 1|1 1 1|1 1 1|1 1 1|1 1 1|0 0 0|0 0 0|0 0 0|0 0 0| -> 1 0 0
 4) |1 1 1|1 1 1|1 1 1|1 1 1|1 1 1|1 0 0|0 0 0|0 0 0|0 0 0| -> 1 1 0
 5) |1 1 1|1 1 1|1 1 1|1 1 1|1 1 1|1 1 1|1 1 1|1 1 1|1 1 1| -> 1 1 1
 6) |1 1 1|1 1 1|1 1 1|0 1 1|1 1 1|1 1 1|1 1 1|1 1 1|1 1 1| -> 0 1 1
 7) |1 1 1|1 1 1|1 1 1|0 0 0|0 0 0|1 1 1|1 1 1|1 1 1|1 1 1| -> 0 1 1
 8) |1 1 1|1 1 1|1 1 1|0 0 0|0 0 0|0 1 1|1 1 1|1 1 1|1 1 1| -> 0 0 1
     |_________________^_________|
                 |_________________^_________|
                             |_________________^_________|
```
Let's take a closer look at case 7). In case 7), the first window component fully
covers all 0s, so the first component of the output padding should be the last
0 component, meaning the second component is 1.

In case 8), however, the first window component does not cover all 0s, so the next component should also be 0. If the second component were 1, information from the last partial window of the input would be lost — which is the current situation in the code, which this PR wants to fix.

In general, the anchor point should be the next position after the right edge of the previous window. Since the anchor is defined by the left pad, `left_pad = window - stride`, and `right_pad = (window - 1) - left_pad`, simplifying to `right_pad = stride - 1`.

Note: CAUSAL was introduced by me three days ago, and currently, I’m the only one using it in the streaming encoder. Therefore, logic changes without backward compatibility are acceptable.

### DIFF
--- a/axlearn/common/layers.py
+++ b/axlearn/common/layers.py
@@ -819,36 +819,45 @@ def conv_dilate_window(*, window: Sequence[int], dilation: Optional[Sequence[int
 # Copied from subroutine in jax.lax.reduce_window.
 # Extend lax.padtype_to_pads for CAUSAL.
 def conv_explicit_padding(
-    *, window: Sequence[int], padding: ConvPaddingType, dilation: Optional[Sequence[int]] = None
+    *,
+    window: Sequence[int],
+    strides: Sequence[int],
+    padding: ConvPaddingType,
+    dilation: Optional[Sequence[int]] = None,
 ) -> ConvPaddingType:
-    """Convert str padding to tuple padding.
+    """Returns the explicit padding for "SAME", "VALID", and "CAUSAL" modes.
 
-    For example, with a window size of 3 and a stride of 2, padding="SAME" is applied as
-    padding=((1,1),) as shown below, as keeping input and output length.
-            pad  |         |pad
-    paddings:   0|0 0 0 0 0|0
-                |___|
-                    |___|
-                        |___|
+    Each mode follows the formulas below:
+    * SAME: (pad_total//2, pad_total - pad_total//2)
+    * VALID: (0, 0)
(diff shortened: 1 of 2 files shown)


ds-hwang:
Hi @ruomingp, @markblee , could you take a look? @ruomingp reviewed it in #868

ruomingp (approved):

ds-hwang:
Thank you for review!

PR #807, opened 2024-11-04
Optimizing the sliding window in the frontend.

ds-hwang:
Creating an index at the sample level and using advanced indexing is quite costly—processing 100 seconds of audio at 24kHz results in 2.4 million samples.

To optimize this, we can index chunk-wise instead. For 100 seconds of audio at 24kHz, with a 10ms hop and a 25ms window, each chunk is 5ms, or 1200 samples. This approach reduces the indexing computation by a factor of 1200.

The newly added `frontend_benchmark.py` shows that a batch of 32 with 10-minute audio has improved from 1.8 seconds to 0.07 seconds on TPUv4.

**In addition**, rename `sliding_window` to `frame` according to tf.signal.frame.

**In addition**, there is a logic change. Currently, the frontend cuts off the last partial frame. After this PR, it keeps the last partial frame, which may increase the output length by 1. While the output tensor shape changes, the output values do not, because the paddings still treat the partial frame as padding. As a result, there will be no change in the neural network's output. ASRModelTest.test_asr_model demonstrates this behavior clearly. When is_training=False, there is no change in the loss value. However, when is_training=True, the loss changes due to dropout. Even with the same seed, different input shapes result in different returned values.

### DIFF
--- a/axlearn/audio/encoder_asr_test.py
+++ b/axlearn/audio/encoder_asr_test.py
@@ -7,6 +7,7 @@ import pytest
 from absl.testing import parameterized
 from jax import numpy as jnp
 
+from axlearn.audio import frontend_utils
 from axlearn.audio.encoder_asr import ASREncoder, SpeechContextNetwork, SpeechFeatureLayer
 from axlearn.audio.test_utils import fake_audio
 from axlearn.common.module import functional as F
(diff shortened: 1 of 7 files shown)


ds-hwang:
@ruomingp, @markblee could you review this PR? It blocks several people.

ruomingp (approved):

PR #820, opened 2024-11-07
Integrate Orbax's emergency checkpoint.

hanzhi713:


axlearn/common/checkpointer_orbax_emergency.py
```diff
@@ -0,0 +1,796 @@
...
+        return info
+    finally:
+        # Restore previous platform settings.
+        if platform != "":
+            os.environ["JAX_PLATFORMS"] = platform
+        else:
+            del os.environ["JAX_PLATFORMS"]
+
+
+class OrbaxEmergencyCheckpointer(BaseCheckpointer):
```
ruomingp, line 476: How does this overlap with https://github.com/apple/axlearn/blob/140a18f33d10178dc695469e36d39d7879bd1ed4/axlearn/common/checkpointer_orbax.py#L169?

hanzhi713: The orbax regular checkpointer saves 1 gcs checkpoint for n slices per save. This checkpointer saves n-1 checkpoints to a local path (usually a ramdisk), and also 1 checkpoint to gcs.

hanzhi713: Unfortunately it's not possible to share code between the two implementations.

ruomingp: Thanks for the clarification.
- How should users use them? Should they use both of them or one of them? How should they pick?
- Or can we replace OrbaxCheckpointer with this class?

hanzhi713: Added a comment to clarify this:
```
    This checkpointer is designed to improve the goodput of large multi-slice training jobs that
    use data-parallelism across slices. At least two data-parallel slices are required. For other
    use cases where this is not applicable or ultimate goodput is not required, please use
    `OrbaxCheckpointer`.

```

ruomingp: OK. So we will need to maintain two separate classes going forward? That doesn't seem ideal.

Could you elaborate on why we need two completely separate implementations?

hanzhi713: I think the fundamental reason is that orbax emergency checkpointer doesn't support non-tensor states, so we have to have a different checkpointer wrapper around it.

ruomingp: I see. What prevents us from using the emergency checkpointer for all jobs?

hanzhi713: It requires two data-parallel slices

hanzhi713: Not just data-parallelism, but data-parallel slices, i.e. it has to be multi-slice training.

ruomingp (commented):
Thanks for the explanation of the constraints. I wonder what the long term plan is. 

Is the emergency checkpointer a temporary solution that will eventually be dropped when the main Orbax checkpointer supports in-memory checkpoints?

Or will we keep maintaining two separate checkpointers, with potentially incompatible ckpt layouts?

hanzhi713:
> Is the emergency checkpointer a temporary solution that will eventually be dropped when the main Orbax checkpointer supports in-memory checkpoints?

I don't know if Google has such a plan. The orbax in-memory checkpointer actually uses the orbax regular checkpointer under the hood, which might be required by design/by nature of the problem that it solves. 

> Or will we keep maintaining two separate checkpointers, with potentially incompatible ckpt layouts?

Since in-memory checkpointer uses the regular orbax checkpointer under the hood, the tensor state in the persistent checkpoint (i.e. the one stored to gcs) can be loaded by `OrbaxStateBuilder` (see https://github.com/apple/axlearn/pull/866). Therefore, we can say that the checkpoints are compatible for **eval and inference** purposes. It's just that the **training** checkpoint will be incompatible, meaning that `OrbaxEmergencyCheckpointer`'s checkpoint cannot be loaded by `OrbaxCheckpointer`.

hanzhi713:
I think in the long term, it's probably possible to unify the checkpoint structure between the two checkpointer (regular and in-memory), but it's unknown whether we can unify the codepath.

axlearn/common/checkpointer_orbax_emergency.py
ruomingp: IIUC, only the local checkpoints require multiple slices, since we will need to restore from another slice upon a slice restart. Could we disable local checkpoints when num_slices=1? This way we always use the emergency checkpointer consistently.

axlearn/common/checkpointer_orbax_emergency.py
```diff
@@ -0,0 +1,814 @@
...
+
+        start_t = time.perf_counter()
+        state_with_tensors = jax.tree.map(
+            lambda x: x if isinstance(x, (Tensor, TensorSpec)) else None, state
+        )
+        # Note that save() waits for prior serialization to finish.
+        self._non_tensor_manager.save(step=step, state=state)
+        self._get_tensor_manager(state_with_tensors).save(
+            step=step, args=ocp.args.PyTreeSave(item=state_with_tensors)
+        )
```
ruomingp, line 748: How do we mark the completion of a checkpoint? It should happen only when both tensor and non-tensor states are saved. How is this ensured?

Please add a comment.

axlearn/common/checkpointer_orbax_emergency.py
```diff
@@ -0,0 +1,814 @@
...
+    persistent checkpoints. If the checkpoint is local, the slice on which that checkpoint is
+    stored will read the checkpoint and broadcast the read values to other slices. Since local
+    checkpoints are scattered across different hosts, the process id, which determines the shard id
+    of locally stored shards, must stay the same for nodes in the healthy replicas to guarantee a
+    correct restore. We provide an utility function `get_consistent_proc_info` that returns the
+    process id and global coordinator address. They must be passed to `jax.distributed.initialize`.
+
+    However, the above procedure doesn't apply to some non-tensor states such as data iterators.
+    Data iterators are unique across jax processes, and thus cannot be stored on nodes. Orbax
+    emergency checkpointer doesn't support non-tensor states. Therefore, we reuse axlearn
```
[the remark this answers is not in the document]
hanzhi713: I know it's doable in the beginning. We'll need to talk to Orbax for their plan to support this feature.

axlearn/common/checkpointer_orbax_emergency.py
hanzhi713: There's no special marker for completion of both. There are only markers for each of them individually. So, during restore, we look for both of them only load a specific step when both marker exists.

axlearn/common/checkpointer_orbax_emergency.py
hanzhi713: Could be an idea.

hanzhi713: Still needs support from orbax though.

hanzhi713:
@ruomingp I guess my question now is what's the plan here. Should we wait for Orbax's support for non tensor states and unified checkpointer API?

 I personally don't see in-mem ckpt as a life-changing feature, so waiting could be an viable option. Alternatively, we can proceed with this PR and make changes later.

axlearn/common/checkpointer_orbax_emergency.py
ruomingp: Thanks. Can you add a comment here and point to where "we look for both of them only load a specific step when both marker exists"?

Do we have testing for incomplete checkpoints?

axlearn/common/checkpointer_orbax_emergency.py
```diff
@@ -0,0 +1,814 @@
...
+
+        # For meaning of these options, refer to
+        # https://github.com/google/orbax/blob/95be2c021bc8cbf4badd83a053ff57b7a9f9b314/checkpoint/orbax/checkpoint/experimental/emergency/checkpoint_manager.py#L277
+        self._tensor_manager = oecp.CheckpointManager(
+            self._local_dir,
+            persistent_directory=os.path.join(cfg.dir, self._TENSORS_PREFIX),
+            global_mesh=thread_resources.env.physical_mesh,
+            abstract_state=self._get_abstract_state(state_with_tensors),
+            options=oecp.CheckpointManagerOptions(
+                local=oecp.LocalCheckpointOptions(
```
ruomingp, line 709: Shall we expose `oecp.LocalCheckpointOptions` to users as `Config.local_checkpoint_options`? User can set it to None to disable local checkpoints.

We can provide a helper function for users to construct `should_save_fn` from their policy.

axlearn/common/checkpointer_orbax_emergency.py
```diff
@@ -0,0 +1,814 @@
...
+        start_t = time.perf_counter()
+        cfg: OrbaxEmergencyCheckpointer.Config = self.config
+        state_with_tensors = jax.tree.map(
+            lambda x: x if isinstance(x, (Tensor, TensorSpec)) else None, state
+        )
+        tensor_manager = self._get_tensor_manager(state_with_tensors)
+        if step is None:
+            # Find the intersection of the checkpoint steps managed by tensor and non-tensor
+            # manager, and then use the latest step in the intersection for restore. `all_steps`
+            # from tensor manager contains both local and persistent checkpoints.
```
ruomingp, line 772: Consider refactoring this logic to a separate function so that it can be tested directly?

axlearn/common/checkpointer_orbax_emergency.py
ruomingp: I see. Right now it assumes local checkpoints are always used: https://github.com/google/orbax/blob/6e80ecc27581a413b1a481d4740e61df7316a4f4/checkpoint/orbax/checkpoint/experimental/emergency/checkpoint_manager.py#L695-L709.

Could you raise this request to the Orbax team and link to the issue?

ruomingp:
> @ruomingp I guess my question now is what's the plan here. Should we wait for Orbax's support for non tensor states and unified checkpointer API?
> 
> I personally don't see in-mem ckpt as a life-changing feature, so waiting could be an viable option. Alternatively, we can proceed with this PR and make changes later.

There's no best solution. I see three possibilities:
1 Wait until Orbax supports a Checkpointer that works for both single-slice and multi-slice settings with a unified persistent layout;
1 Merge this PR for testing, but be willing to discard checkpoints if we decide to change the persistent layout later;
1 Merge this PR and commit to unify the checkpointers and build tools to convert checkpoints for users;

I do not think we want to maintain two Orbax checkpointers in the longer run, especially with incompatible layouts. 

WDYT?

axlearn/common/checkpointer_orbax_emergency.py
hanzhi713: I can add a testcase.

kelvin-zou:
> > @ruomingp I guess my question now is what's the plan here. Should we wait for Orbax's support for non tensor states and unified checkpointer API?
> > I personally don't see in-mem ckpt as a life-changing feature, so waiting could be an viable option. Alternatively, we can proceed with this PR and make changes later.
> 
> There's no best solution. I see three possibilities: 1 Wait until Orbax supports a Checkpointer that works for both single-slice and multi-slice settings with a unified persistent layout; 1 Merge this PR for testing, but be willing to discard checkpoints if we decide to change the persistent layout later; 1 Merge this PR and commit to unify the checkpointers and build tools to convert checkpoints for users;
> 
> I do not think we want to maintain two Orbax checkpointers in the longer run, especially with incompatible layouts.
> 
> WDYT?

I think maybe waiting may not be the best idea, given that Orbax dev is often delayed, it has been for over a year since last time Anthropic raised this in-mem checkpointing feature request. 

I think both options below works for me
1. Merge this PR for testing, but be willing to discard checkpoints if we decide to change the persistent layout later
2. Merge this PR and commit to unify the checkpointers and build tools to convert checkpoints for users;

Maybe let's merge it for testing (we need quite sometime to test throughout), and then wait for a quarter and see if Orbax can really make it happen and available to our users? if so we can discard the tf iterator handling and use the upstream unified one. If not, then we can commit to unification ourselves?

axlearn/common/checkpointer_orbax_emergency.py
hanzhi713: It's already exposed, via `local_keep_last_n` and `local_save_policy`

axlearn/common/checkpointer_orbax_emergency.py
hanzhi713: Done and added a test.

hanzhi713:
@kelvin-zou SGTM.

axlearn/common/checkpointer_orbax_emergency.py
hanzhi713: Done.

hanzhi713:
@ruomingp Could you please approve this PR if it looks good?

ruomingp (approved):
Approving to unblock experiments.

axlearn/common/checkpointer_orbax_emergency.py
```diff
@@ -0,0 +1,829 @@
...
+        if platform != "":
+            os.environ["JAX_PLATFORMS"] = platform
+        else:
+            del os.environ["JAX_PLATFORMS"]
+
+
+class OrbaxEmergencyCheckpointer(BaseCheckpointer):
+    """Checkpointer implementation that uses Orbax emergency checkpoint.
+
+    ## Summary:
```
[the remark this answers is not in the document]
hanzhi713: Done.

ruomingp (approved):

axlearn/common/checkpointer_orbax_emergency.py
```diff
@@ -0,0 +1,832 @@
...
+            return wrapped_save_policy(step=step, evaler_summaries=self._eval_summaries)
+
+        # For meaning of these options, refer to
+        # https://github.com/google/orbax/blob/95be2c021bc8cbf4badd83a053ff57b7a9f9b314/checkpoint/orbax/checkpoint/experimental/emergency/checkpoint_manager.py#L277
+        self._tensor_manager = oecp.CheckpointManager(
+            self._local_dir,
+            persistent_directory=os.path.join(cfg.dir, self._TENSORS_PREFIX),
+            global_mesh=thread_resources.env.physical_mesh,
+            abstract_state=self._get_abstract_state(state_with_tensors),
+            options=oecp.CheckpointManagerOptions(
```
apivovarov, line 713: `local_state_handler` is missing here.
CheckpointManager::__init__ does not have default value for local_state_handler
```
  def __init__(
      self,
      local_directory: epath.PathLike,
      persistent_directory: epath.PathLike,
      global_mesh: jax.sharding.Mesh,
      abstract_state: PyTree,  # a single PyTree describing the state structure
      # TODO: b/330585086 - Support arbitrary items beyond state. We will have
      # to evaluate whether arbitrary items can be a good fit for local
      # checkpointing, given restore+broadcast requirements.
      local_state_handler: CheckpointHandler,
      *,
      options: Optional[CheckpointManagerOptions] = None,
      metadata: Optional[dict[str, Any]] = None,
      logger: Optional[abstract_logger.AbstractLogger] = None,
  ):
```

pre-commit and pytype fail too.

apivovarov: ok, upgrading to orbax-checkpoint-0.11.1 solved the issue

PR #827, opened 2024-11-11
Add bf16 test to subsampler.

ds-hwang:
There was a report about a bf16-related bug in the downstream, so I created a unit test to demonstrate that axlearn doesn't have the bug.


ds-hwang:
Could you review simple unittest? From 887

markblee (approved):

PR #832, opened 2024-11-13
Implement sequence_mask().

ds-hwang:
axlearn version of tf.sequence_mask.
https://www.tensorflow.org/api_docs/python/tf/sequence_mask


ds-hwang:
PTAL? from 893

ruomingp (approved):

PR #834, opened 2024-11-13
Updates orbax and adds support for max save/restore concurrent gb.

markblee:


ruomingp (approved):
Does the internal CI pass with this PR?

PR #836, opened 2024-11-13
Transformer extend_step supports multi steps generation (2/2).

ds-hwang:
In MultiheadAttention.extend_step, logit_bias was hardcoded to have a length of 1.
This PR modified it to support multi-step inputs. This change also makes extend_step more aligned with forward, reducing the overall code complexity.


ds-hwang:
Could you take a look? From 901

markblee (approved):

ds-hwang:
Thank you for review!

PR #847, opened 2024-11-18
Conv1D supports paddings.

ds-hwang:
Conv1D is designed for processing sequence data, so it is natural to handle paddings as part of its functionality.
Thus, it introduces Conv1DWithPadding.


ds-hwang:
@ruomingp could you review? from 913

ruomingp (approved):

PR #851, opened 2024-11-19
improve GCS perf: Change resource limit to request

samos123:
By using a limit we're restricting the resources that GCS needs to achieve decent performance. This PR changes the resource limit to a request, which allows GCS Fuse to use all available resources. This is now also the recommended default for GKE Standard clusters.


ruomingp (commented):
What's the status of this PR? @samos123 and @markblee ?

markblee (approved):
@samos123 can you rebase/push again? It seems to be going through the old circle-ci path.

samos123:
I rebased.

PR #854, opened 2024-11-21
Rollback to Jax 0.4.33

kelvin-zou:
Roll back to fix a few bugs in TPU. Will roll forward with a higher version Jax later.


ruomingp (approved):
Thanks!

PR #865, opened 2024-12-02
Refactoring Convolution; factor out convolutions from layers.py

ds-hwang:
Refactoring Convolution; factor out convolutions from layers.py


ds-hwang:
@ruomingp could you review? from 936

ruomingp (approved):

PR #866, opened 2024-12-02
Support orbax state builder

hanzhi713:
This PR adds `OrbaxStateBuilder` and refactors `TensorStoreStateBuilder`, and prepares for Orbax emergency checkpointing, which cannot be meaningfully represented as a path. Therefore, we add support of using `base_dir` and `step` as arguments to Builder, which abstracts away the underlying checkpoint structure. This paves the way of letting users restore state from the results of emergency checkpoint.

This PR also upgrades orbax slightly so it supports concurrent save and restore gb.

### DIFF
--- a/axlearn/common/state_builder.py
+++ b/axlearn/common/state_builder.py
@@ -26,9 +26,11 @@ from axlearn.common.checkpointer import (
     CheckpointValidationType,
     StateStorage,
     TensorStoreStateStorage,
+    build_step_dir,
     check_state_structure,
     parse_step_from_dir,
 )
+from axlearn.common.checkpointer_orbax import OrbaxCheckpointer
 from axlearn.common.config import (
     REQUIRED,
     ConfigOr,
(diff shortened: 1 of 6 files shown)


axlearn/common/state_builder.py
```diff
@@ -370,26 +372,60 @@ def clone_tree(in_tree: NestedTensor) -> NestedTensor:
...
 
     @config_class
     class Config(Builder.Config):
-        dir: Required[str] = REQUIRED
+        """Configures BaseStateStorageBuilder.
+
+        Attributes:
+            base_dir: Base directory that contains checkpoints of a trainer, usually containing
+                subdirs, one for each checkpointed step.
+            step: Step number to load.
```
ruomingp, line 386: Please see my question about step=None in the internal PR.

ruomingp (changes requested):

ruomingp (approved):

markblee (approved):

PR #867, opened 2024-12-04
Optimizer offloading through weight-only offload

hanzhi713:
This PR requires jax >= 0.4.34, != 0.4.35, >=0.4.36: it works on jax 0.4.34, but is broken on jax 0.4.35 due to libtpu bug. It worked on nightly jax 0.4.36 as of 10/30.

This PR represents effort to enable optimizer offloading. The approach we use in this PR is weight-only offloading, which is based on similar building blocks as activation offloading (aka remat offload). When offloading is enabled, optimizer states are stored on CPU pinned memory. Before apply optimizer to calculate updates, optimizer states are moved from CPU memory to HBM via `jax.device_put`. The new optimizer states are moved back from HBM to CPU.

An alternative approach to this PR is host computation. Host computation means that optimizer transformations are computed on CPU. Before the start of the computation, gradients and weights are transferred to CPU, and after the computation, their new values are transferred back to HBM. This method has lower HBM footprint, but it's much 2x ~ 3x slower due to slow CPU computation. Also, it's very buggy.

TLDR: to be merged after upgrading jax to 0.4.36.

### DIFF
--- a/axlearn/common/optimizer_base.py
+++ b/axlearn/common/optimizer_base.py
@@ -16,14 +16,13 @@ factorization, specifically:
    - weight_decay_scale: control the weight decay rate.
 """
 import dataclasses
-from collections.abc import Sequence
 from typing import Any, Callable, NamedTuple, Optional, Union
 
 import optax
 import typing_extensions
 
-from axlearn.common.base_layer import FactorizationSpec, NestedParameterSpec
-from axlearn.common.utils import Tensor, TensorSpec
+from axlearn.common.base_layer import FactorizationSpec, ParameterSpec
+from axlearn.common.utils import Nested, Tensor, TensorSpec
 
 
 @dataclasses.dataclass
@@ -66,8 +65,7 @@ class TransformUpdateFn(typing_extensions.Protocol):
 
 # Specification of an optimizer state array.
 OptStateSpec = TensorSpec
-NestedOptStateSpec = Union[OptStateSpec, dict, Sequence]
-TransformPartitionSpecFn = Callable[[NestedParameterSpec], NestedOptStateSpec]
+TransformPartitionSpecFn = Callable[[Nested[ParameterSpec]], Nested[OptStateSpec]]
 
 
 class PartitionedGradientTransformation(NamedTuple):
--- a/axlearn/common/optimizers.py
+++ b/axlearn/common/optimizers.py
@@ -36,10 +36,11 @@ import optax
 import typing_extensions
 from absl import logging
 from jax import numpy as jnp
+from jax._src.sharding_impls import TransferToMemoryKind
 from optax._src import numerics
 
 from axlearn.common import schedule, struct
-from axlearn.common.base_layer import NestedParameterSpec, ParameterSpec, PartitionSpec
+from axlearn.common.base_layer import ParameterSpec, PartitionSpec
 from axlearn.common.config import ConfigOr, maybe_instantiate
 from axlearn.common.factorized_rms import scale_by_factored_rms
 from axlearn.common.module import current_context
@@ -51,8 +52,8 @@ from axlearn.common.optimizer_base import (
     TransformPartitionSpecFn,
 )
 from axlearn.common.utils import (
+    MemoryKind,
     Nested,
-    NestedPartitionSpec,
     NestedTensor,
     NestedTree,
     Tensor,
@@ -139,19 +140,40 @@ def with_partition_fn(
     return PartitionedGradientTransformation(init=init_fn, update=update_fn, partition=partition_fn)
 
 
-def copy_partition(param_specs: NestedParameterSpec) -> NestedPartitionSpec:
+def copy_partition(
+    specs: Nested[OptStateSpec],
+    *,
+    pattern: Union[None, str, re.Pattern] = None,
+    memory_kind: Optional[MemoryKind] = None,
+) -> Nested[OptStateSpec]:
+    """Copies OptStateSpec and optionally assigns with a different memory kind.
+
+    Args:
+        specs: Nested[OptStateSpec] to copy from.
+        pattern: Regex to match the full path of each spec. Matched specs will have their memory
+            kind replaced with `memory_kind`.
+        memory_kind: New memory kind. Default to None.
+    Returns:
+        A Nested[OptStateSpec] with possibly a different memory kind.
+    """
     return jax.tree.map(
-        lambda param_spec: OptStateSpec(
-            dtype=param_spec.dtype, shape=param_spec.shape, mesh_axes=param_spec.mesh_axes
+        lambda path, spec: OptStateSpec(
+            dtype=spec.dtype,
+            shape=spec.shape,
+            mesh_axes=spec.mesh_axes,
+            memory_kind=memory_kind
+            if pattern and re.fullmatch(pattern, path)
+            else spec.memory_kind,
         ),
-        param_specs,
+        tree_paths(specs),
+        specs,
     )
 
 
 def trace_partition(
     base: optax.GradientTransformation,
 ) -> PartitionedGradientTransformation:
-    def partition_fn(param_specs: NestedParameterSpec) -> NestedPartitionSpec:
+    def partition_fn(param_specs: Nested[ParameterSpec]) -> Nested[OptStateSpec]:
         return optax.TraceState(trace=copy_partition(param_specs))
 
     return with_partition_fn(base, partition_fn)
@@ -160,7 +182,7 @@ def trace_partition(
 def adam_partition(base: optax.GradientTransformation) -> PartitionedGradientTransformation:
     state: optax.ScaleByAdamState = base.init({})
 
-    def partition_fn(param_specs: NestedParameterSpec) -> NestedPartitionSpec:
+    def partition_fn(param_specs: Nested[ParameterSpec]) -> Nested[OptStateSpec]:
         return optax.ScaleByAdamState(
             count=OptStateSpec(
                 dtype=state.count.dtype, shape=state.count.shape, mesh_axes=PartitionSpec()
@@ -950,7 +972,7 @@ def ema(
         )
         return updates, new_state
 
-    def partition_fn(param_specs: NestedParameterSpec) -> NestedPartitionSpec:
+    def partition_fn(param_specs: Nested[ParameterSpec]) -> Nested[OptStateSpec]:
         def get_ema_partition(param_spec: ParameterSpec) -> OptStateSpec:
             # Store momentum in accumulator_dtype if it is set and p is not scalar.
             if param_spec.shape and accumulator_dtype is not None:
@@ -1412,7 +1434,7 @@ def skip_and_clip_by_global_norm(
             drop_stats=new_drop_stats,
         )
 
-    def partition_fn(param_specs: NestedParameterSpec) -> NestedPartitionSpec:
+    def partition_fn(param_specs: Nested[ParameterSpec]) -> Nested[OptStateSpec]:
         if use_adaptive_drop_norm:
             one = jnp.ones([], jnp.float32)
             dict_thresholds = drop_norm(count=one, mean=one, stddev=one)
@@ -1571,7 +1593,7 @@ def param_ema(
         )
         return updates, ParamEmaState(count=count_inc, ema=new_ema)
 
-    def partition_fn(param_specs: NestedParameterSpec) -> NestedPartitionSpec:
+    def partition_fn(param_specs: Nested[ParameterSpec]) -> Nested[OptStateSpec]:
         return ParamEmaState(
             count=OptStateSpec(dtype=jnp.int32, shape=[], mesh_axes=PartitionSpec()),
             ema=copy_partition(param_specs),
@@ -1617,7 +1639,7 @@ def scale_by_lion(
         updates = jax.tree.map(lambda g, m: jnp.sign((1.0 - b1) * g + b1 * m), updates, state.mu)
         return updates, ScaleByLionState(count=count_inc, mu=mu)
 
-    def partition_fn(param_specs: NestedParameterSpec) -> NestedPartitionSpec:
+    def partition_fn(param_specs: Nested[ParameterSpec]) -> Nested[OptStateSpec]:
         mu_specs = param_specs
         if mu_dtype is not None:
             mu_specs = jax.tree.map(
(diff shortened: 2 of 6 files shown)


axlearn/common/optimizers.py
```diff
@@ -1993,3 +2008,93 @@ def _update2(u: Tensor, param: OptParam):
...
+
+    Example usage:
+    ```python
+    your_opt = adamw_optimizer(...)
+    offloaded_opt = offload_optimizer(your_opt)
+    ```
+
+    Only wrap the optimizer that you actually want to offload with this function to avoid
+    unneseccary overhead. This is usually the optimizer that occupies the most HBM. For example,
+    when you have chained optimizers:
```
ruomingp, line 2050: Where does the overhead come from? Is it from the states of `clip_by_global_norm` being offloaded? If so, could we use regular expressions to specify which states to offload?

axlearn/common/optimizers.py
```diff
@@ -1993,3 +2008,93 @@ def _update2(u: Tensor, param: OptParam):
...
+        return optimizer.init(params)
+
+    def update_fn(updates: optax.Updates, state: optax.OptState, params: NestedOptParam):
+        # TransferToMemoryKind let us change the memory kind of tensors without specifying the full
+        # sharding (i.e. jax.sharding.NamedSharding). Although there's no documentation about it,
+        # it's specified in the API signature. Reference:
+        # https://github.com/jax-ml/jax/blob/21f8885a9e104b8828c9a8b721eed0c68b622691/jax/_src/api.py#L2220
+        state = jax.device_put(state, TransferToMemoryKind(offload_src))
+        updates, state = optimizer.update(updates, state, params)
+        state = jax.device_put(state, TransferToMemoryKind(offload_dst), donate=True)
```
ruomingp, line 2094: Do we need explicit `device_put` calls here? Is it enough to specify the partition spec with the right memory_kind?

axlearn/common/optimizers.py
ruomingp, line 2050: Where does the overhead come from? Is it from the states of `clip_by_global_norm` being offloaded? If so, could we use regular expressions to specify which states to offload?

ruomingp (commented):
Nice!

axlearn/common/optimizer_base.py
```diff
@@ -66,8 +65,7 @@ def __call__(
 
 # Specification of an optimizer state array.
 OptStateSpec = TensorSpec
-NestedOptStateSpec = Union[OptStateSpec, dict, Sequence]
-TransformPartitionSpecFn = Callable[[NestedParameterSpec], NestedOptStateSpec]
+TransformPartitionSpecFn = Callable[[Nested[ParameterSpec]], Nested[OptStateSpec]]
```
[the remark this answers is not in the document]
hanzhi713: Not really, but the NestedXXX is deprecated in favor of Nested[XXX], I made this change.

axlearn/common/optimizers.py
```diff
@@ -1993,3 +2008,93 @@ def _update2(u: Tensor, param: OptParam):
         partition=lambda _: OptStateSpec(shape=[], dtype=jnp.int32, mesh_axes=PartitionSpec()),
     )
     return named_chain(**tx)
+
+
+def offload_optimizer(
+    optimizer: ConfigOr[PartitionedGradientTransformation],
+    *,
+    offload_src: Optional[MemoryKind] = "device",
+    offload_dst: Optional[MemoryKind] = "pinned_host",
```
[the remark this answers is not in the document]
hanzhi713: Good point, we don't need Optional.

axlearn/common/optimizer_base.py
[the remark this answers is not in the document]
ruomingp: OK. In the future could you split clean-up changes to separate PRs?

axlearn/common/optimizers.py
```diff
@@ -1993,3 +2008,93 @@ def _update2(u: Tensor, param: OptParam):
...
+    state and the updated new optimizer state using `jnp.where`, which doesn't support tensors on
+    `pinned_host` memory space.
+    """
+    optimizer = maybe_instantiate(optimizer)
+    if offload_src is None or offload_dst is None:
+        raise ValueError(
+            "offload_src and offload_dst cannot be None when using optimizer offloading."
+        )
+
+    logging.info("Optimizer offloading enabled.")
```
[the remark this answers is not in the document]
hanzhi713: Good point.

axlearn/common/optimizers.py
hanzhi713: I can add regex which could be useful for the future. This will default to ".*". Does this sound good to you?

ruomingp: Is the overhead observed or hypothesized? 

`clip_by_global_norm` doesn't have any state. 

OTOH, if state offload can be an issue, maybe certain states in the optimizer part (e.g., steps for LR schedule) shouldn't be offloaded either. Hence the suggestion to use a regex to control which state to offload.

hanzhi713: Hypothesized. The command wasn't saying that specifically offloading clip by global norm is slow, but in general it's not a good idea to offload an entire chain unless you know what you are doing.

hanzhi713: I agree regex offers finer control. I can add that.

axlearn/common/optimizers.py
hanzhi713: Haven't tested with_sharding_constraint, if you are refering to that. However, specifying full sharding requires us to store the sharding when the partition fn is called, which is not preferred by John (see internal comments)

axlearn/common/optimizer_base.py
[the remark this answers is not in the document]
hanzhi713: Sure.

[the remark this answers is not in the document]
hanzhi713: It was kind of related since there's just onr reference to NestedOptStateSpec after I fix the typing in optimizers.py, so I removed it altogether.

axlearn/common/optimizers.py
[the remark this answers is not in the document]
hanzhi713: Done.

axlearn/common/optimizers.py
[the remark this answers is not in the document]
hanzhi713: Done.

axlearn/common/optimizers.py
hanzhi713: Done.

axlearn/common/optimizers.py
ruomingp: I'm not sure whether we need `with_sharding_constraint`, since jit/pjit should already apply shardings returned by `partition_fn` to the states. What happens without these `device_put` calls?

If necessary, we can always invoke `partition_fn` here to compute the sharding on the fly (instead of storing them) and apply `with_sharding_constraint`.

ruomingp (commented):
A question about device_put...

hanzhi713:
> A question about device_put...

Before the optimizer can be invoked, the offloaded optimizer states need to be transferred to device memory space. If we remove these device_put calls, we will get errors like `xxx is not supported on pined_host memory space`, where xxx is some XLA primitive operations such as `add` (forgot the exact error message but is something like this)

axlearn/common/optimizers.py
```diff
@@ -139,19 +140,40 @@ def update_fn(
     return PartitionedGradientTransformation(init=init_fn, update=update_fn, partition=partition_fn)
 
 
-def copy_partition(param_specs: NestedParameterSpec) -> NestedPartitionSpec:
+def copy_partition(
+    param_specs: Nested[ParameterSpec],
+    *,
+    pattern: Union[None, str, re.Pattern] = None,
+    memory_kind: Optional[MemoryKind] = None,
+) -> Nested[OptStateSpec]:
```
ruomingp, line 148: Instead of coupling creation of `OptStateSpec` and setting of `memory_kind`, how about having a separate function for setting memory kind?

```
def set_memory_kind(opt_state_spec: Nested[OptStateSpec], *, pattern, memory_kind):
```

This allows `set_memory_kind` to be called multiple times, maybe for different memory kind. WDYT?

axlearn/common/optimizers.py
```diff
@@ -1993,3 +2015,93 @@ def _update2(u: Tensor, param: OptParam):
...
+            if re.fullmatch(pattern, path)
+            else tensor,
+            tree_paths(state),
+            state,
+        )
+
+    def update_fn(updates: optax.Updates, state: optax.OptState, params: NestedOptParam):
+        state = _move_fn(state, offload_src)
+        updates, state = optimizer.update(updates, state, params)
+        state = _move_fn(state, offload_dst)
```
ruomingp, line 2099: I wonder whether the explicit `device_put` calls mean that we need to move all optimizer states into device before optimizer computation, leading to high device memory usage.

In theory, optimizer computation can be streaming---the optimizer states of variables can be updated separately so we can stream them into and out of device memory as needed. How do we achieve this?

An alternative is to use `with_sharding_constraint` calls instead of `device_put` calls, where the partition specs given to the `with_sharding_constraint` calls will put the arrays in the device memory. WDYT?

ruomingp (commented):
Thanks for the clarification on `device_put` calls. Could you add a comment on why it's necessary? Also two suggestions...

axlearn/common/optimizers.py
hanzhi713: I do not see how `set_memory_kind` will be different from `copy_partition`. Signature and implementation will be the same.

axlearn/common/optimizers.py
hanzhi713: I've verified in profile before that `device_put` won't move everything at once. When we pass a pytree of arrays to `device_put`, each array in the pytree is moved independent of one another. The exact order is decided by the latency hiding scheduler. The scheduler will try to overlap the transfers of each state with the state update on TPU whenever possible. There is some memory spike due the the temporary state in HBM, but the spike is much less than the full memory usage of all states. Moreover, when the optimizer is run, all activations are released, so we have less memory pressure at that point in time.

If I remove clip by global norm and use `StackedTransformerLayer`, the optimizer transfers can be overlapped with the backward pass, leading to faster step time and less memory spike. The reason is that clip by global norm requires all gradients to be computed before optimizer can be executed, and `RepeatedTransformerLayer` stack weights of all layers into a single buffer which does not allow the layer-by-layer optimizer execution.

axlearn/common/optimizers.py
```diff
@@ -1993,3 +2015,93 @@ def _update2(u: Tensor, param: OptParam):
...
+            context.
+
+    This function returns a new `PartitionedGradientTransformation` that
+    1. Puts matched states of the wrapped optimizer on `offload_dst` through the partition function
+       during state initialization in the trainer.
+    2. Copies the matched states to `offload_src` before `optimizer.update` is called.
+    3. Copies the matched updated states to `offload_dst` after `optimizer.update` is called.
+
+    The regex pattern is matched against the full path of each optimizer state. An example full
+    path is optimizer/1/0/mu/decoder/transformer/repeat/layer/feed_forward/linear1_0. If the
```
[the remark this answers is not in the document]
hanzhi713: Well that's just the current implementation of `adam*_optimizer`

axlearn/common/optimizers.py
```diff
@@ -1993,3 +2015,93 @@ def _update2(u: Tensor, param: OptParam):
...
+    pattern: Union[str, re.Pattern] = ".*",
+    offload_src: MemoryKind = "device",
+    offload_dst: MemoryKind = "pinned_host",
+) -> PartitionedGradientTransformation:
+    """Offload the state of the wrapped optimizer that matches `pattern` to `offload_dst`.
+
+    Args:
+        optimizer: The optimizer to offload.
+        pattern: Regex pattern used to match the path of optimizer states. Matched states will be
+            offloaded. Default to regex that matches all states.
```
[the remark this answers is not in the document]
hanzhi713: Done

axlearn/common/optimizers.py
```diff
@@ -1993,3 +2015,93 @@ def _update2(u: Tensor, param: OptParam):
...
+
+    This function returns a new `PartitionedGradientTransformation` that
+    1. Puts matched states of the wrapped optimizer on `offload_dst` through the partition function
+       during state initialization in the trainer.
+    2. Copies the matched states to `offload_src` before `optimizer.update` is called.
+    3. Copies the matched updated states to `offload_dst` after `optimizer.update` is called.
+
+    The regex pattern is matched against the full path of each optimizer state. An example full
+    path is optimizer/1/0/mu/decoder/transformer/repeat/layer/feed_forward/linear1_0. If the
+    pattern should not depend on model structure, you can use ".*mu.*" to offload all `mu`.
```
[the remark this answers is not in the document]
hanzhi713: Done

axlearn/common/optimizers.py
ruomingp: Imagine in the future we have many types of memory kinds, e.g., "remote_host". Then we can do:

```
opt_state_specs = copy_partition(...)
opt_state_specs = set_memory_kind(..., "pinned_host")
opt_state_specs = set_memory_kind(..., "remote_host")
```

axlearn/common/optimizers.py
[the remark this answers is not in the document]
ruomingp: OK. At least adastar uses named_chain: https://github.com/apple/axlearn/blob/9eabbee86d0596962ef02f8c13da1d5caa30153c/axlearn/common/optimizers.py#L2017

ruomingp (approved):

axlearn/common/optimizers.py
hanzhi713: Sure, will add a comment. Regarding using `with_sharding_constraint`, I would prefer not to use that as I don't want to populate the full sharding spec when alternative methods (i.e. `device_put`) that can modify the memory kind directly exist.

axlearn/common/optimizers.py
hanzhi713: Wouldn't it be the same as
```
opt_state_specs = copy_partition(...)
opt_state_specs = copy_partition(..., "pinned_host")
opt_state_specs = copy_partition(..., "remote_host")
```
Do you suggest using a separate function for slightly better readability?

ruomingp: The difference is that `copy_partition` also performs the type conversion from `Nested[ParameterSpec]` to `Nested[OptStateSpec]`.

hanzhi713: I see. I can change the type of param_specs in `copy_partition` to `Nested[OptStateSpec]` since `ParameterSpec ` is a subclass of `OptStateSpec` and `copy_partition` doesn't use any new fields from `ParameterSpec`. Does this sound good?

ruomingp: Thanks. SG.

hanzhi713: Done.

axlearn/common/optimizers.py
hanzhi713: Done.

axlearn/common/optimizers.py
```diff
@@ -139,19 +140,40 @@ def update_fn(
...
+    memory_kind: Optional[MemoryKind] = None,
+) -> Nested[OptStateSpec]:
+    """Copies OptStateSpec and optionally assigns with a different memory kind.
+
+    Args:
+        specs: Nested[OptStateSpec] to copy from.
+        pattern: Regex to match the full path of each spec. Matched specs will have their memory
+            kind replaced with `memory_kind`.
+        memory_kind: New memory kind. Default to None.
+    Returns:
```
ruomingp, line 156:
```suggestion

    Returns:
```

ruomingp (approved):

axlearn/common/optimizers.py
hanzhi713: Done.

ruomingp (approved):

markblee (approved):

hanzhi713:
@markblee Can you take a look at the pytype errors in CI? Should I change Nested type to include tuple/namedtuple or should I just ignore the errors?

markblee:
> @markblee Can you take a look at the pytype errors in CI? Should I change Nested type to include tuple/namedtuple or should I just ignore the errors?

Maybe we can relax the partition fn return type to include named tuple? Changing Nested may have undesirable impact elsewhere.

hanzhi713:
> Maybe we can relax the partition fn return type to include named tuple? Changing Nested may have undesirable impact elsewhere.

Thanks. I included the return type in Nested[...]. Just NamedTuple wouldn't work.

hanzhi713:
@ruomingp Could you please re-approve this PR? Thanks.

This is an opt-in feature, so it will be fine even if the jax upgrade is reverted.

ruomingp (approved):

PR #876, opened 2024-12-10
DiT: implement init_states and extend_step for DiT transformer

ds-hwang:
Beyond vision, speech also uses it for generative model, and speech needs streaming decoding.


ds-hwang:
@markblee, @ruomingp could you review it? from 946

ruomingp (approved):

PR #881, opened 2024-12-11
Support fine grained activation sharding.

ptoulme-aws:
Hello, this is a draft PR to support fine grained activation sharding. All specs are defaulted to None, so this PR maintains all existing behavior of the Axlearn codebase.

I am requesting feedback and comments on the approach. 

With this PR the user can control more closely how GSPMD partitions activations allowing more control of the collectives generated by the SPMD partitioner. For example, GSPMD will generate all-to-alls around the gather operation of the vocab table, but with this PR user can annotate partition specs to generate a reduce-scatter instead.

Thank you!

### DIFF
--- a/axlearn/common/layers.py
+++ b/axlearn/common/layers.py
@@ -54,6 +54,7 @@ from axlearn.common.quantized_dot_general.layers import DenseGeneralBaseLayer
 from axlearn.common.utils import (
     NestedTensor,
     Tensor,
+    maybe_shard,
     partial_with_fn_metadata,
     with_sharding_constraint,
 )
@@ -331,6 +332,10 @@ class RMSNorm(BaseNormalizationLayer):
         eps: float = 1e-8
         # Cast input to this dtype for the 'forward' call. If None, do not cast.
         forward_dtype: Optional[jnp.dtype] = jnp.float32
+        # If not None, how to partition input activation values.
+        input_partition_spec: Optional[tuple[Optional[str]]] = None
+        # If not None, how to partition output activation values.
+        output_partition_spec: Optional[tuple[Optional[str]]] = None
 
     def _create_layer_parameter_specs(self) -> dict[str, ParameterSpec]:
         cfg = self.config
@@ -341,6 +346,7 @@ class RMSNorm(BaseNormalizationLayer):
     def forward(self, x: Tensor, *, paddings: Optional[Tensor] = None) -> Tensor:
         del paddings  # paddings do not affect LayerNorm results
         cfg = self.config
+        x = maybe_shard(x, cfg.input_partition_spec)
         x_dtype = x.dtype
         if cfg.forward_dtype is not None:
             x = x.astype(cfg.forward_dtype)
@@ -348,6 +354,7 @@ class RMSNorm(BaseNormalizationLayer):
         x = x * jax.lax.rsqrt(moment2 + cfg.eps)
         x = x.astype(x_dtype)
         x = x * self.parameters["scale"]
+        x = maybe_shard(x, cfg.output_partition_spec)
         return x
 
 
@@ -780,6 +787,12 @@ class Embedding(BaseLayer):
 
         num_embeddings: Required[int] = REQUIRED  # Maximum number of embeddings in table.
         dim: Required[int] = REQUIRED  # Embedding vector dimensionality.
+        # If not None, how to partition input activation values.
+        input_partition_spec: Optional[tuple[Optional[str]]] = None
+        # If not None, how to partition embedding table.
+        embedding_partition_spec: Optional[tuple[Optional[str]]] = None
+        # If not None, how to partition output activation values.
+        output_partition_spec: Optional[tuple[Optional[str]]] = None
 
     @classmethod
     def default_config(cls):
@@ -814,8 +827,13 @@ class Embedding(BaseLayer):
         )
 
     def forward(self, x: Tensor) -> Tensor:
+        cfg = self.config
+        x = maybe_shard(x, cfg.input_partition_spec)
         emb = self.parameters["weight"]
-        return emb[x]
+        emb = maybe_shard(emb, cfg.embedding_partition_spec)
+        activation = emb[x]
+        activation = maybe_shard(activation, cfg.output_partition_spec)
+        return activation
 
     def attend(self, x: Tensor) -> Tensor:
         """Apply query array 'x' to the embedding weight array.
--- a/axlearn/common/utils.py
+++ b/axlearn/common/utils.py
@@ -444,6 +444,12 @@ def with_sharding_constraint(x, shardings):
     return jax.lax.with_sharding_constraint(x, shardings)
 
 
+def maybe_shard(x: NestedTensor, partition_spec: Optional[PartitionSpec]) -> NestedTensor:
+    if partition_spec is None:
+        return x
+    return with_sharding_constraint(x, PartitionSpec(*partition_spec))
+
+
 def replicate_to_local_data(x: NestedTensor) -> NestedTensor:
     """Replicates and converts Tensors in `x` to local DeviceArrays.
 
--- a/axlearn/common/layers_test.py
+++ b/axlearn/common/layers_test.py
@@ -18,6 +18,7 @@ import math
 from collections.abc import Sequence
 from functools import partial
 from typing import Optional, Union
+from unittest import mock
 
 import jax.random
 import numpy as np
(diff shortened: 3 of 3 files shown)


axlearn/common/attention.py
```diff
@@ -2771,11 +2788,14 @@ def _linear2(x):
 
         remat_pt2 = "linear2"
         if cfg.structure == "prenorm":
+            inputs = maybe_shard(inputs, cfg.prenorm_partition_spec)
             x = self.norm(inputs)
+            x = maybe_shard(x, cfg.premlp_partition_spec)
```
ruomingp, line 2793: Do we need this? I suppose `norm` usually does not change the partition spec?

axlearn/common/attention.py
```diff
@@ -2355,6 +2356,12 @@ class Config(BaseLayer.Config):
         # Ref: https://github.com/google/praxis/blob/main/praxis/layers/transformers.py#L1129
         # TODO (bwzhang@) Adding a unittest for the hybridnorm.
         structure: str = "prenorm"
+        # If not None, how to partition pre norm activation values.
+        prenorm_partition_spec: Optional[tuple[Optional[str]]] = None
+        # If not None, how to partition pre attention activation values.
+        preattention_partition_spec: Optional[tuple[Optional[str]]] = None
+        # If not None, how to partition post attention activation values.
+        postattention_partition_spec: Optional[tuple[Optional[str]]] = None
```
ruomingp, line 2364: How is this different from setting https://github.com/apple/axlearn/blob/2134a258a869f679e4dd2625565a2ca693126495/axlearn/common/attention.py#L3319?

axlearn/common/attention.py
```diff
@@ -2355,6 +2356,12 @@ class Config(BaseLayer.Config):
         # Ref: https://github.com/google/praxis/blob/main/praxis/layers/transformers.py#L1129
         # TODO (bwzhang@) Adding a unittest for the hybridnorm.
         structure: str = "prenorm"
+        # If not None, how to partition pre norm activation values.
+        prenorm_partition_spec: Optional[tuple[Optional[str]]] = None
```
ruomingp, line 2360: Please see my comment about `input_partition_spec` below.

axlearn/common/attention.py
```diff
@@ -2355,6 +2356,12 @@ class Config(BaseLayer.Config):
...
+        # If not None, how to partition pre attention activation values.
+        preattention_partition_spec: Optional[tuple[Optional[str]]] = None
```
ruomingp, line 2362: ditto.

axlearn/common/attention.py
```diff
@@ -3315,6 +3335,21 @@ def set_ffn_partition_specs(ff_layer: TransformerFeedForwardLayer.Config):
...
+            postattention_partition_spec = (fsdp_axis_names, tp_axis_names, None)
+
+            layer_cfg.self_attention.set(
+                prenorm_partition_spec=prenorm_partition_spec, 
+                preattention_partition_spec=preattention_partition_spec, 
+                postattention_partition_spec=postattention_partition_spec)
+            layer_cfg.feed_forward.set(
+                prenorm_partition_spec=prenorm_partition_spec, 
+                premlp_partition_spec=preattention_partition_spec, 
+                postmlp_partition_spec=postattention_partition_spec)
```
ruomingp, line 3352: Is `postattention_partition_spec` always the same as `prenorm_partition_spec` for `feed_forward` since the outputs of atten are used as inputs for ffn?

ruomingp (changes requested):
Thanks! Some questions about the API...

axlearn/common/attention.py
kelvin-zou: +1, I also have a high level question, should the sharding config auto-derived and optimized by XLA for most of the intermediary activations, based on weights and outputs etc? 
I'd like to see a concrete example to understand why explicit sharding is needed here.

axlearn/common/layers.py
```diff
@@ -814,8 +821,14 @@ def _create_layer_parameter_specs(self) -> dict[str, ParameterSpec]:
         )
 
     def forward(self, x: Tensor) -> Tensor:
+        cfg = self.config
```
kelvin-zou, line 824: Same question here, can we handle this inside XLA compiler directly instead of making it explicit here? Adding this kind of fine-level sharding spec is a divergence from GSPMD programming paradigm imho.

axlearn/common/attention_test.py
```diff
@@ -5091,6 +5091,70 @@ def test_set_double_shard_weights_config_for_list_of_configs(
...
+    ):
+        cfg_layer: TransformerLayer.Config = TransformerLayer.default_config().set(
+            cross_attention=cross_attention_cfg
+        )
+        cfg_layer.self_attention.structure = "prenorm"
+        cfg_layer.feed_forward.structure = "prenorm"
+        cfg_layer.self_attention.attention.input_linear = self_attention_input_linear_cfg
+        cfg_layers = [cfg_layer, cfg_layer]
+
+        cfg_layer.self_attention.prenorm_partition_spec = (fsdp_axis_names, tp_axis_names, None,)
```
kelvin-zou, line 5122: interesting, is it the real use case you are testing nowadays?

axlearn/common/attention.py
```diff
@@ -3315,6 +3335,21 @@ def set_ffn_partition_specs(ff_layer: TransformerFeedForwardLayer.Config):
             set_attn_partition_specs(layer_cfg.cross_attention.attention)
         if isinstance(layer_cfg.feed_forward, TransformerFeedForwardLayer.Config):
             set_ffn_partition_specs(layer_cfg.feed_forward)
+            
+        # Neuron backend needs fine grained activation sharding.
+        if jax.default_backend() == 'neuron':
```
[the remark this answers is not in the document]
ptoulme-aws: Got it. I will revise to make this a separate function instead.

axlearn/common/attention.py
```diff
@@ -2708,6 +2718,13 @@ class Config(BaseLayer.Config):
         # TODO(tlei3): deprecate this feature since we use TensorStats.
         add_value_rms_norm_summary: Sequence[str] = []
 
+        # If not None, how to partition pre norm activation values.
+        prenorm_partition_spec: Optional[tuple[Optional[str]]] = None
+        # If not None, how to partition pre MLP activation values.
+        premlp_partition_spec: Optional[tuple[Optional[str]]] = None
+        # If not None, how to partition post MLP activation values.
+        postmlp_partition_spec: Optional[tuple[Optional[str]]] = None
```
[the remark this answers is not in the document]
ptoulme-aws: Good catch! This one is not needed.

axlearn/common/attention.py
```diff
@@ -2771,11 +2788,14 @@ def _linear2(x):
 
         remat_pt2 = "linear2"
         if cfg.structure == "prenorm":
+            inputs = maybe_shard(inputs, cfg.prenorm_partition_spec)
```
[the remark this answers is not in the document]
ptoulme-aws: Will revise to take this approach.

axlearn/common/attention.py
ptoulme-aws: We want to change the spec post norm to not propagate the spec from norm to MLP linear1. 
So for sequence parallel we would make the premlp_partition_spec = ((fsdp, data), None, None) to force an all-gather on the sequence dimension.

axlearn/common/attention.py
ptoulme-aws: OpenXLA ShardingPropagation often makes non optimal decisions in propagation. SPMD partitioner then partitions based on what HloSharding was propagated to each HloInstruction, so if your sharding propagation is not optimal then you will see collectives generated such as all-to-all to resolve the sharding conflicts.

An example is if you set the partition spec for premlp or preattention to be ((fsdp, data), model, None) to force the computation to be done in sequence parallel, then ShardingPropagation pass will propagate that HloSharding spec down to the matmul causing an all-to-all collective. 

This is not as optimal as an all-gather of the sequence dimension before the matmul.

axlearn/common/attention.py
ptoulme-aws: Will revise to fix this.

axlearn/common/attention.py
ptoulme-aws: Will revise to fix this.

axlearn/common/attention.py
ptoulme-aws: It is; which one would you prefer to keep or rename both to input_partition_spec as you suggested?

axlearn/common/attention_test.py
ptoulme-aws: Yes, this is the real config for sequence parallel. I have confirmed with these specs and internal Neuron changes we can generate partitioned HLOs that perform Norms/Residuals/Dropouts partitioned along the sequence and matmuls partitioned along the model dimension without any unnecessary collectives, such as all-to-alls or collective-permutes or involuntary full rematerialization.

axlearn/common/layers.py
ptoulme-aws: Responded above with a concrete example. The main issues we are trying to solve here are 

1. We want Norms/Dropouts/Residuals to be computed in sequence parallel so partitioned along sequence dimension. This is to avoid redundant compute across tensor parallel workers.
2.. OpenXLA ShardingPropagation often makes non optimal decisions - [https://github.com/openxla/xla/blob/main/xla/service/sharding_propagation.h]([https://github.com/openxla/xla/blob/main/xla/service/sharding_propagation.h])
3. Those non optimal decisions such as say propagating sequence level sharding to the left hand side of the attention QKV proj or MLP up proj will cause the SPMD partitioner to insert all-to-alls or collective-permutes or involuntary full rematerialization.


I am curious, why Axlearn has an activation level sharding spec in use today for `linear2.output_partition_spec`
Was this added to also address non optimal SPMD partitioning or propagation?

axlearn/common/attention.py
ruomingp: Thanks for the explanation. Maybe we can add `output_partition_spec` to RMSNorm.Config to be consistent with Linear.Config. This is also more flexible as RMSNorm can be used in other places. WDYT?

ruomingp:
Thanks. Please re-request review when it's ready.

axlearn/common/utils.py
```diff
@@ -444,6 +444,12 @@ def with_sharding_constraint(x, shardings):
     return jax.lax.with_sharding_constraint(x, shardings)
 
 
+def maybe_shard(x, partition_spec) -> Tensor:
+    if partition_spec is None:
+        return x
+    return with_sharding_constraint(x, PartitionSpec(*partition_spec))
+
+
```
ptoulme-aws, line 452: pre-commit wanted these blank lines. I ran 

`pre-commit run -a`

axlearn/common/attention.py
ptoulme-aws: Yes, I actually originally had this! I have revised to take this approach.

ptoulme-aws:
@kelvin-zou  @ruomingp 

I have revised the PR based on your comments. I removed "WIP," as the PR is now production ready. I removed Neuron specific logic; I think that should be in a separate PR, and this PR should be enabling hardware agnostic activation sharding. 

I ran precommit and pytest and both are passing. Please take a look! Thank you.

ruomingp (commented):
Thanks! Are there tests that can be added for the new logic?

ptoulme-aws:
@ruomingp What test do you envision? What should I assert? This is effectively just putting 

with_sharding_constraint(tensor, PartitionSpec('fdsp', 'model')) 

behind a config.

ptoulme-aws:
@ruomingp  I could add a test where I make an RMSNorm and set the config for partition specs and then assert they are the same specs, but I am not sure what benefit that adds. I do not see tests like that for other configs, as that is like a setter/getter test.

axlearn/common/utils.py
```diff
@@ -444,6 +444,12 @@ def with_sharding_constraint(x, shardings):
     return jax.lax.with_sharding_constraint(x, shardings)
 
 
+def maybe_shard(x, partition_spec) -> Tensor:
```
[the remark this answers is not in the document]
ptoulme-aws: x is not always a tensor. For example, with `aten_output `from attention it is not a tensor. I can add the type for partition_spec

ruomingp:
> @ruomingp I could add a test where I make an RMSNorm and set the config for partition specs and then assert they are the same specs, but I am not sure what benefit that adds. I do not see tests like that for other configs, as that is like a setter/getter test.

How about patching `with_sharding_constraint` to make sure that it's called with the right tensors in forward() when we set partition specs?

axlearn/common/utils.py
```diff
@@ -443,6 +443,11 @@ def with_sharding_constraint(x, shardings):
         return x
     return jax.lax.with_sharding_constraint(x, shardings)
 
+def maybe_shard(x: Tensor, partition_spec: Optional[PartitionSpec]) -> Tensor:
```
ruomingp, line 446: Is `x` a `Nested[Tensor]`?

ptoulme-aws: I think sometimes it is. That is why I did not have a type. I know for `aten_output `it is not a tensor. 

What type do you recommend?

ptoulme-aws:
> > @ruomingp I could add a test where I make an RMSNorm and set the config for partition specs and then assert they are the same specs, but I am not sure what benefit that adds. I do not see tests like that for other configs, as that is like a setter/getter test.
> 
> How about patching `with_sharding_constraint` to make sure that it's called with the right tensors in forward() when we set partition specs?

Do you have an example of this?

ptoulme-aws:
@ruomingp @kelvin-zou 
I have added tests per @ruomingp specifications. All tests pass and pre-commit passes also. 

Thank you!

ruomingp (approved):

kelvin-zou (approved):

PR #882, opened 2024-12-11
Update RoformerQKVLinear to support kv_state

qdavid1:
When attaching QLinear under a Roformer, kv_state is needed as an input, and it is not supported before this PR.

Additionally, since the key is pre-rotated already, we allow the option of not rotating the key again.

### DIFF
--- a/axlearn/common/attention.py
+++ b/axlearn/common/attention.py
@@ -867,6 +867,7 @@ class QKVLinear(BaseQKVLinear):
         *,
         key: Optional[Tensor] = None,
         value: Optional[Tensor] = None,
+        kv_state: Optional[Tensor] = None,
         time_step: Optional[Tensor] = None,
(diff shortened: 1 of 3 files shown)


chongw (approved):

axlearn/common/attention.py
```diff
@@ -1252,6 +1278,7 @@ class Config(BaseQKVLinear.Config):
             RoFormerSinusoidalPositionalEmbedding.default_config()
         )
         input_linear: BaseQKVLinear.Config = QKVLinear.default_config()
+        rotary_key: bool = True
```
chongw, line 1281: nit: add a comment for the rotary_key ?

qdavid1: Done

ruomingp (approved):

PR #884, opened 2024-12-11
Input batch sharding strategy BATCH

apoorvtintin:
Axlearn currently supports DataPartitionType strategies FULL and REPLICATED for input batches. This PR adds support for a third sharding strategy DataPartitionType.BATCH. 

This sharding strategy shards the input batch on sharding axis inferred as "batch_axis". This ensures batches are replicated on the "model" dimension and thus avoids unnecessary collectives to reshard input batches when Tensor parallelism is used.

### DIFF
--- a/axlearn/common/evaler.py
+++ b/axlearn/common/evaler.py
@@ -31,6 +31,7 @@ from axlearn.common.metrics import MetricAccumulator, WeightedScalar
 from axlearn.common.module import Module, OutputCollection
 from axlearn.common.module import functional as F
 from axlearn.common.utils import (
+    DataPartitionType,
     NestedPartitionSpec,
     NestedTensor,
     Tensor,
@@ -81,6 +82,11 @@ class BaseMetricCalculator(Module):
         # evalers, not setting prefix will show the accuracies on the same plot for comparison
         # across evalers.
         prefix: Optional[str] = None
+        # Subset of mesh axis names over which the leaves of the input batch are sharded.
+        batch_axis_names: Union[str, Sequence[str]] = "data"
+        # The input partition:
+        # Options: FULL (default), BATCH, REPLICATED
+        input_partition_type: Optional[DataPartitionType] = DataPartitionType.FULL
 
     def __init__(
         self,
@@ -188,11 +194,11 @@ class BaseMetricCalculator(Module):
             in_shardings=(
                 self._model_param_partition_specs,  # model_params.
                 None,  # replicated_inputs (e.g., prng_key).
-                utils.input_partition_spec(),  # per_example_inputs.
+                utils.data_partition_type_to_spec(partition=self.config.input_partition_type, batch_axis_names=self.config.batch_axis_names),  # per_example_inputs.
             ),
             out_shardings=dict(
                 replicated=None,
-                per_example=utils.input_partition_spec(),
+                per_example=utils.data_partition_type_to_spec( partition=self.config.input_partition_type, batch_axis_names=self.config.batch_axis_names),
             ),
         )
 
@@ -574,6 +580,11 @@ class SpmdEvaler(Module):
         metric_calculator: BaseMetricCalculator.Config = ModelSummaryAccumulator.default_config()
         # If not None, writes input batches and `metric_calculator` forward outputs.
         output_writer: Optional[BaseOutputWriter.Config] = None
+        # Subset of mesh axis names over which the leaves of the input batch are sharded.
+        batch_axis_names: Union[str, Sequence[str]] = "data"
+        # The input partition:
+        # Options: FULL (default), BATCH, REPLICATED
+        input_partition_type: Optional[DataPartitionType] = DataPartitionType.FULL
 
     def __init__(
         self,
@@ -595,7 +606,7 @@ class SpmdEvaler(Module):
         self._add_child("input", maybe_set_config(cfg.input, is_training=False))
         self._add_child(
             "metric_calculator",
-            cfg.metric_calculator.set(eval_dtype=cfg.eval_dtype),
+            cfg.metric_calculator.set(eval_dtype=cfg.eval_dtype, batch_axis_names=cfg.batch_axis_names, input_partition_type=cfg.input_partition_type),
             model=model,
             model_param_partition_specs=model_param_partition_specs,
         )
@@ -691,7 +702,7 @@ class SpmdEvaler(Module):
 
             with jax.profiler.StepTraceAnnotation(cfg.name, step_num=step):
                 with jax.profiler.TraceAnnotation(f"{cfg.name}.forward"):
-                    global_input_batch = utils.host_to_global_device_array(input_batch)
+                    global_input_batch = utils.host_to_global_device_array(input_batch, partition=self.config.input_partition_type, batch_axis_names=self.config.batch_axis_names)
                     forward_outputs = self.metric_calculator.forward(
                         global_input_batch,
                         model_params=model_params,
--- a/axlearn/common/host_array_test.py
+++ b/axlearn/common/host_array_test.py
@@ -21,7 +21,6 @@ from axlearn.common.utils import (
     host_to_global_device_array,
 )
 
-
 def is_supported(
     platform: str,
     mesh_shape: tuple[int, int],
@@ -37,16 +36,15 @@ def is_supported(
         )
     )
 
-
 class HostArrayTest(TestCase):
     @parameterized.parameters(
         filter(
             lambda params: is_supported(*params),
             itertools.product(
                 ("cpu", "tpu"),  # platform,
-                ((1, 1), (4, 1), (2, 2), (8, 1), (4, 2)),  # mesh_shape
+                ((1, 1), (4, 1), (2, 2), (8, 1), (4, 2), (16, 4)),  # mesh_shape
                 (1, 16),  # global_batch_size
-                (DataPartitionType.FULL, DataPartitionType.REPLICATED),  # data_partition
+                (DataPartitionType.FULL, DataPartitionType.REPLICATED, DataPartitionType,BATCH),  # data_partition
             ),
         )
     )
--- a/axlearn/common/utils.py
+++ b/axlearn/common/utils.py
@@ -591,14 +591,17 @@ class DataPartitionType(Enum):
     FULL = "full"
     # Data are fully replicated across all devices.
     REPLICATED = "replicated"
+    # Data are partitioned across batch axis only.
+    BATCH = "batch"
 
-
-def data_partition_type_to_spec(partition: DataPartitionType) -> PartitionSpec:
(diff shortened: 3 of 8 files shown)


axlearn/common/utils.py
```diff
@@ -607,6 +610,7 @@ def host_to_global_device_array(
     host_arrays: Nested[Union[np.ndarray, Tensor]],
     *,
     partition: DataPartitionType = DataPartitionType.FULL,
```
ruomingp, line 612: I think @markblee plans to remove the `DataPartitionType` enum and rely on https://jax.readthedocs.io/en/latest/_autosummary/jax.make_array_from_process_local_data.html to support flexible partition specs.

ruomingp (changes requested):

axlearn/common/utils.py
apoorvtintin: Thanks that sounds promising. 

Hello @markblee, let me know if this PR is needed till you make your changes, or if you have your design in mind I can reshape the PR to make it compatible with your design.

hanzhi713: I think `make_array_from_process_local_data` is already merged: https://github.com/apple/axlearn/pull/781

hanzhi713:
@apoorvtintin With this new sharding strategy, how do you plan to support model parallelism across hosts (i.e. per host bs < 1)? If this is not a supported use case, then this change LGTM.  I could imagine the support for model parallelism across hosts is not needed in the short term for `neuron` since the number of chips per host is large.

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -423,6 +423,7 @@ def get_trainer_kwargs(
         raise NotImplementedError(f"Unknown model size {model_size}.")
     model_kwargs = trainer_kwargs.pop("model_kwargs")
     model_kwargs.setdefault("vocab_size", vocab_size)
+    trainer_kwargs["input_partition_type"] = None if backend != "neuron" else DataPartitionType.BATCH
```
kelvin-zou, line 426: Please go with a configmodifier instead of hardcode in the code, otherwise it becomes hard for people to debug

axlearn/common/evaler.py
```diff
@@ -188,11 +194,11 @@ def _pjit(self, fn: Callable) -> Callable:
             in_shardings=(
                 self._model_param_partition_specs,  # model_params.
                 None,  # replicated_inputs (e.g., prng_key).
-                utils.input_partition_spec(),  # per_example_inputs.
+                utils.data_partition_type_to_spec(partition=self.config.input_partition_type, batch_axis_names=self.config.batch_axis_names),  # per_example_inputs.
             ),
             out_shardings=dict(
                 replicated=None,
-                per_example=utils.input_partition_spec(),
+                per_example=utils.data_partition_type_to_spec( partition=self.config.input_partition_type, batch_axis_names=self.config.batch_axis_names),
```
kelvin-zou, line 201: Did you run through pylint? this line seems quite long.

axlearn/common/utils.py
```diff
@@ -591,14 +591,17 @@ class DataPartitionType(Enum):
...
-
-def data_partition_type_to_spec(partition: DataPartitionType) -> PartitionSpec:
+def data_partition_type_to_spec(partition: DataPartitionType, * , batch_axis_names: Union[str, Sequence[str]] = ("data", "fsdp")) -> PartitionSpec:
     """Returns a PartitionSpec for the given partition type."""
     if partition == DataPartitionType.FULL:
         return input_partition_spec()
     elif partition == DataPartitionType.REPLICATED:
         return None
+    elif partition == DataPartitionType.BATCH:
+        return PartitionSpec(batch_axis_names)
```
kelvin-zou, line 604: Instead of directly assigning PartitionSpec, maybe extend `input_partition_spec ` with an argument batch_axis_names, with default None?

axlearn/common/host_array_test.py
```diff
@@ -37,16 +36,15 @@ def is_supported(
         )
     )
 
-
```
kelvin-zou, line 40: Are you using a different pylint style? Those lines should not be removed.

kelvin-zou:
Sorry for the late review, I should've checked this earlier. 
@apoorvtintin can you actually look into logical batch and use that instead? From what I see here, this feature is a bit redundant (and less flexible as well) than the logical batch size, and with that, probably you don't need this PR if you just want to run something like batch_size=0.5 or 0.25 per chip with model parallelism.

hanzhi713:
> Sorry for the late review, I should've checked this earlier. @apoorvtintin can you actually look into logical batch and use that instead (
> 
> https://github.com/apple/axlearn/blob/2d1fb295801e954d72400a4c69e066254ff0c84d/axlearn/common/input_dispatch.py#L24
> 
> )? From what I see here, this feature is a bit redundant (and less flexible as well) than the logical batch size, and with that, probably you don't need this PR if you just want to run something like batch_size=0.5 or 0.25 per chip with model parallelism.

I think the primary purpose of this PR (as stated in the description) is to avoid the resharding cost from physical bs to logical bs when there's no need to shard data along model axis. This could remove the collective for inputs re-sharding and thus reduce step time. 

@apoorvtintin May I know how much time do we save by removing this collective? If the benefit is not significant, we can stick to the old code.

kelvin-zou:
> I think the primary purpose of this PR (as stated in the description) is to avoid the resharding cost from physical bs to logical bs when there's no need to shard data along model axis. This could remove the collective for inputs re-sharding and thus reduce step time.

But input collective is only triggered within host, which should be insignificant?

apoorvtintin:
> @apoorvtintin With this new sharding strategy, how do you plan to support model parallelism across hosts (i.e. per host bs < 1)? If this is not a supported use case, then this change LGTM. I could imagine the support for model parallelism across hosts is not needed in the short term for `neuron` since the number of chips per host is large.

@hanzhi713 Tensor parallelism is usually used within a device. But yes theoretically tensor parallelism can be across hosts with per host bs<1. To answer your question, yes this PR does not support that case.

apoorvtintin:
> Sorry for the late review, I should've checked this earlier. @apoorvtintin can you actually look into logical batch and use that instead (
> 
> https://github.com/apple/axlearn/blob/2d1fb295801e954d72400a4c69e066254ff0c84d/axlearn/common/input_dispatch.py#L24
> )? From what I see here, this feature is a bit redundant (and less flexible as well) than the logical batch size, and with that, probably you don't need this PR if you just want to run something like batch_size=0.5 or 0.25 per chip with model parallelism.

@kelvin-zou I checked out fractional batches (logical to physical batches). This feature pads '0' batches to get to at least 1 batch per device. Please correct me if I did not understand it correctly.

For sequence parallelism we need all model axis ranks in a group to have replicated batches. Yes theoretically it means we have only 0.25 batches per device (when fsdp=16 and model=4 ==> 16 batches). but the batches are replicated along the model axis. If I use fractional batches with set batches per device to 0.25 the I would get 16 batches but the rest 48 will be padded 0s.

apoorvtintin:
> > Sorry for the late review, I should've checked this earlier. @apoorvtintin can you actually look into logical batch and use that instead (
> > https://github.com/apple/axlearn/blob/2d1fb295801e954d72400a4c69e066254ff0c84d/axlearn/common/input_dispatch.py#L24
> > 
> > )? From what I see here, this feature is a bit redundant (and less flexible as well) than the logical batch size, and with that, probably you don't need this PR if you just want to run something like batch_size=0.5 or 0.25 per chip with model parallelism.
> 
> I think the primary purpose of this PR (as stated in the description) is to avoid the resharding cost from physical bs to logical bs when there's no need to shard data along model axis. This could remove the collective for inputs re-sharding and thus reduce step time.

Correct me if I am wrong, Axlearn currently does not allow input batches to be sharded along the sequence dimension which can be sharded along the model axis. Which means without this PR the only choices are:
- Fractional batches: May not be enough as described in my [previous reply](https://github.com/apple/axlearn/pull/884#issuecomment-2591751866).
- Sharding REPLICATED: Not feasible as it uses too much memory.
- Sharding FULL: We shard the batch axis along all axis available including model. Which means GBS == global_device_count. For optimal sequence parallel training we want GBS == FSDP*DP ranks.

Which brings us to this Sharding BATCH which is implemented by this PR. This allows us to have GBS == FSDP*DP ranks where batches are replicated across model axis.

> @apoorvtintin May I know how much time do we save by removing this collective? If the benefit is not significant, we can stick to the old code.
That depends on the sharding strategy chosen from the options above, if Sharding FULL is chosen then each model parallel group will have to run a reshard to sync batches (like a TP all gather). This example glosses over obvious problems with full sharding like higher batch size (4 times more than we can fit since model=4).

apoorvtintin:
> But input collective is only triggered within host, which should be insignificant?

@kelvin-zou I assume your question assumes we use SHARDING FULL, which has a batch size problem even though we could theoretically achieve batches replicated along 'model' with an extra collective.

axlearn/common/utils.py
markblee: Yes, support for `make_array_from_process_local_data` has been merged. There doesn't seem to be many remaining advantages of passing both a partition type and aux information like batch axis names; instead, we can take partition specs directly:
```
DataPartitionType.FULL -> input_partition_spec()
DataPartitionType.REPLICATED -> None
DataPartitionType.BATCH -> PartitionSpec(batch_axis_names)
```
This brings the API closer to jax native, but may require more changes internally.

axlearn/common/utils_test.py
```diff
@@ -1701,6 +1702,31 @@ def test_length(self):
 class HostToGlobalArrayTest(TestCase):
     """Tests host_to_global_device_array."""
 
+    @pytest.mark.neuron
+    def test_partition_batch(self):
```
markblee, line 1706: Are we able to pass the `test_every_other_process` only relying on `make_array_from_process_local_data`? If not, then it seems that input dispatch may still be needed.

markblee (commented):
Sorry for the late review.

hanzhi713:
> > @apoorvtintin With this new sharding strategy, how do you plan to support model parallelism across hosts (i.e. per host bs < 1)? If this is not a supported use case, then this change LGTM. I could imagine the support for model parallelism across hosts is not needed in the short term for `neuron` since the number of chips per host is large.
> 
> @hanzhi713 Tensor parallelism is usually used within a device. But yes theoretically tensor parallelism can be across hosts with per host bs<1. To answer your question, yes this PR does not support that case.

With fractional batches (logical batch), yes at first batches are padded to per device bs=1, but then it will be resharded to what the actual per device bs is here: https://github.com/apple/axlearn/blob/141239c3376638d551e10ff1094e6947e6a75f75/axlearn/common/trainer.py#L1139C38-L1139C59

This dispatch function will reshard physical batches to logical batches, e.g. ("data", "fsdp", "model") -> ("data", "fsdp") through an einsum. After this point, batches should be replicated along model parallel axis as you would expect.

apoorvtintin:
Closing in favor of fractional batches

PR #885, opened 2024-12-11
Add meshes and config for TRN2/1 for Fuji models

apoorvtintin:
This PR adds meshes for TRN2/1 for Fuji models and transformer layer configuration favorable to Neuron. 

Neuron supports stacked transformer and GroupedQKVLinear instead of FusedGroupedQKVLinear for Grouped Query Attention (GQA)

### DIFF
--- a/axlearn/experiments/text/gpt/fuji.py
+++ b/axlearn/experiments/text/gpt/fuji.py
@@ -15,6 +15,7 @@ import functools
 import itertools
 from typing import Any, Optional, Union
 
+import jax
 from jax.ad_checkpoint import checkpoint_policies as jax_remat_policies
 
 from axlearn.common import causal_lm, config
@@ -22,6 +23,7 @@ from axlearn.common.attention import (
     BaseStackedTransformerLayer,
     FusedGroupedQKVLinear,
     FusedQKVLinear,
+    GroupedQKVLinear,
     GroupedQueryAttention,
     MultiheadAttention,
     RepeatedTransformerLayer,
@@ -174,6 +176,12 @@ def get_trainer_kwargs(
             train_batch_size=train_batch_size,
             max_step=max_step,
             mesh_shape=mesh_shape_from_axes(data=-1, fsdp=8),
+            mesh_rules=(
+                (
+                    "neuron-(trn2|trn2n).48xlarge-64",
+                    mesh_shape_from_axes(fsdp=-1, model=4),
+                ),
+            )
         )
     elif model_size == "3B":
         trainer_kwargs = dict(
@@ -192,6 +200,12 @@ def get_trainer_kwargs(
             train_batch_size=train_batch_size,
             max_step=max_step,
             mesh_shape=mesh_shape_from_axes(data=-1, fsdp=8),
+            mesh_rules=(
+                (
+                    "neuron-(trn2|trn2n).48xlarge-64",
+                    mesh_shape_from_axes(fsdp=-1, model=4),
+                ),
+            )
         )
     elif model_size == "7B":
         trainer_kwargs = dict(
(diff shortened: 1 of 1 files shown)


axlearn/experiments/text/gpt/fuji.py
```diff
@@ -174,6 +176,12 @@ def get_trainer_kwargs(
             train_batch_size=train_batch_size,
             max_step=max_step,
             mesh_shape=mesh_shape_from_axes(data=-1, fsdp=8),
+            mesh_rules=(
+                (
+                    "neuron-(trn2|trn2n).48xlarge-64",
+                    mesh_shape_from_axes(fsdp=-1, model=4),
```
ruomingp, line 182: Comment on why we set `model=4` for neuron?

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -473,7 +504,9 @@ def model_config(
         ffn_dim = scaled_hidden_dim(scale=8 / 3, round_up_to_multiples_of=256)
     if num_kv_heads:
         atten_cfg = GroupedQueryAttention.default_config()
-        atten_input_linear = FusedGroupedQKVLinear.default_config().set(num_kv_heads=num_kv_heads)
+        backend = jax.default_backend()
```
ruomingp, line 507: The fuji config should not depend on `jax.default_backend()`, otherwise the golden configs will not reflect the actual config being used.

Instead, we can create separate configs for a backend that requires different settings.

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -417,12 +443,17 @@ def get_trainer_kwargs(
...
+                    "neuron-(trn2|trn2n).48xlarge-64",
+                    mesh_shape_from_axes(fsdp=-1, model=4),
+                ),
             ),
         )
     else:
         raise NotImplementedError(f"Unknown model size {model_size}.")
     model_kwargs = trainer_kwargs.pop("model_kwargs")
     model_kwargs.setdefault("vocab_size", vocab_size)
+    model_kwargs.setdefault("stack_cfg", None if backend != "neuron" else StackedTransformerLayer.default_config())
```
ruomingp, line 456: Will the use of `StackedTransformerLayer` (vs. `RepeatedTransformerLayer`) lead to large XLA programs and long compilation time?

ruomingp (changes requested):

axlearn/experiments/text/gpt/fuji.py
kelvin-zou: +1, please follow this example instead if you really need to overwrite some configs, you can add another custom `LayerConfigModifier`like this one: https://github.com/apple/axlearn/blob/main/axlearn/common/trainer_config_modifier.py#L69,

apoorvtintin: Thanks for the review, will update the PR with a custom LayerConfigModifier.

axlearn/experiments/text/gpt/fuji.py
apoorvtintin: We are in the middle of optimizing RepeatedTransformer to use a new hardware feature in TRN2 to make dynamic memory operations faster. In the meantime, please continue to use StackedTransformer. Neuron compiler has a module to detect repeating blocks, compile once and reuse. So, compile time won't grow with the number of layers.

ruomingp: > We are in the middle of optimizing RepeatedTransformer to use a new hardware feature in TRN2 to make dynamic memory operations faster. In the meantime, please continue to use StackedTransformer. Neuron compiler has a module to detect repeating blocks, compile once and reuse. So, compile time won't grow with the number of layers.

Nice! Could you add this as a comment?

apoorvtintin:
Opened a new PR from my fork of Axlearn (https://github.com/apple/axlearn/pull/916). All comments in this discussion have been addressed there.

apoorvtintin:
Closing since there is a new version of the PR in (https://github.com/apple/axlearn/pull/916)

PR #887, opened 2024-12-11
Add default compiler options for v6e

samos123:


axlearn/common/compiler_options.py
```diff
@@ -55,11 +110,20 @@ def default_xla_options(
             xla_tpu_data_parallel_opt_different_sized_ops="true",
             # Group non-blocking DCN collectives into as few stages as possible.
             xla_tpu_enable_sunk_dcn_allreduce_done_with_host_reduction="true",
+            # Change to 16GB. The default is 4GB which is too small for larger models. This
+            # cause the step time to be double. You should increase this
+            # further if you see "Allocator failed to allocate". A feature
+            # to dynamically allocate may come later: b/380514965
+            megascale_grpc_premap_memory_bytes=17179869184,
```
Ethanlm, line 117: This will affect all jobs. Is it a desired change? Should we limit it to v6e only?

axlearn/common/compiler_options.py
```diff
@@ -44,6 +44,61 @@ def default_xla_options(
...
+            # Uncomment below flags to enable it.
+            # xla_sc_disable_megacore_partitioning="true",
+            # xla_tpu_use_tc_device_shape_on_sc="true",
+            # tpu_use_continuations="true",
+            # xla_jf_crs_combiner_threshold_count=10,
+            # xla_sc_enable_instruction_fusion="false",
+            # xla_sc_disjoint_spmem="false",
+            # xla_tpu_enable_sparse_core_collective_offload_all_reduce="true",
+        )
+        options["2a886c8_chip_config_name"] = "megachip_tccontrol"
```
Ethanlm, line 101: I remember this is only required for sparsecore offloading. If so, should we disable it by default?

samos123: It's the new default in Jax nightly since Dec 10. It will become the default in Jax 0.4.38. Let me disable by default for now though and comment that it's only needed for sparsecore all reduce.

axlearn/common/compiler_options.py
samos123: I think this is important for all jobs. The only downside is that it takes host memory.

Ethanlm (approved):
Thanks

axlearn/common/compiler_options.py
samos123: Done, but I do still recommend enabling this for all jobs. Let's tackle that in separate PR.

axlearn/common/compiler_options.py
samos123: Done.

axlearn/common/compiler_options.py
```diff
@@ -31,7 +31,7 @@ def default_xla_options(
     if backend != "tpu":
         raise NotImplementedError(backend)
     version = infer_tpu_version(infer_tpu_type(instance_type))
-    options = dict(
+    options: Dict[str, int | str | bool] = dict(
```
Ethanlm, line 34:
```
File "/home/runner/work/axlearn/axlearn/axlearn/common/compiler_options.py", line 34, in default_xla_options: Invalid type annotation 'Dict[str, int | str | bool]'  [invalid-annotation]
  Missing parameter 'y' in call to function int.__or__
         Expected: (self, y)
  Actually passed: (self)
File "/home/runner/work/axlearn/axlearn/axlearn/common/compiler_options.py", line 48, in default_xla_options: Function dict.update was called with the wrong arguments [wrong-arg-types]
         Expected: (self, megascale_grpc_premap_memory_bytes, xla_enable_async_all_gather: int, ...)
  Actually passed: (self, megascale_grpc_premap_memory_bytes, xla_enable_async_all_gather: str, ...)

```

ruomingp (approved):

axlearn/common/compiler_options.py
```diff
@@ -44,6 +44,68 @@ def default_xla_options(
...
+        options.update(
+            # Change to 16GB. The default is 4GB which is too small for larger models. This
+            # cause the step time to be double. You should increase this
+            # further if you see "Allocator failed to allocate". A feature
+            # to dynamically allocate may come later: b/380514965
+            megascale_grpc_premap_memory_bytes=17179869184,
+            # Improved performance for v6e.
+            xla_tpu_scoped_vmem_limit_kib=98304,
+            xla_tpu_enable_async_collective_fusion=True,
+            xla_tpu_enable_async_collective_fusion_fuse_all_gather=True,
```
Ethanlm, line 57: Needs These need to be "true" instead of `True`, otherwise it will be converted to 1 and then fail

```
I1212 21:41:11.645763 132689403348992 launch.py:112] LIBTPU_INIT_ARGS='--xla_tpu_spmd_rng_bit_generator_unsafe=1 --xla_tpu_enable_latency_hiding_scheduler=true --xla_tpu_perform_spmd_cse_prevention=false --megascale_grpc_premap_memory_bytes=17179869184 --xla_tpu_scoped_vmem_limit_kib=98304 --xla_tpu_enable_async_collective_fusion=1 --xla_tpu_enable_async_collective_fusion_fuse_all_gather=1 --xla_tpu_enable_async_collective_fusion_multiple_steps=1 --xla_tpu_overlap_compute_collective_tc=1 --xla_enable_async_all_gather=1 --xla_tpu_enable_all_experimental_scheduler_features=1 --xla_tpu_enable_scheduler_memory_pressure_tracking=1 --xla_tpu_host_transfer_overlap_limit=24 --xla_tpu_aggressive_opt_barrier_removal=1 --xla_lhs_prioritize_async_depth_over_stall=1 --xla_tpu_enable_ag_backward_pipelining=1 --xla_should_allow_loop_variant_parameter_in_chain=1 --xla_should_add_loop_invariant_op_in_chain=1 --xla_max_concurrent_host_send_recv=100 --xla_tpu_scheduler_percent_shared_memory_limit=90 --xla_latency_hiding_scheduler_rerun=2 --xla_tpu_use_enhanced_launch_barrier=1'
2024-12-12 21:41:14.669321: I external/tsl/tsl/platform/default/grpc_credentials.cc:30] gRPC insecure client credentials are used.
I1212 21:41:14.670917 132689403348992 distributed.py:119] Connecting to JAX distributed service on ethanli-fuji-70b-v2-test1-job-0-0.ethanli-fuji-70b-v2-test1:8476
2024-12-12 21:41:14.999583: I external/xla/xla/pjrt/distributed/client.cc:135] Connected to distributed JAX controller
ERROR: Illegal value '1' specified for flag 'xla_tpu_enable_async_collective_fusion_fuse_all_gather'; expected one of true/enabled, false/disabled or auto
ERROR: Illegal value '1' specified for flag 'xla_enable_async_all_gather'; expected one of true/enabled, false/disabled or auto
ERROR: Illegal value '1' specified for flag 'xla_tpu_enable_scheduler_memory_pressure_tracking'; expected one of true/enabled, false/disabled or auto
ERROR: Illegal value '1' specified for flag 'xla_tpu_aggressive_opt_barrier_removal'; expected one of true/enabled, false/disabled or auto
ERROR: Illegal value '1' specified for flag 'xla_lhs_prioritize_async_depth_over_stall'; expected one of true/enabled, false/disabled or auto
ERROR: Illegal value '1' specified for flag 'xla_should_allow_loop_variant_parameter_in_chain'; expected one of true/enabled, false/disabled or auto
ERROR: Illegal value '1' specified for flag 'xla_should_add_loop_invariant_op_in_chain'; expected one of true/enabled, false/disabled or auto

```

samos123: that's what I had originally, but pytype doesn't like that and will fail. So I see 2 options:
1. Change behavior of axlearn that converts True boolean to 1 and let it return "true" instead.
2. Ignore pytype check

axlearn/common/compiler_options.py
```diff
@@ -44,6 +44,68 @@ def default_xla_options(
...
+            # when the scheduler thinks it goes out memory, it may not actually happen due
+            # to other factors controlled by other compiler passes, or the initial memory
+            # limit is already set too low. Cutting the memory limit to 90% of previous one
+            # though, may make the scheduler weighting too much on the memory usage instead
+            # of latency side.
+            xla_latency_hiding_scheduler_rerun=2,
+            xla_tpu_use_enhanced_launch_barrier=True,
+            # Sparsecore offloading for all reduce.
+            # Uncomment below flags to enable it.
+            # xla_sc_disable_megacore_partitioning=True,
```
kelvin-zou, line 98: @Ethanlm  do you recall why you commented all them off?

samos123: We weren't sure if this would perform better for everyone. It's a newer feature as well. So Ethan recommended not enabling by default. For both llama 2 70b and 405b, we saw much better performance with it being enabled though.

PR #888, opened 2024-12-12
MaskFnAttentionBias._bool_value passes the same rank position tensors to mask_fn.

ds-hwang:
MaskFnAttentionBias._bool_value passes the same rank position tensors to mask_fn.

When target_positions is set, a rank 3 target_positions and a rank 2
source_positions are passed to mask_fn. From the perspective of a downstream
defining mask_fn, this is a big surprise.


ds-hwang:
From 950

apghml (changes requested):
This PR is under discussion internally.

apghml (approved):
Thanks!

ruomingp (approved):

ds-hwang:
Thank you for review!

PR #890, opened 2024-12-12
use "true" and "false" instead of 0 and 1

samos123:
Jax/XLA seem to expect true and false strings instead of using 0 and 1.

### DIFF
--- a/axlearn/common/compiler_options.py
+++ b/axlearn/common/compiler_options.py
@@ -138,7 +138,7 @@ def xla_flags_from_options(xla_options: dict[str, Union[str, bool, int]]) -> str
     flags = []
     for k, v in xla_options.items():
         if isinstance(v, bool):
-            v = "1" if v else "0"
+            v = "true" if v else "false"
         flags.append(f"--{k}={v}")
     return " ".join(flags)
 


ruomingp (approved):
Thanks! Is there a test that can added or updated?

markblee (approved):
+1 on adding test case.

axlearn/common/compiler_options_test.py
```diff
@@ -24,10 +24,10 @@ def f(x: Tensor) -> Tensor:
         )
         self.assertEqual(f_compiled(5), 15)
 
-    def atest_xla_flags_from_options(self):
```
samos123, line 27: this test was never being run because of the `a` before `test` in function name.

samos123:
I updated the existing test and renamed the test so it's being run correctly now.

axlearn/common/compiler_options.py
```diff
@@ -144,7 +144,7 @@ def xla_flags_from_options(xla_options: dict[str, Union[str, bool, int]]) -> str
     flags = []
     for k, v in xla_options.items():
         if isinstance(v, bool):
-            v = "1" if v else "0"
+            v = "true" if v else "false"
```
apghml, line 147: Should we convert all the "true" and "false" strings in `default_xla_options` to bools to be consistent?

apghml, line 147: Would it make sense to do an aot compilation or run the trainer for a few steps to confirm these changes work? (If you haven't already?)

axlearn/common/compiler_options.py
```diff
@@ -144,7 +144,7 @@ def xla_flags_from_options(xla_options: dict[str, Union[str, bool, int]]) -> str
...
         flags.append(f"--{k}={v}")
```
apghml, line 148: Could you clarify whether there are two types of flags, bool and tristate, or just one type, tristate?

samos123: there are 2 types of flags, but a bool should be converted to "true" or "false" so it can be used for either type of flag.

axlearn/common/compiler_options.py
samos123: I did test it on v6e afair last year, but any additional runs would be great.

samos123: I think that should be preferred. I just wanted to keep this PR small to make merging and reviewing easier. Happy to include that change in this PR or do a follow up.

apghml: Maybe test it again on v5p using the current jax/jaxlib version pinned in AXLearn?

apghml: Thanks, I think it makes sense to include in this PR.

markblee: Were you able to run this?

markblee (approved):

axlearn/common/compiler_options.py
samos123: No I don't have readily access to all TPUs. I only have occasional access. Is there no CI that runs tests on PRs so we can verify changes like this during pre-submit?

samos123:
It's still relevant. @muyangyuapple encountered the same issue on pathways. But feel free to discard this and make a separate PR.

tl;dr we should always true or false instead of 1 or 0 for a flag is boolean like.

PR #898, opened 2024-12-17
Special remat for Neuron

apoorvtintin:
This PR adds special remat configuration for TRN2 and Fuji-70B. This is done by adding a new remat policy that uses regex to match against multiple regex patterns for remat names. This allows more flexibility in remat configurations for different backends and device types.

Misc: Enable remat for StackedTransformer


axlearn/common/utils.py
```diff
@@ -1459,3 +1459,30 @@ def validate_contains_paths(x: Nested[Tensor], paths: Sequence[str]):
                 f"Input is expected to contain '{path}'; "
                 f"instead, it contains: '{jax.tree_structure(x)}'."
             ) from e
+
+
+def save_only_these_regex_patterns(*regex_patterns_to_save):
```
apghml: Add an annotation for the args and return type?

axlearn/common/utils.py
```diff
@@ -1459,3 +1459,30 @@ def validate_contains_paths(x: Nested[Tensor], paths: Sequence[str]):
...
+    """Save only the values that match the regex pattern.
+
+    Args:
+        regexes_to_save: List of regex patterns to save
+
+    Returns:
+        Callable: Policy that matches regex
```
apghml: Nit: we include type information in the function signature, not the docstring.

axlearn/experiments/text/gpt/common.py
```diff
@@ -277,8 +269,7 @@ def model_config(
         layer_cfg.self_attention.attention.input_linear = attention_qkv_linear
     layer_cfg.self_attention.structure = atten_structure
     layer_cfg.self_attention.attention.atten_logit_cap = atten_logit_cap
-    if stack_cfg.klass is RepeatedTransformerLayer:
-        update_model_remat_config(stack_cfg=stack_cfg, layer_cfg=layer_cfg)
+    update_model_remat_config(stack_cfg=stack_cfg, layer_cfg=layer_cfg)
```
apghml, line 272: Could you explain why we originally only did this for RepeatedTransformer and why it is okay to do it for everything now?

axlearn/common/utils.py
```diff
@@ -1459,3 +1459,30 @@ def validate_contains_paths(x: Nested[Tensor], paths: Sequence[str]):
                 f"Input is expected to contain '{path}'; "
                 f"instead, it contains: '{jax.tree_structure(x)}'."
             ) from e
+
+
+def save_only_these_regex_patterns(*regex_patterns_to_save):
+    """Save only the values that match the regex pattern.
+
+    Args:
+        regexes_to_save: List of regex patterns to save
```
ruomingp: Comment on what these regex match?

axlearn/common/attention.py
```diff
@@ -3934,6 +3935,33 @@ def forward(
 _SavePattern = Union[str, re.Pattern, None]
 
 
+def save_only_these_regex_patterns(*regex_patterns_to_save):
```
ruomingp, line 3938: How is this different from https://github.com/apple/axlearn/blob/92205bc8928adaef14b980f7e4dd4936ee953e2f/axlearn/common/attention.py#L3933-L3956?

axlearn/common/attention.py
```diff
@@ -3934,6 +3935,33 @@ def forward(
...
+    Returns:
+        Callable: Policy that matches regex
+    """
+    regex_patterns_to_save = frozenset(regex_patterns_to_save)
+
+    def policy(*_, **params):
+        if "name" in params:
+            param_name = params["name"]
+            for regex_to_save in regex_patterns_to_save:
+                if re.search(regex_to_save, param_name):
```
ruomingp, line 3953: Use re.fullmatch for consistency with other AXLearn code?

ruomingp (changes requested):

axlearn/common/attention.py
kelvin-zou: +1, you should probably move 
[axlearn/axlearn/common/attention.py](https://github.com/apple/axlearn/blob/92205bc8928adaef14b980f7e4dd4936ee953e2f/axlearn/common/attention.py#L3933-L3956)

into https://github.com/apple/axlearn/blob/92205bc8928adaef14b980f7e4dd4936ee953e2f/axlearn/common/utils.py#L129 and make it part of extended_remat_policies, if you need it outside of attention.py.

axlearn/experiments/text/gpt/common.py
kelvin-zou: I think it was used to avoid applying to pipelinedTransformer, where we may have explosion of activation memories due to saving intermediaries, and both stacked and repeated transformer are safe to apply the same remat thing here.

apghml: We still have PipelinedTransformer in the codebase. Will this increase the memory usage of it?

axlearn/common/attention.py
apoorvtintin: I agree on reusing this policy, will also modify this policy to ingest multiple regex patterns as a sequence. I feel It is hard to put all remat names in just a single regex pattern.

axlearn/experiments/text/gpt/common.py
kelvin-zou: Technically, it is gated at a later stage, so it may be fine here. https://github.com/apple/axlearn/blob/main/axlearn/common/attention.py#L3994

apghml: IIUC, you're saying the check this PR removes may never have been needed at all? Is this PR urgent? Ideally, it would be good to check with Mark.

apghml: Or can we defensively leave the check for now with a TODO to revisit?

axlearn/common/attention.py
hanzhi713: Can you elaborate on why it's hard to put all names into a single regex? You can use "|" to separate patterns and use "()" to make things more readable.

apoorvtintin: Yes, by hard I meant readable. It is this
regex_patterns_to_save=[
    r"TransformerAttentionLayer\.residual_add",
    r"\.?(k|q|v)_proj$",
    r"\.?linear1_0",
    r"\.?linear1_1",
    r"TransformerFeedForwardLayer\.mlp_residual",
]
vs 
r"TransformerAttentionLayer\.residual_add|\.?(k|q|v)_proj$|\.?linear1_[01]|TransformerFeedForwardLayer\.mlp_residual"

Second one keeps the changes minimal, but the first may be a little cleaner. let me know the decision on this, I am okay with both

hanzhi713: I think we should reuse the API of `_save_and_offload_only_these_names_regex ` and avoid changing it since it appears in golden configs. Also cc @ruomingp for some input.

kelvin-zou: `r"TransformerAttentionLayer\.residual_add|\.?(k|q|v)_proj$|\.?linear1_[01]|TransformerFeedForwardLayer\.mlp_residual"  `, regarding this part, is it possible to make it a few substrings and then concat them and make one, so that it is more readable while not creating a new API?

apoorvtintin: I made updates to the PR to use `_save_and_offload_only_these_names_regex` as we discussed. The changes look clean. I put each different remat regex patterns neatly into different lines.
```
  names_which_can_be_saved=r"(TransformerAttentionLayer\.residual_add"  # pylint: disable=C0301
  "|.*\.?(k|q|v)_proj"
  "|.*\.?linear1_[01]"
  "|TransformerFeedForwardLayer\.mlp_residual)",
```

axlearn/common/utils.py
apoorvtintin: Removed this function from the PR

axlearn/common/utils.py
apoorvtintin: Removed this function in favor of `_save_and_offload_only_these_names_regex`

apoorvtintin:
Thanks for the reviews, most comments are resolved and PR looks clean, let me know if more changes are needed.

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -417,6 +418,36 @@ def get_trainer_kwargs(
...
+                                    "model.decoder.transformer.layer": RematSpec(
+                                        prevent_cse=True,
+                                        policy=config_for_function(
+                                            _save_and_offload_only_these_names_regex
+                                        ).set(
+                                            # pylint: disable=anomalous-backslash-in-string
+                                            names_which_can_be_saved=r"(TransformerAttentionLayer\.residual_add"  # pylint: disable=C0301
+                                            "|.*\.?(k|q|v)_proj"
+                                            "|.*\.?linear1_[01]"
+                                            "|TransformerFeedForwardLayer\.mlp_residual)",
```
ruomingp, line 439: Consider defining it as a constant so that it can be reused across configs?

ruomingp, line 439: Nit: we don't really need the outer most "(...)" and can do:

"|".join(["...", "...", "..."])

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -417,6 +418,36 @@ def get_trainer_kwargs(
...
+                                            names_which_can_be_offloaded=None,
```
ruomingp, line 440: Does Neuron support offloading activations?

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -417,6 +418,36 @@ def get_trainer_kwargs(
...
+                                mesh_shape=mesh_shape_from_axes(fsdp=-1, model=4)
+                            ),
+                            RematSpecModifier.default_config().set(
+                                remat_policies={
+                                    "model.decoder.transformer.layer": RematSpec(
+                                        prevent_cse=True,
+                                        policy=config_for_function(
+                                            _save_and_offload_only_these_names_regex
+                                        ).set(
+                                            # pylint: disable=anomalous-backslash-in-string
```
ruomingp, line 435: Do we need this?

axlearn/experiments/text/gpt/fuji.py
apoorvtintin: Added them as a Enum class

apoorvtintin: Done

axlearn/experiments/text/gpt/fuji.py
apoorvtintin: Yes don't need this since I use raw strings now.

axlearn/experiments/text/gpt/fuji.py
ptoulme-aws: Not yet.

axlearn/experiments/text/gpt/common.py
ruomingp: How about adding
```suggestion
    if issubclass(stack_cfg.klass, (RepeatedTransformerLayer, StackedTransformerLayer)):
        update_model_remat_config(stack_cfg=stack_cfg, layer_cfg=layer_cfg)
```

ruomingp (commented):
A suggestion to address the concern from @apghml. WDYT?

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -86,6 +87,14 @@ class Version(enum.Enum):
 }
 
 
+# Regex patterns for matching remat names
+class RematRegex(enum.Enum):
+    QKV_PROJ = r".*\.?(k|q|v)_proj"
+    LINEAR1_X = r".*\.?linear1_[01]"
+    RESIDUAL_ADD = r"(TransformerAttentionLayer\.residual_add"
```
hanzhi713, line 94: Mismatched parentheses?

apoorvtintin: thanks for pointing that out

apoorvtintin:
Thanks for the guidance, I addressed all comments. Let me know if more changes are needed

axlearn/audio/evaler_asr_test.py
```diff
@@ -133,13 +133,13 @@ def _compute_metrics(
...
-        cfg: WordErrorRateMetricCalculator.Config = (
-            WordErrorRateMetricCalculator.default_config().set(
-                vocab=config_for_class(seqio.SentencePieceVocabulary).set(
-                    sentencepiece_model_file=vocab_file,
-                ),
-                model_method_kwargs=decode_kwargs,
-            )
+        cfg: (
+            WordErrorRateMetricCalculator.Config
+        ) = WordErrorRateMetricCalculator.default_config().set(
```
ruomingp, line 138: Is this change intended?

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -86,6 +87,14 @@ class Version(enum.Enum):
 }
 
 
+# Regex patterns for matching remat names
+class RematRegex(enum.Enum):
+    QKV_PROJ = r".*\.?(k|q|v)_proj"
+    LINEAR1_X = r".*\.?linear1_[01]"
+    RESIDUAL_ADD = r"TransformerAttentionLayer\.residual_add"
+    MLP_RESIDUAL = r"TransformerFeedForwardLayer\.mlp_residual"
```
ruomingp, line 95: How do we check whether these regex match the right activations?

Consider adding a test as in https://github.com/apple/axlearn/blob/a15a3bcbb976c14db157a8958df368a48c614c1f/axlearn/common/attention_test.py#L3831-L3876.

axlearn/audio/evaler_asr_test.py
apoorvtintin: No looks like precommit flow made this change let me revert this

axlearn/experiments/text/gpt/fuji.py
apoorvtintin: Added a test

axlearn/common/attention.py
```diff
@@ -2801,6 +2805,7 @@ def _linear2(x):
         self._add_tensor_stats("inputs", inputs)
 
         remat_pt2 = "linear2"
+        remat_pt3 = "feed_forward_output"
```
kelvin-zou, line 2808: hmm, is this necessary? Seems quite wasteful to checkpoint an activation after a norm? I would recommend restraining from making change here.

axlearn/common/attention.py
```diff
@@ -2508,23 +2508,27 @@ def attention_thunk(target: Tensor) -> tuple[Optional[NestedTensor], Tensor]:
                 atten_state, atten_output = attention_thunk(TensorSpec(target.shape, target.dtype))
                 return dict(attention=atten_state), atten_output
 
+        remat_pt1 = "attention_output"
```
kelvin-zou, line 2511: Is this really needed? How much perf gain we get from saving `attention_output`?

kelvin-zou (commented):
High level comment, probably not to add more activation checkpoints in attention transformer layer class unless we must do so, otherwise the change looks good to me. Will approve once revert that part.

ruomingp (commented):
Please address @kelvin-zou 's comments. Otherwise LGTM.

apoorvtintin:
I removed the extra remat save points as discussed. Please let me know if more changes are needed. Thank you!

axlearn/common/attention_test.py
```diff
@@ -3875,6 +3875,72 @@ def f(x, layer_params):
...
+        spec: Any = build_remat_spec(mock.MagicMock())
+
+        policy = (
+            config_for_function(_save_and_offload_only_these_names_regex)
+            .set(
+                names_which_can_be_saved="|".join(
+                    [
+                        RematRegexSavePatterns.QKV_PROJ.value,
+                        RematRegexSavePatterns.LINEAR1_X.value,
+                        RematRegexSavePatterns.ATTENTION_OUTPUT.value,
```
hanzhi713, line 3914: It looks like the names here doesn't match those listed in `RematRegexSavePatterns`

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -417,6 +418,38 @@ def get_trainer_kwargs(
...
+                                        prevent_cse=True,
+                                        policy=config_for_function(
+                                            _save_and_offload_only_these_names_regex
+                                        ).set(
+                                            names_which_can_be_saved="|".join(
+                                                [
+                                                    RematRegexSavePatterns.QKV_PROJ.value,
+                                                    RematRegexSavePatterns.LINEAR1_X.value,
+                                                    RematRegexSavePatterns.RESIDUAL_ADD.value,
+                                                    RematRegexSavePatterns.MLP_RESIDUAL.value,
```
hanzhi713, line 440: Same here

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -417,6 +418,36 @@ def get_trainer_kwargs(
...
+                                remat_policies={
+                                    "model.decoder.transformer.layer": RematSpec(
+                                        prevent_cse=True,
+                                        policy=config_for_function(
+                                            _save_and_offload_only_these_names_regex
+                                        ).set(
+                                            names_which_can_be_saved="|".join(
+                                                [
+                                                    RematRegexSavePatterns.QKV_PROJ.value,
+                                                    RematRegexSavePatterns.LINEAR1_X.value,
```
hanzhi713, line 438: How is this different from `dots_saveable`?

apoorvtintin: We are not saving all dots, for example out_proj. More saves would use more memory. Flexibility in remat is needed to ensure the model fits. in memory

axlearn/experiments/text/gpt/fuji.py
apoorvtintin: Thank you fixed both, and the test.

axlearn/common/attention.py
```diff
@@ -3956,15 +3956,19 @@ def policy(prim, *_, **params):
...
 
 
-SELF_ATTENTION_SAVE_PATTERN = ".*([qkvo]_proj|context)"
-FEED_FORWARD_SAVE_PATTERN = ".*linear[12]_.*"
+# Regex patterns for matching remat names
+class RematRegexSavePatterns(enum.Enum):
+    QKV_PROJ = r".*\.?(k|q|v)_proj"
+    LINEAR1_X = r".*\.?linear1_[01]"
+    SELF_ATTENTION = ".*([qkvo]_proj|context)"
+    FEED_FORWARD = ".*linear[12]_.*"
```
kelvin-zou, line 3964: Hmm, can we make the regex more composable, e.g., seems like `QKV_PROJ` is a subset of `SELF_ATTENTION`, and `LINEAR1_X` is a subset of `FEED_FORWARD`. Otherwise, we may get into an exponential number of save patterns. 

You can try to build something similar to the following. 
```suggestion
class RematRegexSavePatterns(enum.Enum):
    QKV_PROJ = r".*\.?(k|q|v)_proj"
    O_PROJ=r".*o_proj"
    CONTEXT=".*context"
    SELF_ATTENTION = re.compile( '|'.join( [O_PROJ,QKV_PROJ, CONTEXT]) )
```

hanzhi713 (approved):

axlearn/common/attention.py
apoorvtintin: Added this, thanks

apoorvtintin:
Updated the PR with the requested changes. thanks everyone! Let me know if this PR is ready to merge.

ruomingp (commented):
Hi @hanzhi713 and @kelvin-zou , since you are updating remat policy code, could you take a look at this PR? Thanks!

hanzhi713:
@ruomingp This change LGTM, except that we'll need to update the golden configs for all internal models too. Should we avoid this? cc @kelvin-zou

kelvin-zou:
> @ruomingp This change LGTM, except that we'll need to update the golden configs for all internal models too. Should we avoid this? cc @kelvin-zou

+1, we should avoid golden config change, it would make rebase much more difficult.

axlearn/common/attention.py
```diff
@@ -3956,15 +3956,22 @@ def policy(prim, *_, **params):
...
-SELF_ATTENTION_SAVE_PATTERN = ".*([qkvo]_proj|context)"
-FEED_FORWARD_SAVE_PATTERN = ".*linear[12]_.*"
+# Regex patterns for matching remat names
+class RematRegexSavePatterns(enum.Enum):
+    QKV_PROJ = r".*[kqv]_proj"
+    O_PROJ = r".*o_proj"
+    CONTEXT = r".*context"
+    LINEAR1_X = r".*linear1_[01]"
+    LINEAR2_X = r".*linear2_[01]"
+    SELF_ATTENTION = re.compile("|".join([QKV_PROJ, O_PROJ, CONTEXT])).pattern
```
hanzhi713, line 3966:
```suggestion
    SELF_ATTENTION = ".*([qkvo]_proj|context)"
```

We prefer keeping this unchanged to avoid updating golden config.

apoorvtintin: Done

apoorvtintin:
Addressed final rounds of comments, thank you!

kelvin-zou (approved):
Thank you for addressing all the comments!

hanzhi713 (approved):

apoorvtintin:
Hello @ruomingp, looks like merge is blocked on your requested changes, can you please mark them as resolved so this PR can be merged. Thank you!

ruomingp (approved):

apoorvtintin:
Updated PR to avoid merge conflicts. Let's merge this soon, thanks everyone!

axlearn/common/attention_test.py
```diff
@@ -3901,6 +3901,70 @@ def f(x, layer_params):
...
+                is_training=True,
+                prng_key=jax.random.PRNGKey(0),
+            )
+            return forward_outputs
+
+        # Ignore type errors.
+        spec: Any = build_remat_spec(mock.MagicMock())
+
+        policy = (
+            config_for_function(_save_and_offload_only_these_names_regex)
```
kelvin-zou, line 3934: Can you use the one in utils.py instead?

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -19,14 +19,15 @@
...
-    SELF_ATTENTION_SAVE_PATTERN,
     BaseStackedTransformerLayer,
     FusedGroupedQKVLinear,
     FusedQKVLinear,
     GroupedQueryAttention,
     MultiheadAttention,
+    RematRegexSavePatterns,
     RepeatedTransformerLayer,
     RoFormerQKVLinear,
+    _save_and_offload_only_these_names_regex,
```
kelvin-zou, line 30: `_*` functions should not be imported, use the one in utils.py instead please.

apoorvtintin:
@kelvin-zou thanks for the review, I resolved your comments.

axlearn/common/attention_test.py
```diff
@@ -58,14 +57,15 @@
...
+    RematRegexSavePatterns,
     RepeatedTransformerLayer,
     RoFormerQKVLinear,
     StackedTransformerLayer,
     TransformerAttentionLayer,
     TransformerFeedForwardLayer,
     TransformerLayer,
     _next_power_of_two,
-    _save_and_offload_only_these_names_regex,
+    save_and_offload_only_these_names_regex,
```
kelvin-zou, line 68: Remove this one as well?

axlearn/common/attention.py
```diff
@@ -4001,15 +4001,22 @@ def _save_and_offload_only_these_names_regex(
...
-FEED_FORWARD_SAVE_PATTERN = ".*linear[12]_.*"
+# Regex patterns for matching remat names
+class RematRegexSavePatterns(enum.Enum):
+    QKV_PROJ = r".*[kqv]_proj"
+    O_PROJ = r".*o_proj"
+    CONTEXT = r".*context"
+    LINEAR1_X = r".*linear1_[01]"
+    LINEAR2_X = r".*linear2_[01]"
+    SELF_ATTENTION = ".*([qkvo]_proj|context)"
+    FEED_FORWARD = "|".join([LINEAR1_X, LINEAR2_X])
```
kelvin-zou, line 4012: `".*linear[12]_.*"` is not equivalent to `.*linear1_[01]|.*linear2_[01]` right? there is also a `.*linear1_[02]` if I remember correctly?

axlearn/common/attention_test.py
```diff
@@ -3901,6 +3902,70 @@ def f(x, layer_params):
...
+            layer_params,
+        )
+        _, full_remat_backward = jax.linearize(
+            jax.remat(f),
+            jnp.asarray(target),
+            layer_params,
+        )
+
+        # Eliminated the remat of qkv_proj and linear1_0 = 4 dots. This assumes
+        # FlashAttention is not enabled.
```
kelvin-zou, line 3962: Does flash attention play any role here? They should have the same diff w or w/o flash?

hanzhi713: I believe this comment is copied from one of my unit tests without changes. Indeed, here flash attention shouldn't matter and the comment is no longer relevant.

apoorvtintin: flash attention in neuron kernel interface (NKI) replaces all of the attention dots with a custom-call op. So the dot count will be different if flash attention is enabled.

hanzhi713: @apoorvtintin Wouldn't the difference still be the same?

hanzhi713: i.e. here we're calculating the difference in the number of dots before and after remat.

apoorvtintin: Oh yeah makes sense, flash attention kernel does not replace the input projections. Removing this comment

apoorvtintin:
Addressed all comments. Thanks for the review @kelvin-zou @hanzhi713

hanzhi713 (approved):

hanzhi713:
Thanks. In the meantime, could you address comments in #884?

hanzhi713:
@apoorvtintin Can you fix the pre-commit errors?

apoorvtintin:
@hanzhi713 Fixed the pre-commit errors

kelvin-zou (approved):

apoorvtintin:
I see test `TfIteratorTest.test_no_save_input_iterator ` failed, This test passes in my development environment. Any idea why it fails in the workflow?

hanzhi713:
The test itself is poorly written @markblee. It didn't wait for async future to finish. I will send a PR to fix it. 

Is there a way to retry the CI?

hanzhi713:
See #924

PR #899, opened 2024-12-18
Implements FlashDecoding with Sparsity Support

hanzhi713:
**To be merged in 2025**

This PR implements GPU FlashDecoding. See the docstring of `gpu_decoding.py` for description of the kernel.

Also, `gpu_attention_benchmark.py` is completely rewritten to use Jax mosaic profiling tools. Previously, `triton` testing suite is used, which gives meaningless results as triton uses cudaEvent that records to the default stream. That is not the cudaStream that jax uses.


apghml (dismissed):
It looks like there is still an unresolved internal comment.

ruomingp (approved):

PR #900, opened 2024-12-19
Learner: add new unittests using Model.

ds-hwang:
These tests are similar how trainer.py uses Learner, which demonstrates how API users should use the API while ensuring that it works correctly.


ds-hwang:
@ruomingp could you review? From 959

apghml (approved):

PR #904, opened 2024-12-30
Remove version pin of typing-extensions

wangkuiyi:
The pinned version is pretty old. When I am using a modern pydantic, it complains the need of a newer version of typing-extentions.

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
pydantic 2.10.1 requires typing-extensions>=4.12.2, but you have typing-extensions 4.11.0 which is incompatible.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ core = [
     "tensorflow_text==2.16.1; platform_machine == 'x86_64'", # implied by seqio, but also used directly for text processing
     "tensorstore>=0.1.63",  # used for supporting GDA checkpoints
     "toml",  # for config management
-    "typing-extensions==4.11.0",
+    "typing-extensions",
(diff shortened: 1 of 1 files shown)


ruomingp (commented):
Thanks. Is there an internal PR that passes CI?

wangkuiyi:
Yes, @ruomingp . The internal PR 966 passed the CI.

wangkuiyi:
I apologize that I couldn't reproduce the issue that trigger me to file this pull request. When I tried to install Axlearn and projects depending on it, I got typing_extensions==4.11.0 that works well with pydantic==2.9.2.

```
pydantic                       2.9.2
pydantic_core                  2.23.4
typing_extensions              4.11.0
```

PR #906, opened 2025-01-06
`AdaptiveLayerNormModulation` now supports sequence conditions.

ds-hwang:
When using seq2seq models with DiT, the condition may have the same sequence length as the input.
For example:
- Input shape: `[batch, seq_len, dim]`
- Condition shape: `[batch, seq_len, cond_dim]`

The module has been modified to support such cases as well.


ds-hwang:
Could you review? From 969

ruomingp (approved):

PR #908, opened 2025-01-07
Optimize TPU Flash Attention (20x XLA compilation speed-up on 32k long context)

ds-hwang:
Support GAN loss to CompositeLearner.

Currently, `CompositeLearner` computes gradients for all sublayers using the
default `loss` from `forward`. However, in cases like GANs, the discriminative
network requires gradients computed using the discriminative loss for parameter
updates.

To generalize this functionality, a `sublearner_to_loss` flag is added to
`CompositeLearner`. This flag maps a sublearner name to its corresponding loss
name for gradient computation. `model.forward` returns `tuple(loss, aux)`, and
the loss name picks which loss to use.
- If set to `None`, all sublearners use the default loss (the first value of the tuple).
- If the value is `"loss"`, the default loss is used.
- For a specific loss name, it retrieves the loss from the `aux` dict
  (e.g., aux["loss_name"]). The specified loss must be a scalar float Tensor.
- Sublearner names not explicitly mapped are automatically assigned to the default loss.

Example: For a GAN, users can configure `CompositeLearner.Config` as follows:
```
cfg = CompositeLearner.default_config().set(
    rules=[(".*encoder.*", "encoder"), (".*decoder.*", "decoder")],
    learners={
        "encoder": Learner.default_config(),
        "decoder": Learner.default_config(),
    },
    sublearner_to_loss=dict(encoder="loss", decoder="discriminative_loss"),
)
```

The implementation calculates gradients in `forward_and_backward()` by invoking
`value_and_grad` for each loss name and its corresponding parameters. Updates
are then merged.

### DIFF
--- a/axlearn/common/attention_bias.py
+++ b/axlearn/common/attention_bias.py
@@ -687,7 +687,11 @@ def sliding_window_causal_mask(sliding_window_size: int) -> MaskFn:
     def mask(query_position: Tensor, key_position: Tensor):
         return query_position - key_position <= sliding_window_size
 
-    return and_masks(causal_mask, mask)
+    fun = and_masks(causal_mask, mask)
+    # Flash attention needs to recognize sliding window size in _to_splash_mask().
+    # pylint: disable-next=protected-access
+    fun._sliding_window_size = sliding_window_size
+    return fun
 
 
 def make_causal_biases(seq_len: int) -> Tensor:
--- a/axlearn/common/flash_attention/tpu_attention.py
+++ b/axlearn/common/flash_attention/tpu_attention.py
@@ -42,15 +42,16 @@ from axlearn.common.utils import Tensor
 
 
 def tpu_flash_attention(
-    query: Tensor,  # [batch_size, source_len, num_heads, head_dim]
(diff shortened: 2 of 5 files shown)


ds-hwang:
@apghml could you review? From 916

apghml (approved):

ds-hwang:
Thank you for review!

hanzhi713:
@ds-hwang nit: the PR description is wrong (perhaps copy-pasted from a different PR?).

ruomingp:
Wait, how is this PR approved and merged? I'm not sure that the discussion has been resolved on the internal PR.

apghml:
The PR description is wrong. This is not related to gans. The title is correct.

apghml:
Oh, it looks like you hadn't approved this PR internally either though. I mistakenly thought you had.

ds-hwang:
Sorry for the confusion. The PR that was merged is the Flash Attention PR, not the GAN loss PR. I copied and pasted the content from the internal PR by mistake. Fortunately, the PR title and commit description that were merged are correct, and I’ve now updated the body text.

PR #909, opened 2025-01-08
Introduce the scale enum flag in Embedding layer for LLM embedding.

ds-hwang:
The activation component should roughly have a magnitude of 1. Since the embedding tensor is initialized with a scale of `1/sqrt(dim)`, the activation is multiplied by `sqrt(dim)` to maintain the desired scale. e.g. Gemma [1]
[1] https://github.com/google-deepmind/gemma/blob/0d6ae857591248422127ca14c027909546362e6a/gemma/modules.py#L80

In addition, unsloth [2] discovered that `sqrt(dim)` needs to be computed in float32. [2] Sec 3 in https://unsloth.ai/blog/gemma-bugs

TODO(axlearn-team): Use UNIT scale enum for AFM+. This will require re-sweeping hyperparameters (e.g., learning rate).

### DIFF
--- a/axlearn/common/layers.py
+++ b/axlearn/common/layers.py
@@ -18,6 +18,7 @@
 """Basic layers."""
 
 import enum
+import math
 from collections.abc import Sequence
 from typing import Any, Callable, Optional, Union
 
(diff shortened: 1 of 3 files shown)


ds-hwang:
@ruomingp Could you review? From 970

axlearn/common/layers.py
```diff
@@ -781,6 +782,21 @@ class Embedding(BaseLayer):
...
+        1. **UNIT**: Scale the activation components to ~1.
+
+        The activation component should roughly have a magnitude of 1. Since the embedding tensor is
+        initialized with a scale of `1/√dim`, the activation is multiplied by `√dim` to
+        maintain the desired scale. e.g. Gemma [1]
+        [1]
+        https://github.com/google-deepmind/gemma/blob/0d6ae857591248422127ca14c027909546362e6a/gemma/modules.py#L80
+        """
+
+        UNIT = 1
```
ruomingp, line 798: Nit: use str values to be more readable.

```suggestion
        UNIT = "unit"
```

ruomingp (approved):

PR #914, opened 2025-01-09
Optional `positions` support in decoder and attention layers

changlan:
We have assumed that `positions` was only used for learnable positional embedding in the decoder and thus is not passed into the underlying transformer layers. Absolute positional encoding schemes fused in QKV projection layer (e.g. RoPE) are assumed to use `arange(seq_len)` as position ids of tokens.

This PR passes optional `positions` into `forward` of `BaseStackedTransformerLayer` and `BaseQKVLinear` for greater flexibility.


apghml (approved):
(But please wait to merge until the bump pr passes CI.)

ruomingp (approved):

PR #916, opened 2025-01-10
TRN2 Meshes and Configurations

apoorvtintin:
This PR adds meshes for TRN2/1 for Fuji models and transformer layer configuration favorable to Neuron.

Neuron supports stacked transformer and GroupedQKVLinear instead of FusedGroupedQKVLinear for Grouped Query Attention (GQA)

This is a newer version of the PR https://github.com/apple/axlearn/pull/885. This PR resolved all comments and requested changes mentioned in the linked PR.


apoorvtintin:
Added a ModelConfigModifier that overrides the class for a module. Allowing different model configurations based on Model size and platform.

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +147,67 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+            The modified trainer config.
+        """
+
+        for module_name, model_cfg in self._model_cfg_modifications.items():
+            # No modification if None
+            if not model_cfg:
+                continue
+            # Here we assume x.y.z format.
+            # One example would be model.decoder.transformer.layer.
+            target_modules = module_name.split(".")
```
kelvin-zou, line 188: Can you try to extract a common util function named something like
`def replace_module_recursive(target_modules:str, config_key: str, target_config)` and make it applied to both here and RematSpecModifier

axlearn/common/trainer_config_modifier_test.py
```diff
@@ -65,6 +67,26 @@ def test_remat_policy_override(self):
...
+                model_cfg_modifications={
+                    "model.decoder.transformer": RepeatedTransformerLayer.default_config(),
+                }
+            )
+            .instantiate()
+        )
+
+        cfg = cfg_modifier(cfg)
+        # The default StackedTransformerLayer should have changed to RepeatedTransformerLayer
+        self.assertRegex(str(cfg.model.decoder), ".*RepeatedTransformerLayer")
```
kelvin-zou, line 87: Can we do exact match here for the config, something like following: 
```suggestion
        self.assertTrue(cfg.model.decoder is RepeatedTransformerLayer.default_config())
```

axlearn/common/trainer_config_modifier_test.py
```diff
@@ -65,6 +67,26 @@ def test_remat_policy_override(self):
...
+            .set(
+                model_cfg_modifications={
+                    "model.decoder.transformer": RepeatedTransformerLayer.default_config(),
+                }
+            )
+            .instantiate()
+        )
+
+        cfg = cfg_modifier(cfg)
+        # The default StackedTransformerLayer should have changed to RepeatedTransformerLayer
```
kelvin-zou, line 86: Add a test to catch mismatch exception as well?

kelvin-zou (commented):
Thank you for making such change, overall looks good. A few nit comments.

axlearn/common/trainer_config_modifier.py
apoorvtintin: I extractred a helper function, let me know if this looks good

axlearn/common/trainer_config_modifier.py
```diff
@@ -17,7 +18,42 @@
...
 from axlearn.common.metrics import MetricAccumulator
 from axlearn.common.trainer import SpmdTrainer
-from axlearn.common.utils import HybridMeshShape, MeshShape
+from axlearn.common.utils import HybridMeshShape, MeshShape, PartitionSpec
+
+
+def find_target_module(
+    module_name: str, cfg: SpmdTrainer.Config
+) -> tuple[ConfigModifier.Config, ConfigModifier.Config, ConfigModifier.Config]:
+    """Recursively search for the target module matching module_name in provided cfg
```
ruomingp, line 27: End comments with .

```suggestion
    """Recursively search for the target module matching module_name in provided cfg.
```

axlearn/common/trainer_config_modifier.py
```diff
@@ -17,7 +18,42 @@
 from axlearn.common.gradient_accumulation import with_minibatch_steps
 from axlearn.common.metrics import MetricAccumulator
 from axlearn.common.trainer import SpmdTrainer
-from axlearn.common.utils import HybridMeshShape, MeshShape
+from axlearn.common.utils import HybridMeshShape, MeshShape, PartitionSpec
+
+
+def find_target_module(
+    module_name: str, cfg: SpmdTrainer.Config
+) -> tuple[ConfigModifier.Config, ConfigModifier.Config, ConfigModifier.Config]:
```
ruomingp, line 26: Do not return tuples. Return a struct or namedtuple:

https://github.com/apple/axlearn/blob/main/docs/ml_api_style.md#avoid-returning-a-tuple-of-values

axlearn/common/trainer_config_modifier.py
```diff
@@ -17,7 +18,42 @@
 from axlearn.common.gradient_accumulation import with_minibatch_steps
 from axlearn.common.metrics import MetricAccumulator
 from axlearn.common.trainer import SpmdTrainer
-from axlearn.common.utils import HybridMeshShape, MeshShape
+from axlearn.common.utils import HybridMeshShape, MeshShape, PartitionSpec
+
+
+def find_target_module(
```
ruomingp, line 24: Does this need to be a public method?

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +175,100 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+            ValueError: The target module is not found.
+
+        Returns:
+            The modified trainer config.
+        """
+
+        for module_name, model_cfg in self._model_cfg_modifications.items():
+            # No modification if None
+            if not model_cfg:
+                continue
```
ruomingp, line 213: Do we need to support this? Users can always omit the entry.

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +175,100 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+
+            found_module, parent_module, key_in_parent = find_target_module(module_name, cfg)
+
+            # Copy configurations from the config being replaced on a best effort basis
```
ruomingp, line 217: Wait, this behavior is not explained in the class comments. So we are not replacing but merging the configs? Maybe we should support a merge function instead?

ruomingp (changes requested):

axlearn/common/trainer_config_modifier.py
apoorvtintin: Yeah the goal is to change the config to a similar module. This means most of the configuration can be reused from before. Essentially replacing the module but merging the config. Let me extract out a merge function.

apoorvtintin: Abstracted out a merge function let me know if more changes are needed for this.

apoorvtintin:
@ruomingp Thank you for the review, I have addressed all your comments, please let me know if more changes are needed.

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +186,111 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+        """Merge configurations from the config being replaced on a best effort basis.
+        Args:
+            target_cfg: configuration that will replace found_module.
+            found_module: existing configuration whose class will be replaced
+                but it's confguration will be merged with target_cfg.
+
+        Returns:
+            The modified config.
+
+        """
```
ruomingp, line 218: Please follow Google style guide strictly in terms of linebreaks.

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +186,111 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+            ValueError: The target module is not found.
+
+        Returns:
+            The modified trainer config.
+        """
+
+        for module_name, model_cfg in self._model_cfg_modifications.items():
+            found_module = _find_target_module(module_name, cfg)
+
+            model_cfg = self._merge_configs(model_cfg, found_module)
```
ruomingp, line 242: `_merge_configs` actually mutates `model_cfg` IIUC. Is it safe to invoke `__call__` multiple times?

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +186,111 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+
+        model_cfg_modifications: Required[Dict[str, ConfigBase]] = REQUIRED
+
+    def __init__(self, cfg: Config):
+        super().__init__(cfg)
+        cfg = self.config
+        self._model_cfg_modifications = cfg.model_cfg_modifications
+
+    def _merge_configs(self, target_cfg: ConfigBase, found_module: ConfigBase) -> ConfigBase:
+        """Merge configurations from the config being replaced on a best effort basis.
```
ruomingp, line 209: This comment is vague about how merging actually works. Looks like the rule is:

* Use target_cfg.klass
* Use the field value from found_module if the field exists in both configs (and is not "klass")
* Otherwise keep the value from target_cfg

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +186,111 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+    @config_class
+    class Config(ConfigModifier.Config):
+        """Configure ModelConfigModifier.
+
+        Attributes:
+            model_cfg_modifications: A mapping from module path
+                (e.g. `model.decoder.transformer.layer`) to a Config.
+        """
+
+        model_cfg_modifications: Required[Dict[str, ConfigBase]] = REQUIRED
```
ruomingp, line 201: How about:

```suggestion
        model_cfg_modifications: Dict[str, Callable[[ConfigBase], ConfigBase]] = {}
```

?

That is, the values are config transformer functions.

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +186,111 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+
+        Raises:
+            ValueError: The target module is not found.
+
+        Returns:
+            The modified trainer config.
+        """
+
+        for module_name, model_cfg in self._model_cfg_modifications.items():
+            found_module = _find_target_module(module_name, cfg)
```
ruomingp, line 240: In utils.py we have `get_recursively` and `set_recursively` for `Nested[...]`. I wonder if it will be useful to add corresponding methods to ConfigBase. Then we can do something like:

```suggestion
        for cfg_path, cfg_modification in self._model_cfg_modifications.items():
            child_cfg = cfg.get_recursively(cfg_path)
            child_cfg = cfg_modification(child_cfg, path=cfg_path)
            cfg.set_recursively(cfg_path, value=child_cfg)
```

ruomingp (changes requested):

axlearn/common/trainer_config_modifier.py
apoorvtintin: Agree, it is possible that a modified config is dropped if a parent config is changed. I can sort the modification modules by string so that we always modify parents first. Let me know if you had another solution in mind.

axlearn/common/trainer_config_modifier.py
apoorvtintin: Added a more verbose comment

apoorvtintin:
Added a more flexible partition_spec modifier that can modify multiple partition_spec attributes in a single module config.

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +186,113 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+    @config_class
+    class Config(ConfigModifier.Config):
+        """Configure ModelConfigModifier.
+
+        Attributes:
+            model_cfg_modifications: A mapping from module path
+                (e.g. `model.decoder.transformer.layer`) to a Config.
+        """
+
+        model_cfg_modifications: Dict[str, Callable[[ConfigBase], ConfigBase]] = {}
```
kelvin-zou, line 201: There is a chain modifier already, should this be something simpler, saying taking two arguments, 
1. target_module
2. modified_config
If you rely on chain, you don't really need a dict here.

axlearn/common/trainer_config_modifier.py
```diff
@@ -17,7 +18,53 @@
...
+    """
+
+    # Here we assume x.y.z format.
+    # One example would be model.decoder.transformer.layer.
+    target_modules = module_name.split(".")
+    curr_module = cfg
+    key_in_parent = None
+    parent_module = None
+
+    for target_module in target_modules:
```
kelvin-zou, line 59: nit, maybe name `target_module` to `target_module_key` since it is more approriate?

kelvin-zou (commented):
Mostly lgtm, some minor comments.

axlearn/common/trainer_config_modifier.py
apoorvtintin: I felt the function of chain modifier is to chain different types of modifiers together. We have a tradeoff here I can certainly use the same ModelConfigModifier multiple times but that will make the config more verbose, while it will make the modifier simpler. I am open to both approaches.

apoorvtintin:
@ruomingp I addressed your comments, please let me know if anymore changes are needed.

axlearn/common/trainer_config_modifier.py
markblee: I wonder if an alternative (which aims to simplify the `ConfigBase` api) is to do something similar to Python's `sorted`; we allow `utils.get_resursively` to take a value fn:
```
# Default behavior is to use key lookup:
utils.get_recursively(..., value_fn=lambda k,v: v[k])

# Custom behavior can be attribute lookup:
utils.get_recursively(..., value_fn=lambda k,v: getattr(v,k))
```
A benefit is that other non-config instances can also leverage `get_recursively`.

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -151,6 +155,60 @@ def get_trainer_kwargs(
...
+        # So compile time does not grow with the number of layers.
+        ModelConfigModifier.default_config().set(
+            target_config="model.decoder.transformer",
+            modification=StackedTransformerLayer.default_config(),
+        )
+    ]
+    if version != Version.V1:
+        trn2_model_modifications.append(
+            ModelConfigModifier.default_config().set(
+                target_config="model.decoder.transformer.layer.self_attention.attention."
```
markblee, line 170: A downside of representing these deeply nested configs as string paths is that they are brittle, and can quickly become outdated.

Have we considered using `cfg.visit` to achieve some of these modifications (e.g., https://github.com/apple/axlearn/blob/1c22688c1dc480777c43a1b8c8dd866dccda70e0/axlearn/common/layers.py#L266)?

(A bit late to review, so apologies if this discussion has already taken place.)

axlearn/common/config.py
```diff
@@ -394,6 +394,66 @@ def set(self, **kwargs):
...
+
+        if len(key_path) == 1:
+            # Completed traversal
+            return self.TraverseResult(parent=self, key=key_path[0])
+
+        # Continue searching recursively
+        value = getattr(self, target_key)
+        return value.recursive_traverse(key_path[1:])
+
+    def get_recursively(self, key_path: Sequence[str]) -> Any:
```
ruomingp, line 429:
```suggestion
    def get_recursively(self, path: Sequence[str]) -> Any:
```

axlearn/common/config.py
```diff
@@ -394,6 +394,66 @@ def set(self, **kwargs):
...
+        """Recursively find the target key in the config and return its value.
+
+        Args:
+            key_path: A sequence of keys for indexing to get the target value.
```
ruomingp, line 433: Can path be empty? Maybe it can return `self` if path is empty?

axlearn/common/config.py
```diff
@@ -394,6 +394,66 @@ def set(self, **kwargs):
...
+            value at the key_path.
+        """
+        traverse_result = self.recursive_traverse(key_path)
+        return getattr(traverse_result.parent, traverse_result.key)
+
+    def set_recursively(self, key_path: Sequence[str], new_value: Any):
+        """Recursively find the target key in the config and set its value.
+
+        Args:
+            key_path: A sequence of keys for indexing to set the target value.
```
ruomingp, line 448: Can path be empty?

axlearn/common/config.py
```diff
@@ -394,6 +394,66 @@ def set(self, **kwargs):
...
+            new_value: New value to replace the target value.
+
+        Raises:
+            ValueError: A key in key_path is not found.
+        """
+        traverse_result = self.recursive_traverse(key_path)
```
ruomingp, line 454: Can we do something like:
```suggestion
        if not path:
            raise ValueError(...)
        parent = self.get_recursively(path[:-1])
```

axlearn/common/config.py
```diff
@@ -394,6 +394,66 @@ def set(self, **kwargs):
...
 
+    class TraverseResult(NamedTuple):
+        """Result of a recurisve traverse in a nested ConfigBase."""
+
+        # The parent that contains the reulting key.
+        parent: _ConfigBase
+        # The key string.
+        key: str
+
+    def recursive_traverse(self, key_path: Sequence[str]) -> tuple[Any, str]:
```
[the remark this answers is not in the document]
apoorvtintin: Calling value.recursive_traverse() results in a pylint error "Access protected method from non member function". Let me see if I can avoid that somehow.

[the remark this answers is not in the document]
ruomingp: I don't think we need this method. Please see my other comments.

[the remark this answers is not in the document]
apoorvtintin: I removed this method

axlearn/common/config.py
apoorvtintin: Added a check to see if path is empty, throws value error now if path is empty

axlearn/common/config.py
apoorvtintin: I removed the `recursive_traverse` method and added a value error if path is empty. I am still using path[1:] instead path[:-1] since we are traversing parents first.

axlearn/common/trainer_config_modifier.py
```diff
@@ -146,6 +137,110 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
         return cfg
 
 
+class ModelConfigModifier(ConfigModifier):
```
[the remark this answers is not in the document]
apoorvtintin: The class makes some assumptions around the `klass` attribute of configs which are only present in modules, which is also what makes it non generic.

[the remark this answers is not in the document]
apoorvtintin: @markblee I would love to get some more guidance on a name that fits this class better, do you have any suggestions?

axlearn/common/config.py
```diff
@@ -394,6 +394,66 @@ def set(self, **kwargs):
...
+        """Recursively traverse the config to find the target key.
```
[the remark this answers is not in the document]
apoorvtintin: Both loops and recursion are fair choices I was following the pattern of `utils.get_recusively()` based on previous guidance.

apoorvtintin:
Thank you for the reviews @markblee @ruomingp @kelvin-zou!
@ruomingp I addressed most of your comments, please let me know if more changes are needed. @markblee We can discuss the usage of `cfg.visit` instead of using full module paths further, but I would like to address this in a follow up PR if possible.

axlearn/common/config.py
```diff
@@ -394,6 +394,54 @@ def set(self, **kwargs):
...
+
+        Raises:
+            ValueError: A key in path is not found or path is empty
+
+        Returns:
+            value at the path.
+        """
+        if not path:
+            raise ValueError("Path is empty.")
+
```
ruomingp, line 411:
```suggestion
        if not path:
            return self

```

axlearn/common/config.py
```diff
@@ -394,6 +394,54 @@ def set(self, **kwargs):
...
+        if path[0] not in self:
+            raise ValueError(f"{path[0]} is not found in {self}.")
+
+        child = getattr(self, path[0])
+
+        if len(path) == 1:
+            return child
+
+        return child.get_recursively(path[1:])
```
ruomingp, line 420: Nit: Can we avoid recursion with a loop?

axlearn/common/config.py
```diff
@@ -394,6 +394,54 @@ def set(self, **kwargs):
...
+        if not path:
+            raise ValueError("Path is empty.")
+
+        if path[0] not in self:
+            raise ValueError(f"{path[0]} is not found in {self}.")
+
+        child = getattr(self, path[0])
+
+        if len(path) == 1:
+            return setattr(self, path[0], value)
```
ruomingp, line 441:
```suggestion
        if len(path) == 1:
            return setattr(self, path[0], value)
        
        child = getattr(self, path[0])
``````

File: axlearn/common/config.py [hunk 27]
```diff
@@ -394,6 +394,54 @@ def set(self, **kwargs):
...
+
+        return child.set_recursively(path[1:], value=value)
```
ruomingp, line 443: Avoid recursion?

axlearn/common/config.py
```diff
@@ -394,6 +394,54 @@ def set(self, **kwargs):
...
+            new_value: New value to replace the target value.
+
+        Raises:
+            ValueError: A key in path is not found or path is empty
+        """
+        if not path:
+            raise ValueError("Path is empty.")
+
+        if path[0] not in self:
+            raise ValueError(f"{path[0]} is not found in {self}.")
```
ruomingp, line 436: Do we need this check? getattr and setattr will raise anyways.

axlearn/common/trainer_config_modifier.py
ruomingp: Hi @markblee , maybe we can do this in a follow-up PR?

axlearn/common/trainer_config_modifier.py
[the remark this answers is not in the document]
markblee: Hm, but all configs deriving from `Configurable` have `klass`. Did you mean `ModuleConfigModifier` perhaps, as it replaces arbitrary modules by path? (model -> module?)

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -434,7 +552,7 @@ def get_trainer_kwargs(
             ),
             learner_kwargs=dict(peak_lr=1.5e-4, weight_decay=0.1),
             max_sequence_length=max_sequence_length,
-            train_batch_size=train_batch_size,
+            train_batch_size=8,
```
markblee, line 555: Is this change intended?

markblee (commented):
Hi @apoorvtintin, please feel free leave `TODO(markblee):` for my comments and go ahead once @ruomingp approves. I'll be happy to send a follow-up PR to avoid blocking this one.

axlearn/common/trainer_config_modifier.py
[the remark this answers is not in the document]
apoorvtintin: Yeah I agree `ModuleConfigModifier` is better, thanks!

apoorvtintin:
Addressed all comments. @markblee I added TODOs for you to revisit cfg.visit later, please let me know if I should add more.

axlearn/common/config.py
```diff
@@ -394,6 +394,58 @@ def set(self, **kwargs):
...
+
+        while path and index < len(path):
+            key = path[index]
+
+            # TODO(markblee): maybe use cfg.visit instead of getattr
+            current = getattr(current, key)
+            index += 1
+
+            if index == len(path):
+                return current
```
ruomingp, line 421:
```suggestion
```

axlearn/common/config.py
```diff
@@ -394,6 +394,58 @@ def set(self, **kwargs):
...
+            raise ValueError("Path is empty.")
+
+        current = self
+        for i, key in enumerate(path):
+            if i == len(path) - 1:
+                setattr(current, key, value)
+                return
+            else:
+                # TODO(markblee): maybe use cfg.visit instead of getattr
+                current = getattr(current, key)
```
ruomingp, line 447:
```suggestion
        parent = self.get_recursively(path[:-1])
        setattr(current, path[-1], value)
```

axlearn/common/config.py
apoorvtintin: Thanks for the review @ruomingp. Can I get some more details here?

axlearn/common/config.py
apoorvtintin: I moved to loop based on the previous comments, just to clarify we don't want loops for set_recursively but only for get_recursively?

apoorvtintin:
Thanks for the latest comments @ruomingp. I need some clarification on your latest comments. This PR is almost done, let's push this to the finish line! :)

axlearn/common/config.py
apoorvtintin: Got this, will make this change

axlearn/common/config.py
apoorvtintin: Changed to `break`, let me know if you had something else in mind

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -151,6 +155,72 @@ def get_trainer_kwargs(
 
     rope_theta = ROPE_THETA[version]
 
+    # TRN2 specific model config modifications
```
kelvin-zou, line 158: Can we move all the modifications in a helper function? 
saying

def _generate_trainium2_custom_configs():
    ...
    return trn2_model_modifications, trn2_partition_spec_modifications

axlearn/experiments/text/gpt/fuji.py
```diff
@@ -151,6 +155,72 @@ def get_trainer_kwargs(
...
+                        "model",
+                        ("expert", "fsdp", "seq"),
+                    ),
+                    "input_partition_spec": ("fsdp", None),
+                    "output_partition_spec": ("fsdp", None, None),
+                    "embedding_partition_spec": ("model", None),
+                },
+                # Sequence parallel shardings for norms.
+                "model.decoder.transformer.layer.self_attention.norm": {
+                    "input_partition_spec": ("fsdp", "model", None),
```
hanzhi713, line 192: Why is there no "data"? Is it because trn2 don't support data and fsdp at the same time?

axlearn/experiments/text/gpt/fuji.py
apoorvtintin: Addressed it, lmk if it looks good

kelvin-zou (commented):
Approve overall, please address @hanzhi713 's comment.

apoorvtintin:
Updated the PR to address failing tests, can we re-trigger the CI please? Thank you

axlearn/experiments/text/gpt/fuji.py
apoorvtintin: Yes "data" over "fsdp" is supported. I'll make this change

apoorvtintin:
@kelvin-zou @hanzhi713 Thank you both for the review, I addressed all the comments. Let's merge this if it looks good

kelvin-zou (approved):
Thank you!

hanzhi713 (approved):

apoorvtintin:
Hello @ruomingp, can I please get an approval if this PR looks good? Looks like that is needed for the PR to merge.

kelvin-zou:
maybe @markblee  can have a second eye for the final round?

axlearn/common/config.py
```diff
@@ -394,6 +394,49 @@ def set(self, **kwargs):
...
+
+        while path and index < len(path):
+            key = path[index]
+
+            # TODO(markblee): maybe use cfg.visit instead of getattr
+            current = getattr(current, key)
+            index += 1
+
+            if index == len(path):
+                break
```
markblee, line 420: Hm, did I miss something or can we just do:
```suggestion
        for part in path:
            current = getattr(current, part)
```

(In any case, I can make some changes in a follow-up.)

markblee (approved):
Approve to unblock.

markblee:
(Looks like we still need @ruomingp 's approval to unblock the 'requested changes'.)

ruomingp (approved):

axlearn/common/config.py
apoorvtintin: Made this change, thanks @markblee

markblee (approved):

apoorvtintin:
Hello @ruomingp @markblee looks like this PR green for merge, let me know if any more changes are needed.

apoorvtintin:
Rebased the PR to include latest Axlearn changes.

kelvin-zou:
Verified internally on upgrade PR, will merge it.

PR #920, opened 2025-01-13
Introduce `BaseAttentionBias.has_value()`.

ds-hwang:
`BaseAttentionBias.value()` is also used to check whether a bias value exists. However, calling `value()` creates the actual bias on the CPU, which is quadratic O(T^2) — especially when debugging long sequence lengths (e.g., 32k). Hence, we introduce a lightweight `has_value()` method.

Normally, unused tensors are pruned from the graph during XLA compilation, and using `BaseAttentionBias.value()` to check for bias presence relies on XLA pruning. But `BaseAttentionBias.value()` is still expensive on the CPU for unittests and debugging, so `has_value()` saves Python runtime.

The new `has_value()` method checks whether `value()` actually exists by calling `jax.eval_shape`. Since `jax.eval_shape` invokes `value()` through a tracer, it doesn’t materialize the actual value (proposed by John Peebles).

### DIFF
--- a/axlearn/common/attention_bias.py
+++ b/axlearn/common/attention_bias.py
@@ -60,6 +60,28 @@ class BaseAttentionBias:
     dtype: Optional[jnp.dtype] = struct.field(kw_only=True, default=None, pytree_node=False)
 
     @final
+    def eval_shape(self) -> tuple[int, int, int, int]:
+        """Return the shape of the bias tensor.
+
+        Note: this doesn't materialize the value. jax.eval_shape calls value(), but it only does so
+        using tracers.
+
+        Returns
+            shape: [batch or 1, num_heads or 1, target_len, source_len].
+
+        Raises:
+            ValueError: If the bias has no value.
+        """
+        if not self.has_value():
+            raise ValueError("AttentionBias has no value.")
+        return jax.eval_shape(self.value).shape
+
+    @final
+    def has_value(self) -> bool:
+        """Return whether to the bias has a value."""
+        return jax.eval_shape(self.value) is not None
+
+    @final
     def value(self) -> Optional[Tensor]:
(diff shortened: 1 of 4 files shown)


ds-hwang:
@ruomingp, @markblee Could you take a look? From 972

axlearn/common/attention_bias.py
```diff
@@ -59,6 +59,28 @@ class BaseAttentionBias:
...
+        Note: this doesn't materialize the value. jax.eval_shape calls value(), but it only does so
+        using tracers.
+
+        Returns
+            shape: [batch or 1, num_heads or 1, target_len, source_len].
+
+        Raises:
+            ValueError: If the bias has no value.
+        """
+        if not self.has_value():
```
markblee, line 75: I suppose multiple calls to jax.eval_shape are automatically cached?

markblee (approved):

axlearn/common/attention_bias.py
ds-hwang: I couldn't find caching mechanism in jax impl. Would you mind asking how it's related to this PR? Can I merge it? 
```
@api_boundary
def eval_shape(fun: Callable, *args, **kwargs):
  """Compute the shape/dtype of ``fun`` without any FLOPs.

  This utility function is useful for performing shape inference. Its
  input/output behavior is defined by::

```

markblee: Oh, in `eval_shape`, we make two calls to `jax.eval_shape(self.value)` (once in `has_value` and once in return) -- so ideally we only trace once. I checked and it seems to be the case, so please feel free to ignore.

ds-hwang: Ah, I see. Good point. Thank you for check!

PR #922, opened 2025-01-14
Enabled running Pallas Flash Attention on CPU.

ds-hwang:
Enabled running Pallas Flash Attention on CPU.

Pallas supports CPU simulation (`interpret=True`), so we can use the same
TPU Pallas kernel on CPU — making code debugging easier.

This change lets the following unittests run on CPU as if they were on TPU,
enabling easier testing and debugging:
- `axlearn/common/flash_attention/tpu_attention_test.py`

Similarly, `gpu_attention_test.py` can also be run on CPU as if they were on GPU.
- `axlearn/common/flash_attention/gpu_attention_test.py`

Now CI covers those tests on CPU as well.
In M3 Max MacBook Pro, test coverages and processing time are as follows,
* axlearn/common/flash_attention/gpu_attention_test.py: 3024 passed, 1345 skipped in 200.38s (0:03:20)
* axlearn/common/flash_attention/tpu_attention_test.py: 18 passed, 435 skipped in 34.82s

### DIFF
--- a/axlearn/common/flash_attention/gpu_attention_test.py
+++ b/axlearn/common/flash_attention/gpu_attention_test.py
@@ -28,7 +28,7 @@ from axlearn.common.flash_attention.gpu_decoding import NEG_INF, flash_decoding
 from axlearn.common.flash_attention.utils import _repeat_kv_heads, mha_reference
 from axlearn.common.test_utils import TestCase
 
-if jax.default_backend() != "gpu":
+if jax.default_backend() not in ("gpu", "cpu"):
     pytest.skip(reason="Incompatible hardware", allow_module_level=True)
 
 
@@ -69,6 +69,8 @@ def test_triton_fwd_only_against_ref(
         kv_seq_len = seq_len
     if kv_seq_len != seq_len and use_segment_ids:
         pytest.skip()
+    if jax.default_backend() == "cpu" and kv_seq_len > 128:
+        pytest.skip(reason="CI got OOM.")
     k1, k2, k3, k4, k5 = jax.random.split(jax.random.PRNGKey(0), 5)
     q = jax.random.normal(k1, (batch_size, seq_len, num_heads, per_head_dim), dtype=input_dtype)
     k = jax.random.normal(k2, (batch_size, kv_seq_len, num_heads, per_head_dim), dtype=input_dtype)
@@ -101,6 +103,7 @@ def test_triton_fwd_only_against_ref(
         causal=causal,
         softmax_scale=softmax_scale,
         dropout_rate=dropout_rate,
+        interpret=(jax.default_backend() == "cpu"),
     )
     o_ref = mha_reference(
         q,
@@ -152,6 +155,8 @@ class FlashDecodingTest(TestCase):
         kv_head_factor: int,
         window_len: int,
     ):
+        if jax.default_backend() == "cpu" and seq_len >= 512:
+            pytest.skip(reason="Too slow on CPU.")
         self.assertEqual(num_heads % kv_head_factor, 0)
         assert num_heads % kv_head_factor == 0
         k1, k2, k3, k4 = jax.random.split(jax.random.PRNGKey(42), 4)
@@ -180,7 +185,14 @@ class FlashDecodingTest(TestCase):
         if window_len > 0:
             mask_fn = sliding_window_causal_mask(window_len)
         o = flash_decoding(
-            q, k, v, bias=bias, softmax_scale=softmax_scale, kv_seq_len=seq_len, mask_fn=mask_fn
+            q,
+            k,
+            v,
+            bias=bias,
+            softmax_scale=softmax_scale,
+            kv_seq_len=seq_len,
+            mask_fn=mask_fn,
+            interpret=(jax.default_backend() == "cpu"),
         )
         if bias is not None:
             bias = bias[:, :, :, :seq_len]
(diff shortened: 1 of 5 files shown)


ds-hwang:
@ruomingp Could you take a look? From 975

axlearn/common/flash_attention/gpu_attention_test.py
```diff
@@ -152,6 +153,8 @@ def test_decode_against_ref(
         kv_head_factor: int,
         window_len: int,
     ):
+        if jax.default_backend() != "gpu" and seq_len > 1024:
```
ruomingp, line 156: Nit: can we check it against "cpu" directly instead of `!= "gpu"`?

axlearn/common/flash_attention/gpu_attention_test.py
```diff
@@ -346,6 +357,9 @@ def test_cudnn_against_triton_ref(
     causal: bool,
     dtype: jnp.dtype,
 ):
+    if jax.default_backend() == "cpu":
```
ruomingp, line 360: Likewise, let's avoid assuming that the backend is either gpu or cpu in multiple places.

```suggestion
    if jax.default_backend() != "gpu":
```

axlearn/common/flash_attention/gpu_attention_test.py
```diff
@@ -414,6 +444,8 @@ def test_cudnn_dropout_against_xla_dropout(
     by setting V to the identity matrix. However, this only works when seq_len == per_head_dim,
     i.e. when the shape of output is the same as the shape of the dropout mask.
     """
+    if jax.default_backend() == "cpu":
+        pytest.skip(reason="cudnn function needs GPU.")
```
ruomingp, line 448: And here and elsewhere.

axlearn/common/flash_attention/tpu_attention_test.py
```diff
@@ -97,12 +92,11 @@ def test_to_splash_mask(self, mask, expected):
 
     @parameterized.product(
         batch_size=[4],
-        seq_len=[1024, 32768],
+        seq_len=[1024],
```
ruomingp, line 95: Since the sliding window size is 1024, it will be useful to keep a test case for seq_len > 1024. We can enable the test only on TPU if it's too slow on CPU. We can also use a seq_len such as 2048 for cpu if it's fast enough.

axlearn/common/flash_attention/utils.py
```diff
@@ -268,20 +269,18 @@ def get_segment_ids(segment_ids: SegmentIdAttentionBias) -> Optional[Tensor]:
...
                 bias=explicit_bias.value(),
                 segment_ids=get_segment_ids(segment_ids),
-                # The `from_sequence()` function guarantees that if there is only one
-                # mask, it is returned without modification.
-                # This allows the `causal` path in `_legacy_tpu_flash_attention()` to work.
-                mask=mask if not isinstance(mask, ZeroAttentionBias) else None,
+                mask=mask,
                 softmax_scale=softmax_scale,
                 block_size=block_size,
+                interpret=(backend == "cpu"),
```
ruomingp, line 275: Given how often we do this across locations, I wonder if we can do the following:

- Make `interpret` default to None (instead of False);
- If it's None, assume interpret=True if the backend is "cpu";

WDYT?

ruomingp (commented):
A few thoughts missed in earlier reviews...

axlearn/common/flash_attention/tpu_attention_test.py
ds-hwang: Done. I changed it back to match the first PR.
image

axlearn/common/flash_attention/utils.py
ds-hwang: Thank you for your suggestion. `interpret=True` applies only to the Pallas kernel. Therefore, having an `interpret` variable in the flash layer is not aligned with the appropriate level of abstraction—neither the JAX fallback nor the cudnn code paths needs this variable.

Additionally, this line was added so contributors can easily debug the Pallas kernel on the CPU. For instance, changing the `if` statement to:
```
elif backend in ("cpu", "tpu"):
```
would allow debugging in `layer_test.py`.

axlearn/common/flash_attention/gpu_attention_test.py
ds-hwang: Yes, done.

axlearn/common/flash_attention/gpu_attention_test.py
ds-hwang: I'll leave this code as-is as you asked
> Nit: can we check it against "cpu" directly instead of != "gpu"?

In addition, at the begin of file, it allows only "gpu" and "cpu". So `== "cpu"` is `!= "gpu"` in this code.
```
if jax.default_backend() not in ("gpu", "cpu"):
```

axlearn/common/flash_attention/gpu_attention_test.py
ds-hwang: As mentioned above, keep using `jax.default_backend() == "cpu":`

ds-hwang (commented):
Thank you for review. I responded all comments. Could you check it again?

axlearn/common/flash_attention/gpu_attention_test.py
ruomingp: > In addition, at the begin of file, it allows only "gpu" and "cpu". So == "cpu" is != "gpu" in this code.

I know you are making this assumption, but such dependency is fragile---what if we extend the supported backends in the future?

In this case, requiring the backend to be "gpu" is both more robust and readable. What's the downside?

ruomingp (approved):
It seems possible to support `interpret=None` to simplify the code, with the following behavior:

- `interpret=True/False`: enable/disable interpret;
- `interpret=None`: let the implementation choose whether to interpret, depending on whether pallas is used and running on cpu vs. accelerator;

But I don't want to block this PR, as we can simplify it later.

ds-hwang:
Thank you for review!

ds-hwang:
@ruomingp Could you reapprove it? I rebased the commit because it’s been pending for a few days and couldn’t be merged due to an unrelated test failure.

ruomingp (approved):

PR #923, opened 2025-01-14
DiT now supports sequence conditions.

ds-hwang:
When using seq2seq models with DiT, the condition may have the same sequence length as the input.
For example:
- Input shape: `[batch, seq_len, dim]`
- Condition shape: `[batch, seq_len, cond_dim]`

AdaptiveLayerNormModulation now supports conditions in both `[batch, cond_dim]` and `[batch, seq_len, cond_dim]` formats. It outputs conditions in the shape `[batch, 1|seq_len, cond_dim]`, depending on whether `seq_len` is present.

Accordingly, DiT has been updated to handle rank-3 conditions. The codebase has also become simpler. Previously, `jnp.expand_dims` was scattered across many places, but now `AdaptiveLayerNormModulation` adjusts the rank of the condition to match the input and returns it accordingly.

Speech detokenizer will use this DiT.

### DIFF
--- a/axlearn/common/dit.py
+++ b/axlearn/common/dit.py
@@ -13,6 +13,8 @@ Ref: https://github.com/facebookresearch/DiT
 
 from typing import Optional, Union
 
+import chex
+import einops
 import jax
 import jax.numpy as jnp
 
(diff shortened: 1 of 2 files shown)


ds-hwang:
@ruomingp Could you take a look? From 979

axlearn/common/dit.py
```diff
@@ -211,15 +227,18 @@ def forward(self, input: Tensor) -> Tensor:
...
 
         Returns:
             A list of tensors with length num_outputs.
-                Each tensor has shape [batch_size, ..., dim].
+                Each tensor has shape [batch_size, 1|num_length, dim].
         """
         cfg = self.config
         x = get_activation_fn(cfg.activation)(input)
         output = self.linear(x)
+        assert output.ndim in (2, 3)
```
ruomingp, line 239: Raise a ValueError instead of `assert` (which should only be used to enforce internal logic errors and cannot be triggered by user error).

ruomingp (approved):

axlearn/common/dit.py
ds-hwang: As I clicked automerge, this your comment was not handled. I submitted #981 to handle it.

PR #924, opened 2025-01-14
Fix tf iter unit test

hanzhi713:
This tf iter unit test didn't wait for async future to finish and can cause some flakiness in CI. Originally added in https://github.com/apple/axlearn/pull/830.


markblee (approved):
Thanks for catching my mistake! 🙂

PR #926, opened 2025-01-15
Allow external positions to be inputed in RoPE embedding layer

Firenze11:
Use case: In RoPE embedding, position embeddings are applied to Q, K, V values after `i_proj`. Unlike the implementation of current `RoFormerQKVLinear`, in MaskedDiT we need to customize positions to indicate masked versus non-masked positions in the position embedding. When we convert this masked roformer attention module to flash attention, we need its signature to be supported by `MultiheadAttention`.


axlearn/common/attention.py
```diff
@@ -1216,18 +1216,37 @@ class Config(BaseLayer.Config):
         dim: Required[int] = REQUIRED  # The dimensionality of the positional embedding.
         theta: float = 10000.0  # The scale of base frequency.
 
-    def forward(self, positions: Tensor) -> Tensor:
+    def default_query_positions(self, max_seq_len: int) -> Tensor:
```
markblee, line 1219:
```suggestion
    def _default_query_positions(self, max_seq_len: int) -> Tensor:
```
Users should pass `max_seq_len` rather than calling this method publicly.

axlearn/common/attention.py
```diff
@@ -1216,18 +1216,37 @@ class Config(BaseLayer.Config):
...
+    def forward(
+        self, positions: Optional[Tensor] = None, max_seq_len: Optional[int] = None
+    ) -> Tensor:
         """
         TODO(bwzhang): 1. verify the performance under float32.
 
         Args:
             positions: A tensor representing the token position IDs.
                 The shape is [batch_size, seq_len].
+            max_seq_len: Max length of sequence, required if positions is not provided
```
markblee, line 1234:
```suggestion
            max_seq_len: Max length of sequence, required if positions is not provided.
```

axlearn/common/attention.py
Firenze11: There might be situations that we want to access the default query positions from the outside, such as getting the default positions and computing the positions based on it before passing it to `forward`. Therefore we want to keep this public.

ruomingp: Do we expect subclasses to override this method?

If not, it will be more readable for callers to call `jnp.arange` directly given the implementation is only one line.

axlearn/common/attention.py
```diff
@@ -1216,18 +1216,37 @@ class Config(BaseLayer.Config):
...
 
-    def forward(self, positions: Tensor) -> Tensor:
+    def default_query_positions(self, max_seq_len: int) -> Tensor:
+        """Compute default `positions` value to be inputed into forward when `positions` is
+        not provided to the corresponding QKVLinear class such as `RoFormerQKVLinear`
+        """
+        return jnp.arange(max_seq_len)[None]  # [batch_size=1, max_seq_len].
+
+    def forward(
+        self, positions: Optional[Tensor] = None, max_seq_len: Optional[int] = None
```
ruomingp, line 1226: Do we need to support both args? It seems simpler and more readable to take `positions` only and let the caller call `jnp.arange` if necessary.

axlearn/common/attention.py
Firenze11: Yes we do expect subclasses to override this method. One example is that we might want to specify the rotation start index and rotation end index for this embedding class. And the default embedding positions will be different then.

axlearn/common/attention.py
kelvin-zou: +1, it seems cleaner to not touch this function, instead extend more logic in RoFormerQKVLinear class directly where you can set position directly?

kelvin-zou: Thanks @Firenze11 for reminding me that this design has been discussed internally already. Keeping it here sgtm.

kelvin-zou (approved):
Thanks

axlearn/common/attention.py
ruomingp, line 1226:
```suggestion
        self, *, positions: Optional[Tensor] = None, max_seq_len: Optional[int] = None
```

https://github.com/apple/axlearn/blob/main/docs/ml_api_style.md#avoid-multiple-positional-arguments

axlearn/common/attention.py
```diff
@@ -1216,18 +1216,37 @@ class Config(BaseLayer.Config):
...
 
         Returns:
             Rotary Positional Embedding. Shape is [seq_len, dim].
+
+        Raises:
+            ValueError: If positions is None and max_seq_len is None.
         """
         cfg = self.config
+        if positions is None:
+            if max_seq_len is None:
```
ruomingp, line 1244: what happens if both `positions` and `max_seq_len` are provided? should we check that they are consistent?

axlearn/common/attention.py
ruomingp: Thanks for the explanation. Given this, it's reasonable to consolidate the position computation logic in this class.

axlearn/common/attention.py
Firenze11: In that case `positions` will take precedence and we ignore `max_seq_len`. We won't need `max_seq_len` if the client provides explicit positions. Will add that to docstring.

apghml: Maybe we can error if they  are both provided?

Firenze11: Sure can do this. @ruomingp any opinion on this?

ruomingp: It seems better to throw an error if they are both provided or when they are inconsistent. WDYT?

Firenze11: Done. @ruomingp Please take a look?

ruomingp (approved):

PR #928, opened 2025-01-16
Add prefill hidden states as module outputs.

markblee:
Some use cases involve reusing hidden states from the decoding prefix. cc @kevinye0108


kevinye0108:
LGTM. Thanks @markblee !

ruomingp (approved):

PR #930, opened 2025-01-16
AdaptiveLayerNormModulation raises ValueError, instead of assert.

ds-hwang:
Wrong inputs should be user error, not internal logic errors.


ds-hwang:
Hi, could you take a look? From 981

PR #931, opened 2025-01-17
Add ReadOptions args to _make_autoregressive_inputs

RsEnts:


axlearn/common/input_grain_lm.py
```diff
@@ -30,6 +31,8 @@ def _make_autoregressive_inputs(
     max_len: int,
     input_key: str = "target_labels",
     split_fn: Optional[ConfigOr[_SplitFn]] = None,
+    num_threads: int = 1,
```
markblee, line 34: This is a private method, how do users configure it? Should we pass through `text_to_lm_training_input`?

Also, would it be more general to `read_options` directly in case it eventually supports some form of autotune?

RsEnts: Thanks for the view. I just changed it to read_options as args, and also add it to text_to_lm_training_input.

zxybazh (approved):
LGTM, thanks!

markblee (approved):

PR #932, opened 2025-01-17
Enable GCP Workload Monitoring

Perseus14:
Integrate Workload Observability to AXLearn Framework

- Users can now send workload metrics (heartbeat and performance (step_time)) to GCP and CEs will monitor them for issues with workload
- Can simple modify config file to start sending the metrics, do not require any additional code change.

### DIFF
--- a/axlearn/common/input_grain_lm.py
+++ b/axlearn/common/input_grain_lm.py
@@ -119,9 +119,7 @@ class _StreamingPackingDatasetIterator(grain.DatasetIterator):
             # If there are still leftover tokens when we have already reached the window limit, we
             # should decide whether to keep this sequence.
             if self._current_token_count > 0 and self._reach_window_limit():
-                next_element = self._pop_element()
-                assert self._current_token_count == 0 and not self._current_examples_list
-                return next_element
+                return self._pop_element()
 
             # Keeps filling up the sequence until reaching the limit.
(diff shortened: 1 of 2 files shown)


ruomingp (commented):
Will defer to @Ethanlm and @markblee for approval.

axlearn/common/trainer.py
```diff
@@ -219,6 +221,26 @@ def __init__(
         super().__init__(cfg, parent=parent)
         cfg = self.config
 
+        self.gcp_workload_monitor = None
```
markblee, line 224: The trainer shouldn't have any dependencies on GCP specific logic. BTW, I wonder whether the monitoring events can be inferred from the events recorded by the `measurement` utils -- what are the main differences?

Perseus14: Got it! I will decouple the GCP logic from `Trainer`. I am thinking to do the following

- Create a `GCPRecorder` class that inherits from `Recorder`. 
- Register the `GCPRecorder` with the decorator `@register_recorder`.
- Finally, in `Trainer` if GCP monitoring is enabled, I will initialize `measurement` with the `recorder_type` set to `gcp:gcp` to use GCPRecorder.

Hopefully this will enable GCP monitoring in a cleaner way!

Let me know if it make sense.

PR #933, opened 2025-01-18
Supports flexible input partition specs.

markblee:


ruomingp (approved):
Did you validate that training still works?

markblee:
> Did you validate that training still works?

Added some links in the internal description. This should be a no-op to internal runs but I fixed and added a test for fuji.

axlearn/common/input_base.py
```diff
@@ -2,23 +2,121 @@
...
+        path: An optional path or path regex. None means match everything.
+        rank: An optional rank (ndim). None means match everything.
+    """
+
+    path: Optional[Union[str, re.Pattern]]
+    rank: Optional[int]
+
+
+def partition_by_path_rank(
+    path_rank_to_partition: dict[PathAndRank, PartitionSpec],
```
ruomingp, line 49: Nit: maybe we don't really need a dict:

```suggestion
    path_rank_to_partition: Sequence[IntputFieldPartitionSpec],
```

it simplifies the data structure and `IntputFieldPartitionSpec` is more extensible than `PartitionSpec`.

ruomingp (approved):

axlearn/common/input_base.py
markblee: Thanks -- to clarify, did you mean to specify `InputFieldPartitionSpec` as a `NamedTuple` of `(path, rank, PartitionSpec)`, or something else?

(To unblock my next PR, I am happy to address this in a follow-up, if that's OK.)

ruomingp: > Thanks -- to clarify, did you mean to specify `InputFieldPartitionSpec` as a `NamedTuple` of `(path, rank, PartitionSpec)`

Yes, a namedtuple, a dataclass, or a chex.struct class.

>, or something else?
> 
> (To unblock my next PR, I am happy to address this in a follow-up, if that's OK.)

Sure.

PR #934, opened 2025-01-19
Makes causal lm metrics configurable.

markblee:


ruomingp (changes requested):
Please see my comment on PredictModel in the internal PR.

markblee:
> Please see my comment on PredictModel in the internal PR.

Thanks. Removed as it's not critical to the PR.

ruomingp (approved):

axlearn/common/causal_lm.py
```diff
@@ -32,16 +32,242 @@
...
+        target_num_bytes: Optional[Tensor] = input_batch.get("target_num_bytes")
+        logits = predict_outputs["logits"]
+
+        live_targets = target_labels >= 0
+        num_targets = live_targets.sum()
+        accuracy = (
+            jnp.equal(jnp.argmax(logits, axis=-1), target_labels) * live_targets
+        ).sum() / jnp.maximum(1, num_targets)
+
+        loss, loss_dict = cross_entropy(
```
apivovarov, line 103: Hi Mark, do we need a separate accuracy calculation here (above cross_entropy) if the loss_dict from cross_entropy already contains an accuracy entry? @markblee

PR #940, opened 2025-01-22
Add GKE A3 Ultra support

samos123:
* Adds support for easily adding more GKE GPU accelerators by using a base class
* Adds Fuji v2 70B benchmark results


Dockerfile
```diff
@@ -101,14 +93,63 @@ COPY . .
 # GPU container spec.                                                          #
 ################################################################################
 
-FROM base AS gpu
+# This causes INTERNAL: No valid engine configs for Matmul error
```
samos123, line 96: @markblee @kelvin-zou are you fine with moving to nvidia/cuda:12.6.3-cudnn-devel-ubuntu22.04 as the base image. I couldn't get A3 Ultra working on the original python base image.

Dockerfile
```diff
@@ -101,14 +93,63 @@ COPY . .
...
+# End copy original base
+
+
+RUN apt update -y && apt-get install -y google-perftools glibc-tools
 
 # TODO(markblee): Support extras.
 ENV PIP_FIND_LINKS=https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
 RUN pip install .[core,gpu]
 COPY . .
+RUN pip install -U "jax[gpu]==0.4.38" "jax==0.4.38" "jaxlib==0.4.38" \
```
samos123, line 149: TODO: this should be removed. Maybe we should wait until axlearn main upgrades to 0.4.38.

Dockerfile
```diff
@@ -97,13 +97,52 @@ COPY . .
 # GPU container spec.                                                          #
 ################################################################################
 
-FROM base AS gpu
+# Using `FROM base as GPU` causes INTERNAL: No valid engine configs for Matmul error.
+# So we're using the NVIDIA provided cuda image instead which works.
+FROM nvidia/cuda:12.6.3-cudnn-devel-ubuntu22.04 as gpu
+
+# Copy from original base
```
ruomingp, line 104: Does it mean that we will need to keep the following commands consistent with those in BASE? Does Dockerfile support functions?

Dockerfile
```diff
@@ -97,13 +97,52 @@ COPY . .
 # GPU container spec.                                                          #
 ################################################################################
 
-FROM base AS gpu
+# Using `FROM base as GPU` causes INTERNAL: No valid engine configs for Matmul error.
+# So we're using the NVIDIA provided cuda image instead which works.
+FROM nvidia/cuda:12.6.3-cudnn-devel-ubuntu22.04 as gpu
```
ruomingp, line 102: Let's make sure this change doesn't break our GPU training on AWS.

ruomingp (commented):
Thanks!

Dockerfile
samos123: I did some research and sadly the answer is no. We could try to create a bash script that gets used in both instead of copy pasting code. I do agree that we should figure out a way to reuse the same setup steps.

Dockerfile
samos123: I think it's fine but defer to @kelvin-zou

samos123:
I still have access to internal capacity so wanted to get some reviews and make changes before I lose the ability to verify any changes.

Dockerfile
```diff
@@ -97,13 +97,52 @@ COPY . .
...
+RUN pip install --upgrade pip
+
 
 # TODO(markblee): Support extras.
 ENV PIP_FIND_LINKS=https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
 RUN pip install .[core,gpu]
 COPY . .
 
+# TODO(samos123): remove this once axlearn upgrades to Jax 0.4.38.
+RUN pip install -U "jax[gpu]==0.4.38" "jax==0.4.38" "jaxlib==0.4.38"
```
kelvin-zou, line 142: can we rely on axlearn's jax version instead?

Dockerfile
kelvin-zou: Second to what Ruoming recommended, maybe split it into Dockerfile.gcp if there are some gcp specific things. and import this docker file as base.  The base image import is fine though.

Dockerfile
samos123: yes will remove this before merge since I think 0.4.37 works as well.

Dockerfile
samos123: There isn't anything GCP specific except installation of `gcloud` which also happens in the axlearn base image. Happy to move to `Dockerfile.gcp` though since it uses a different base image. I think that makes sense.

samos123: I actually think it's better to keep as-is, we don't want 2 separate GPU images, 1 broken and 1 working. At some point, it may be helpful to switch to Jax Stable Stack image so we can re-use the same validated base image on both GPU and TPU.

Are you fine with keeping as-is since there is no GCP specific things happening in the gpu image?

samos123:
I was able to get it working with original Docker image for GPU.  I think the newer Jax and NVIDIA drivers may have removed the need to use the CUDA image from NVIDIA.

@markblee @kelvin-zou could you please review again and get this merged? We've people ready to start adding A4 support and benchmark that. However that work builds on top of this PR. So trying to get this merged ASAP.

I also plan to write an end to end tutorial to train models using AXLearn on GPU so any GCP user can take advantage of AXLEarn. Also those without TPUs.

samos123:
@markblee @kelvin-zou friendly ping. We're starting A4 benchmarking now as well. We are using this PR as the base for that work. So would like to get it merged.

kelvinzou:
> @markblee @kelvin-zou friendly ping. We're starting A4 benchmarking now as well. We are using this PR as the base for that work. So would like to get it merged.

Thanks for updating this PR!

two quick comments
1. we don't really use the axlearn to submit GCP GPU jobs, so this is more like a PoC.
2. given that, can you actually peel off the part that is unique to A3-Ultra? This would be great guidance to us if we ever want to use A3-Ultra. IIUC, comparing to A3-Mega, there is the mlx7 replacing Google NIC and H200 not H100, and the main diff should be only this?

axlearn/cloud/gcp/job.py
```diff
@@ -304,6 +303,11 @@ def from_flags(cls, fv: flags.FlagValues, **kwargs) -> Config:
         cfg.builder = cls.builder.from_flags(fv, **kwargs)
         return cfg
 
+    @classmethod
+    def with_builder(cls, builder: type[BaseReplicatedJob]):
+        cls.builder = builder
+        return cls
```
markblee, line 309: Can you follow the pattern here to avoid in-place update:
https://github.com/apple/axlearn/blob/dcfdbc186cdbed248d77a91c77317f265b01031a/axlearn/cloud/gcp/jobs/launch.py#L214-L216

axlearn/cloud/gcp/jobs/gke_runner.py
```diff
@@ -143,6 +147,11 @@ def validate_inner(cls):
         if cls.inner is None:
             raise ValueError(f"A GKERunnerJob should subclass {cls} and define `inner`.")
 
+    @classmethod
+    def with_inner(cls, inner: type[GKEJob]):
+        cls.inner = inner
```
markblee, line 152: (Same as above.)

axlearn/cloud/gcp/jobs/gke_runner.py
```diff
@@ -531,19 +540,13 @@ def from_flags(cls, fv: flags.FlagValues, **kwargs):
...
     if instance_type.startswith("tpu"):
         return TPUGKERunnerJob
-    elif instance_type.startswith("gpu-a3"):
-        # TODO(markblee): We can directly construct:
-        # GKERunnerJob.with_inner(GKEJob.with_jobset(A3ReplicatedJob))
-        return GPUGKERunnerJob
+    elif instance_type.startswith("gpu-a3-ultra"):
+        return GKERunnerJob.with_inner(GKEJob.with_builder(A3UltraReplicatedJob))
+    elif instance_type.startswith("gpu-a3-high"):
+        return GKERunnerJob.with_inner(GKEJob.with_builder(A3HighReplicatedJob))
```
markblee, line 549: Neat!

markblee (commented):
Thanks! Mostly LGTM. Will let @Ethanlm @kelvinzou leave a review.

axlearn/cloud/gcp/jobs/gke_runner.py
samos123: This is like praising yourself since you came up with it lol

samos123:
> 2. given that, can you actually peel off the part that is unique to A3-Ultra? This would be great guidance to us if we ever want to use A3-Ultra. IIUC, comparing to A3-Mega, there is the mlx7 replacing Google NIC and H200 not H100, and the main diff should be only this?

Correct about difference between A3 Mega and A3 Ultra. Note that total bandwidth is different to A3 Mega (1.6 Gbps) and A3 Ultra (3.2 Gbps).

I think that's what I did in this PR, only separating out the things that are unique to A3 Ultra.

Both A3 High and A3 Mega would require their own classes since they have different sidecars and main container specs. I purposely created a base class in a way that A4 would only have to define a new main container. A3 High and A3 Mega both need sidecars and specific main container.

I think there may be a chance to re-use the same class for both A3 Ultra and A4, but A4X is likely going to need it's own class once again since it's using Arm.

Note we may also build automation to test a multi-node training job on GPU on our side. That would also use this PR.

I will make a follow up commit addressing Mark's comments soon.

axlearn/cloud/gcp/job.py
samos123: done

samos123: tests are failing after making this change.  I will do follow up.

samos123: should be fixed now. Tested both gke_runner and job.

Ethanlm (approved):

axlearn/cloud/gcp/jobset_utils.py
```diff
@@ -670,24 +670,220 @@ def __call__(self) -> Sequence[Nested[Any]]:
...
+            "--xla_gpu_enable_reduce_scatter_combine_by_dim=false",
+            "--xla_disable_hlo_passes=rematerialization",
+            "--xla_gpu_enable_while_loop_double_buffering=true",
+        ]
+        env_vars["XLA_FLAGS"] = " ".join(default_xla_flags)
+
+        # NCCL flags needed
+        env_vars.update(
+            {
+                # Enable auto PGLE available in jax 0.4.33
```
Ethanlm, line 800: Does it still apply if we are already on jax 0.4.38

axlearn/cloud/gcp/jobset_utils.py
```diff
@@ -670,24 +670,220 @@ def __call__(self) -> Sequence[Nested[Any]]:
...
+            }
+        )
+
+        # Override env vars with user provided env vars.
+        env_vars.update(cfg.env_vars)
+        # K8s expects each env variable to be a dict.
+        k8s_env_vars = [{"name": name, "value": value} for name, value in env_vars.items()]
+        k8s_env_vars.append(
+            {
+                "name": "PROCESS_ID",
```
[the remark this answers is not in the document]
samos123: It's an AXLearn specific thing: https://github.com/apple/axlearn/blob/b1e7b375b3b38e57d3c9c7cd772a413ba09cdd22/axlearn/common/launch.py#L78

```
flags.DEFINE_integer(
    "process_id",
    os.environ.get("PROCESS_ID", None),
    "Rank of the current process. Must be None on tpu, otherwise required.",
)
```

We set at as env variable so it works out of the box.

axlearn/cloud/gcp/jobset_utils.py
```diff
@@ -670,24 +670,220 @@ def __call__(self) -> Sequence[Nested[Any]]:
...
+
+        Returns:
+            A nested dict corresponding to a k8s Container config.
+        """
+        cfg: GPUReplicatedJob.Config = self.config
+
+        volume_mounts = [
+            {"name": "shared-memory", "mountPath": "/dev/shm"},
+            {"name": "nvidia-install-dir-host", "mountPath": "/usr/local/nvidia/lib64"},
+            {"name": "gib", "mountPath": "/usr/local/gib"},
```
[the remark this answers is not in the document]
samos123: In A3 High we have a tcpx volume to share a socket.

In A3 Ultra there is a gib volume that's mounted from GKE node. The gib volume provides configs needed for CX7 and some binaries:
```
                "NCCL_SHIMNET_GUEST_CONFIG_CHECKER_CONFIG_FILE": (
                    "/usr/local/gib/configs/guest_config.txtpb"
                ),
                "NCCL_TUNER_CONFIG_PATH": "/usr/local/gib/configs/tuner_config.txtpb",
```

axlearn/cloud/gcp/jobset_utils.py
samos123: removed this comment

samos123:
@kelvin-zou could you do another review and Approve or Request changes? I responded to your comment but did not make any changes based on your comment.

Dockerfile
```diff
@@ -99,6 +99,9 @@ COPY . .
 
 FROM base AS gpu
 
+# Needed for NVIDIA CX7 based RDMA (not cloud specific).
+RUN apt-get update && apt-get install -y ibverbs-utils
```
kelvin-zou, line 103: Nice!

axlearn/cloud/gcp/jobset_utils.py
```diff
@@ -670,24 +670,220 @@ def __call__(self) -> Sequence[Nested[Any]]:
...
+            "--xla_gpu_enable_reduce_scatter_combine_by_dim=false",
+            "--xla_disable_hlo_passes=rematerialization",
+            "--xla_gpu_enable_while_loop_double_buffering=true",
+        ]
+        env_vars["XLA_FLAGS"] = " ".join(default_xla_flags)
+
+        # NCCL flags needed
+        env_vars.update(
+            {
+                "JAX_ENABLE_PGLE": "True",
```
kelvin-zou, line 801: let's turn off PGLE, PGLE has some accuracy issues which we haven't been able to triage.

axlearn/cloud/gcp/jobset_utils.py
```diff
@@ -670,24 +670,220 @@ def __call__(self) -> Sequence[Nested[Any]]:
...
+                "JAX_PGLE_PROFILING_RUNS": "3",
+                # This is needed for flash attention + auto PGLE to work
+                "JAX_REMOVE_CUSTOM_PARTITIONING_PTR_FROM_CACHE_KEY": "True",
+                "CUDA_DEVICE_MAX_CONNECTIONS": "1",
+                "NVTE_FUSED_ATTN": "1",
```
kelvin-zou, line 806: not useful, we are not using TE, instead cudnn which doens't require this flag.

kelvin-zou (commented):
Thanks, the change looks great overall, some small suggestions.

axlearn/cloud/gcp/jobset_utils.py
```diff
@@ -670,24 +670,220 @@ def __call__(self) -> Sequence[Nested[Any]]:
...
+                # Needed to help resolve GPU OOM on fuji v2 70B
+                "XLA_PYTHON_CLIENT_MEM_FRACTION": "0.85",
+                "TF_FORCE_GPU_ALLOW_GROWTH": "true",
+                "NCCL_DEBUG": "INFO",
+                "NCCL_SOCKET_IFNAME": "=eth0,eth1",
+                "NCCL_CROSS_NIC": "0",
+                "NCCL_NET_GDR_LEVEL": "PIX",
+                "NCCL_P2P_NET_CHUNKSIZE": "65536",
```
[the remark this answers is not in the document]
samos123: The tcpx flags are only relevant on A3 High and they're still in that class. On A3 Mega it's yet another set of NCCL TCPXO flags.

On A3 Ultra and A4  we also use different NCCL flags. It looks like for different accelerators we're recommending different flags as the default. There is less of a difference between A3 Ultra and A4 however A4X is yet another big change because now we also have to deal with Arm and lower amount of GPUs per node.

On GCP right now it's better to have per machine type classes because they are turning out to be so different.

I do like the idea of using recomposable NCCL flags. Some of the XLA and NCCL flags could be re-usable potentially between GPU accelerators. I suggest we make that change as part of the A4 PR so we have a better view of how different the flags are. There will be an A4 PR very soon :)

axlearn/cloud/gcp/jobset_utils.py
samos123: Removed. Great catch!  This was coming from Maxtext which I believe was using TE.  I was unaware that this only impacted TE.

axlearn/cloud/gcp/jobset_utils.py
samos123: Removed, but will impact performance.

Our Jax GPU team would love to improve that. I do understand that you probably don't have bandwidth to triage right now. Let me know if you feel there is an easy way for me to reproduce in AXLearn and I Can have someone take that on.

axlearn/cloud/gcp/jobset_utils.py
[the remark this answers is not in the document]
samos123: @kelvin-zou I am fine to make it in this PR too if you feel strongly about it. I do see the value in it, just prefer deferring it to A4 work since that's basing off this PR right now.

samos123:
We would really like to get this merged since main keeps on having significant changes to these files. This is critical for Google to be able to run AXLearn benchmarks. Can you please confirm whether you're fine with the changes overall so I can rebase and handle the conflicts?

We have an A4/B200 PR that will be submitted soon after this PR.

samos123:
Closing this in favor for https://github.com/apple/axlearn/pull/1097 which adds A3 Ultra, A4 and A3 Mega support.

PR #956, opened 2025-01-28
Use InputDispatcher for fuji models

hanzhi713:
Now, `trainer_batch_size` is the global logical batch size. This change makes it easier for external users to test out smaller per device batch size.

### DIFF
--- a/axlearn/experiments/text/gpt/common.py
+++ b/axlearn/experiments/text/gpt/common.py
@@ -54,6 +54,7 @@ from axlearn.common.embedding import TransformerTextEmbeddings
 from axlearn.common.evaler import BaseMetricCalculator, ModelSummaryAccumulator, SpmdEvaler
 from axlearn.common.evaler import every_n_steps_policy as eval_every_n_steps_policy
 from axlearn.common.flash_attention.layer import FlashAttention
+from axlearn.common.input_dispatch import InputDispatcher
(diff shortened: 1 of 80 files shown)


axlearn/experiments/text/gpt/common.py
```diff
@@ -636,6 +637,7 @@ def get_trainer_config_fn(
     keep_every_n_steps: int = 50_000,
     save_every_n_steps: Optional[int] = None,
     init_state_builder: Optional[state_builder.Builder.Config] = None,
+    logical_feed_indices: Optional[Sequence[int]] = None,
```
ruomingp, line 640: Do we need this? I think InputDispatcher can figure it out automatically.

hanzhi713: I think it's still needed when per host bs < 1. In other cases, it's optional.

ruomingp: See https://github.com/apple/axlearn/blob/fad264b2b2d60746585cdbc46089528ee037fb56/axlearn/common/input_dispatch.py#L62-L64.

hanzhi713: Done. Then it looks like a lot of our internal models have this redundantly specified.

ruomingp (approved):

PR #958, opened 2025-01-29
SplashAttention performance tuning for v6e

hanzhi713:
This PR adds a config modifier to modify the block size of SplashAttention. By setting it to 1024, it improves the performance of SplashKernel by 1/3 compared to baseline on v6e.


ruomingp (approved):

hanzhi713:
@ruomingp Fixed a pytype problem and updated golden config. Please re-approve, thanks!

markblee (approved):

PR #968, opened 2025-02-06
Don't keep initial key/value inputs in the KV cache.

ds-hwang:
The current code is weird. It stores the input key/value in the KV cache, but this doesn’t make sense in either init_states or prefill:
* init_states: This is not prefill, so key/value should not be stored in the KV cache.
* prefill: The extend_step() function overrides this part anyway.

Thus, this PR removes this unnecessary and confusing logic. The logic was introduced in https://github.com/apple/axlearn/pull/860


ds-hwang:
Could you review? From 1031

markblee (approved):

ds-hwang:
Thank you for review!

PR #972, opened 2025-02-07
Use broadcasting trick for KV update

changlan:
We implement a broadcast-based approach that supports multi-step update and avoids the expensive scatter operations in prefill.

On v5p-8 with 70B model: 
- prefill: 2.2s -> 71ms
- decode: 107ms -> 69ms

### DIFF
--- a/axlearn/common/attention.py
+++ b/axlearn/common/attention.py
@@ -845,17 +845,16 @@ class BaseQKVLinear(BaseLayer):
             k_proj = k_proj.astype(cached_key.dtype)
             v_proj = v_proj.astype(cached_value.dtype)
 
-            # TODO(dhwang2): jax.lax.dynamic_update_slice_in_dim is generally faster than advanced
-            # indexing, but an unusual slowdown was observed, with RLHF sampling taking up to
-            # 3 hours per run. Investigate and fix it.
-            # Note: All X_idx are small, so generating them on-demand is not costly.
-            b, _, n, h = cached_key.shape
-            b_idx = jnp.arange(b)[:, None, None, None]
-            t_idx = (jnp.arange(k_proj.shape[1])[None] + time_step[:, None])[:, :, None, None]
-            n_idx = jnp.arange(n)[None, None, :, None]
-            h_idx = jnp.arange(h)[None, None, None, :]
-            k_proj = cached_key.at[b_idx, t_idx, n_idx, h_idx].set(k_proj)
-            v_proj = cached_value.at[b_idx, t_idx, n_idx, h_idx].set(v_proj)
+            # Function to update the cache for a single batch element.
+            def update_single(cached_kv_slice, kv_proj_slice, time_idx):
+                return jax.lax.dynamic_update_slice_in_dim(
+                    cached_kv_slice, kv_proj_slice, time_idx, axis=0
+                )
+
+            # Use jax.vmap to vectorize over the batch dimension.
+            vmap_update = jax.vmap(update_single)
+            k_proj = vmap_update(cached_key, k_proj, time_step)
+            v_proj = vmap_update(cached_value, v_proj, time_step)
 
             updated_state.update(key=k_proj, value=v_proj)
         return updated_state, self.Output(query=q_proj, key=k_proj, value=v_proj)
--- a/axlearn/common/attention.py
+++ b/axlearn/common/attention.py
@@ -803,7 +803,7 @@ class BaseQKVLinear(BaseLayer):
         Args:
             cached_states: A `NestedTensor` object containing tensors which are the results of
                 previous attentions, and index used for fast decoding. Contains "key" and "value" of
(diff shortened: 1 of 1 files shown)


ruomingp (approved):

dongyin92 (approved):

axlearn/common/attention.py
```diff
@@ -837,25 +837,41 @@ def extend_step(
...
-            # 3 hours per run. Investigate and fix it.
-            # Note: All X_idx are small, so generating them on-demand is not costly.
-            b, _, n, h = cached_key.shape
-            b_idx = jnp.arange(b)[:, None, None, None]
-            t_idx = (jnp.arange(k_proj.shape[1])[None] + time_step[:, None])[:, :, None, None]
-            n_idx = jnp.arange(n)[None, None, :, None]
-            h_idx = jnp.arange(h)[None, None, None, :]
-            k_proj = cached_key.at[b_idx, t_idx, n_idx, h_idx].set(k_proj)
-            v_proj = cached_value.at[b_idx, t_idx, n_idx, h_idx].set(v_proj)
+            target_len = cached_key.shape[1]
```
dongyin92, line 844: From the doc string of this function, it says `cached_states` contains "key" and "value" of shape [batch, num_heads, per_head_dim, target_length]. Then why does `target_len = cached_key.shape[1]`?

changlan: I think the docstring is stale. Let me update it as well.

markblee: Usually we use `source_len` to refer to the key/value.

axlearn/common/attention.py
```diff
@@ -837,25 +837,41 @@ def extend_step(
...
+            k_proj = (k_proj * oh_indices).astype(cached_key.dtype)
+            v_proj = (v_proj * oh_indices).astype(cached_value.dtype)
+
+            # Sum across step dimension to get final projection values: [B, T, N, H]
+            k_proj = jnp.sum(k_proj, axis=1)
+            v_proj = jnp.sum(v_proj, axis=1)
+
+            # Combine with cached values
+            k_proj = (cached_key * negated_oh_indices) + k_proj
+            v_proj = (cached_value * negated_oh_indices) + v_proj
```
markblee, line 874: I wonder if we can simplify a bit:
```suggestion
            source_len = cached_key.shape[1]
            
            # Create a dispatch matrix of shape [B, T=step, S].
            oh_indices = jax.nn.one_hot(
                time_step[:, None] + jnp.arange(num_query_steps), source_len, dtype=k_proj.dtype
            )
            # Create a mask of shape [B, S, 1, 1].
            negated_oh_indices = (1 - oh_indices.sum(axis=1))[..., None, None]
            k_proj = jnp.einsum("bt...,bts->bs...", k_proj, oh_indices)
            v_proj = jnp.einsum("bt...,bts->bs...", v_proj, oh_indices)
            k_proj = cached_key * negated_oh_indices + k_proj
            v_proj = cached_value * negated_oh_indices + v_proj
```

markblee (commented):
Thanks, @changlan !

axlearn/common/attention.py
changlan: This looks great! I updated the impl.

markblee (approved):

axlearn/common/attention.py
ds-hwang: Is this code as fast as `dynamic_update_slice_in_dim` during prefill?  
I'm a bit concerned about having to create a [B, T, S] tensor, especially for very long prefill like pdf or speech or video.
I want to take this opportunity to learn the speed relationship between **dynamic slice, broadcast, and advanced indexing**.

axlearn/common/attention.py
```diff
@@ -837,25 +837,22 @@ def extend_step(
...
-            # 3 hours per run. Investigate and fix it.
-            # Note: All X_idx are small, so generating them on-demand is not costly.
-            b, _, n, h = cached_key.shape
-            b_idx = jnp.arange(b)[:, None, None, None]
-            t_idx = (jnp.arange(k_proj.shape[1])[None] + time_step[:, None])[:, :, None, None]
-            n_idx = jnp.arange(n)[None, None, :, None]
-            h_idx = jnp.arange(h)[None, None, None, :]
-            k_proj = cached_key.at[b_idx, t_idx, n_idx, h_idx].set(k_proj)
-            v_proj = cached_value.at[b_idx, t_idx, n_idx, h_idx].set(v_proj)
+            source_len = cached_key.shape[1]
```
ds-hwang, line 844: Could you add comment why we don't use dynamic_update_slice for future code reader?

changlan: Sure, done.

axlearn/common/attention.py
changlan: Indeed, so I did a profile with 8k length prefill, dynamic slice and broadcast are roughly the same, and indexing is 2 orders of magnitude slower.

ruomingp (approved):

PR #978, opened 2025-02-11
Allow metrics layers to have state.

markblee:


axlearn/common/layers.py
```diff
@@ -1524,3 +1525,29 @@ def forward(self, x: Tensor) -> Tensor:
         self.add_state_update("value", new_moving_average)
         self.add_state_update("count", 1 + self.parameters["count"])
         return new_moving_average
+
+
+class BaseLossMetrics(BaseLayer):
```
ruomingp, line 1530: layers.py doesn't seem the right file for this class. Can we keep it in metrics.py or create a new file?

ruomingp (changes requested):

axlearn/common/layers.py
markblee: Keeping in metrics.py leads to a circular import, so I moved to a new file `axlearn.common.loss_metrics`. I considered moving the other `*LossMetrics` classes into the file, but it'll cause quite a bit of golden config updates so left a TODO to unblock for now. WDYT?

markblee: (Also responded to internal comment.)

ruomingp (approved):

PR #979, opened 2025-02-11
Transpose kv cache for better decode performance

changlan:
Transposing KV state as a TPU fusion optimization improves decode step time.


changlan:
Pending internal bump PR.

ruomingp (approved):

markblee (approved):
Thanks!

PR #981, opened 2025-02-11
Context parallelism for TPU

hanzhi713:
This PR implements context parallel. This is an improvement compared to the "seq" axis in axlearn, as it removes the expensive resharding from partitioning sequence dim to head dim before and after attention. Currently, this parallelism type is only supported on TPU. See https://docs.nvidia.com/megatron-core/developer-guide/latest/api-guide/context_parallel.html for a description of this parallelism method.

TPU splash attention has internal support for context parallel (e.g. `q_seq_shards` and `manual_sharding_spec`). This PR moves the mask computation logic of splash to outside of the shard map, and then partitions the computed mask through shard map, hence generating the correct mask for  each shard when Q is sequence sharded.

Additional notes:
1. This PR will break existing runs that set seq > 1 and doesn't use the sharding spec from `layer.py`.
2. This PR enables segment id support for SplashAttention. SplashAttention supports segment ids, but it was not enabled previously for some unknown reasons.
3. This PR enables GQA/MQA optimization in SplashAttention by not repeating kv heads before passing kv to Splash.


markblee (changes requested):
(Pending internal comments and CI.)

kelvin-zou (approved):

ruomingp (approved):

hanzhi713:
@markblee Could you please approve this PR?

markblee:
As discussed, let's make sure internal passes first.

hanzhi713:
@markblee Internal CI passed.

markblee (approved):

hanzhi713:
@markblee I pushed a fix that's required after #995. See https://github.com/apple/axlearn/pull/981/commits/bbad4f0876f6e9af9ff333b49a031640e914ffd0#diff-22479d8930dc7b23057c6517b633f9c03c00e805e1099f9c6f30fd9da2109f05R291

markblee (approved):

hanzhi713:
@markblee Could you please re-approve? Thanks. Ran all flash-related TPU tests overnight and some small fixes is required for the test.

ruomingp (approved):

PR #984, opened 2025-02-12
GoodPut minor fix: only process 0 should start goodput uploader

jiya-zhang:
Line 106 in this file specifies that only worker 0 (`jax.process_index() == 0`) will create a GoodputMonitor object and return it. All other processes will return nothing and therefore introduce an error in `self._monitor.start_goodput_uploader(*args, **kwargs)`. Example error: 

```
 File "/root/axlearn/common/measurement.py", line 148, in start_monitoring
    global_recorder.start_monitoring()
  File "/root/axlearn/cloud/gcp/measurement.py", line 110, in start_monitoring
    self._monitor.start_goodput_uploader(*args, **kwargs)
  File "/opt/venv/lib/python3.10/site-packages/ml_goodput_measurement/src/monitoring.py", line 137, in start_goodput_uploader
    if self._uploader_thread_running:
AttributeError: 'GoodputMonitor' object has no attribute '_uploader_thread_running'
```

This PR fixes this error. cc @dipannita08 @markblee

### DIFF
--- a/axlearn/cloud/gcp/measurement.py
+++ b/axlearn/cloud/gcp/measurement.py
@@ -107,5 +107,6 @@ class GoodputRecorder(measurement.Recorder):
                 include_badput_breakdown=True,
             )
 
-        self._monitor.start_goodput_uploader(*args, **kwargs)
-        logging.info("Started Goodput upload to Tensorboard in the background!")
+        if jax.process_index() == 0:
+            self._monitor.start_goodput_uploader(*args, **kwargs)
+            logging.info("Started Goodput upload to Tensorboard in the background!")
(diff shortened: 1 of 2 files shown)


apghml (commented):
Thanks! Can we add a test that fails without this change? (E.g., mock the value of `jax.process_index()` and the `start_goodput_uploader()` function and check whether the function is called based on the process index?)

markblee (commented):
Thanks! +1 on adding a unit test.

jiya-zhang:
Just added a unit test. PTAL!

markblee (approved):

PR #987, opened 2025-02-12
Fix Missing return statement in base_layer_test.py::ExplicitFanLayer::_compute_fan_axes

apivovarov:
### Description

This PR addresses missing return statements in base_layer_test.py::ExplicitFanLayer::_compute_fan_axes

1. **`ComputeFanAxesTest`**:
   - Fixed the missing return statement in the `_compute_fan_axes` method to ensure it returns the result of `super()._compute_fan_axes(name, parameter_spec)` as intended.

#### pre-commit / pytype
```
$ pre-commit run --files $(git diff --name-only HEAD~1)
Check Yaml...........................................(no files to check)Skipped
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
black....................................................................Passed
isort....................................................................Passed
pylint...................................................................Passed
(axlearn-venv-cpu) 

$ pytype -j auto $(git diff --name-only HEAD~1)        
...
Success: no errors found
```

### DIFF
--- a/axlearn/common/base_layer_test.py
+++ b/axlearn/common/base_layer_test.py
@@ -614,7 +614,7 @@ class ComputeFanAxesTest(TestCase):
         def _compute_fan_axes(self, name: str, parameter_spec: ParameterSpec) -> Optional[FanAxes]:
             if name == "fan_axes_specified_weight":
                 raise RuntimeError("Should not be invoked.")
-            super()._compute_fan_axes(name, parameter_spec)
+            return super()._compute_fan_axes(name, parameter_spec)
 
     @parameterized.parameters(
         (


axlearn/common/struct.py
```diff
@@ -30,6 +30,8 @@ def _sort_fields(klass: _T) -> _T:
     klass.__dataclass_fields__ = dict(
         sorted(klass.__dataclass_fields__.items(), key=lambda x: x[0])
     )
+    # Return the class after sorting the fields (for functional style)
+    return klass
```
apghml, line 34: It seems like this could be a footgun since people might not realize it also mutates its argument. Whereas trying to use the returned None value will loudly fail.

axlearn/common/checkpointer.py
```diff
@@ -803,6 +803,7 @@ def __enter__(self):
         if self._within_context:
             raise ValueError("Already in a context.")
         self._within_context = True
+        return self
```
apghml, line 806: I might be mistaken, but IIRC, it's not required for `__enter__()` to return anything in Python. I'm not necessarily opposed to adding the return here though.

apghml (commented):
Thanks for the PR!

apghml:
OOI, was your PR description AI-generated? If so, it might be better to manually write your pr description in the future to avoid mistakes like (2).

axlearn/common/checkpointer.py
apivovarov: https://docs.python.org/3/reference/datamodel.html#object.__enter__
```
object.__enter__(self)

Enter the runtime context related to this object. 
The "with" statement will bind this method’s return value 
to the target(s) specified in the as clause of the statement, if any.
```

Typical signature for `__enter__` method is
```
def __enter__(self) -> Self:
```

apivovarov:
In the BaseCheckpointer/Checkpointer class, the `__enter__()` method should return `self` because this allows the instance to be used within a `with` statement, giving access to its attributes and methods inside the block.

Example:
https://github.com/apple/axlearn/blob/main/axlearn/common/checkpointer_test.py#L748

```py
    def test_context(self, checkpointer_cls):
        ckpt = _checkpointer_config(checkpointer_cls).instantiate(parent=None)

        if checkpointer_cls is Checkpointer:
            with ckpt:
                self.assertIsNotNone(ckpt._gc_thread)
            self.assertIsNone(ckpt._gc_thread)
```

axlearn/common/struct.py
apivovarov: The function declaration says 
```
def _sort_fields(klass: _T) -> _T:
```
Seems that the function should return input parameter, following a functional programming style...

axlearn/common/checkpointer.py
apghml: I think it's not required though. If you look at some of the example in the python docs, they sometimes show returning things other than the object itself.

It's usually redundant to return the object itself since in order to do `with xyz` you already need xyz in scope so you can then use it in the body of the with without needing to do `with xyz as xyz`.

axlearn/common/checkpointer.py
```diff
@@ -950,6 +951,7 @@ def __init__(self, cfg: Config, *, parent: Optional[Module]):
     def __enter__(self):
         super().__enter__()
         self._start_gc_thread()
```
apghml, line 953: Could you clarify whether the things you are changing in this and your other PR were found with the assistance of AI code checking tools?

axlearn/common/struct.py
apghml: In that case, we should remove the return type annotation. I believe we rely on the mutation behavior.

axlearn/common/checkpointer.py
apivovarov: just standard mypy (static type checker for Python)

axlearn/common/checkpointer.py
apivovarov: ok, removed axlearn/common/checkpointer.py changes from the PR

apivovarov:
reduced the scope of the PR to fixes in base_layer_test.py

axlearn/common/checkpointer.py
apghml: Thanks for the clarification. The standard type checker we use in this project is `pytype`. I imagine if you run mypy it may report a fairly large number of things across the project? While we can continue with this PR, I'm not necessarily sure it makes sense to devote time to fixing everything mypy flags, especially if it is otherwise correct python.

But if you are interested in fixing type checker errors, one useful thing could be to upgrade pytype to the latest version and PR the combination of changes / disabling checks needed to accomplish that.

apghml (approved):
In the future, it may not make sense to PR this kind of change since it is relatively common to have a mocked implementation of a subclass used for testing puruposes that doesn't fully faithfully implement the superclass's contract.

PR #988, opened 2025-02-13
Fix Parameter Redefinition in learner_test.py

apivovarov:
### Description
Removed the line `tree: dict:`
Since tree is already defined as a parameter with type `Nested[Any]`, there's no need to redefine it as `dict`

#### Testing
```
pytest -v -n 96 -m "not (gs_login or tpu or high_cpu or fp64)"  \
axlearn/common/learner_test.py

35 passed
```

### DIFF
--- a/axlearn/common/learner_test.py
+++ b/axlearn/common/learner_test.py
@@ -1452,7 +1452,6 @@ class CompositeLearnerTest(TestCase):
                     rule: The rule from `cfg.masking_rules` to check agains.
                 """
                 cfg = self.config
-                tree: dict
 
                 rule_dict: dict[str, Union[optax.MaskedNode, None]] = cfg.rules[rule]
 


axlearn/common/learner_test.py
```diff
@@ -1452,7 +1452,6 @@ def _check_masking(self, tree: Nested[Any], rule: str):
                     rule: The rule from `cfg.masking_rules` to check agains.
                 """
                 cfg = self.config
-                tree: dict
```
apghml, line 1455: `Nested[Any]` is actually equivalent to `Any` from a type checking perspective since `Nested[XYZ]` allows for a bare `XYZ` not contained in any dict. So they are not equivalent.

apghml (commented):
Thanks for the PR!

axlearn/common/learner_test.py
apghml: The line is just shorthand that is similar to `tree = typing.cast(dict, tree)`

apghml: It only has a purpose in type checking, not when the program executes.

apivovarov: Got it! Thank you for the clarification. Closing the PR....

PR #992, opened 2025-02-13
Basic FP8 Training Support

hanzhi713:
This PR adds FP8 training support using either in-batch scaling or delayed scaling. There's a one scaling factor per tensor.

### DIFF
--- a/axlearn/common/learner_test.py
+++ b/axlearn/common/learner_test.py
@@ -47,7 +47,9 @@ from axlearn.common.optimizers import (
 )
 from axlearn.common.quantized_dot_general.layers import (
     DotGeneralQuantizationType,
+    FP8AmaxHistoryParams,
     QuantizedDotGeneral,
+    get_all_fp8_param_names,
 )
 from axlearn.common.test_utils import TestCase
 from axlearn.common.update_transformation import (
@@ -1546,17 +1548,7 @@ class CompositeLearnerTest(TestCase):
                 OverrideInplaceUpdateTransformation.default_config()
             )
             optimizer.transformation = transformation
-            optimizer.rules = [
-                f".*{x}"
-                for x in [
-                    "input_scale",
-                    "kernel_scale",
-                    "output_grad_scale",
-                    "input_amax_history",
-                    "kernel_amax_history",
-                    "output_grad_amax_history",
-                ]
-            ]
+            optimizer.rules = [f".*{x}" for x in get_all_fp8_param_names()]
         else:
             optimizer = transformation
         cfg = Learner.default_config().set(name="test", optimizer=optimizer)
@@ -1651,15 +1643,13 @@ class CompositeLearnerTest(TestCase):
         updated_params = fwd_bwd_outputs.backward_outputs.updated_params
         # After learner update, amax history should be positive.
         if use_override_inplace_update:
-            self.assertGreater(updated_params["input_amax_history"][0], 0.0)
-            self.assertGreater(updated_params["kernel_amax_history"][0], 0.0)
-            self.assertGreater(updated_params["output_grad_amax_history"][0], 0.0)
+            for x in FP8AmaxHistoryParams:
+                self.assertGreater(updated_params[x.value][0], 0.0)
         else:
             # If using a regualr update transformation, learner will try to use adam rule for
             # these parameters, resulting in negative values.
-            self.assertLess(updated_params["input_amax_history"][0], 0.0)
-            self.assertLess(updated_params["kernel_amax_history"][0], 0.0)
-            self.assertLess(updated_params["output_grad_amax_history"][0], 0.0)
+            for x in FP8AmaxHistoryParams:
+                self.assertLess(updated_params[x.value][0], 0.0)
 
         model_params = jax.tree.map(lambda x: OptParam(x, None, 1.0), updated_params)
         fwd_bwd_outputs, learner_output_collection = step()
--- a/axlearn/common/learner_test.py
+++ b/axlearn/common/learner_test.py
@@ -39,14 +39,21 @@ from axlearn.common.optimizers import (
     ParamEmaState,
     adafactor_optimizer,
     adam_optimizer,
+    adamw_optimizer,
     chain,
     clip_by_global_norm,
+    optax,
     sgd_optimizer,
+)
+from axlearn.common.quantized_dot_general.layers import (
+    DotGeneralQuantizationType,
+    QuantizedDotGeneral,
 )
 from axlearn.common.test_utils import TestCase
 from axlearn.common.update_transformation import (
     ForwardOutputs,
     ForwardPass,
+    OverrideInplaceUpdateTransformation,
     Updates,
     UpdateTransformation,
 )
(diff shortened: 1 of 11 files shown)


axlearn/common/trainer_config_modifier.py
```diff
@@ -277,3 +286,51 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+            f".*/{x}"
+            for x in [
+                "input_scale",
+                "kernel_scale",
+                "output_grad_scale",
+                "input_amax_history",
+                "kernel_amax_history",
+                "output_grad_amax_history",
+            ]
+        ]
```
ruomingp, line 330: Can we use update_rules? See https://github.com/apple/axlearn/blob/daec8c530741594eeea12faba1a454bb42b6c03c/axlearn/common/learner_test.py#L343.

axlearn/common/utils.py
```diff
@@ -1158,6 +1159,55 @@ def prune_tree(
...
+    )
+
+
+def tree_merge(
+    primary: Nested[Any],
+    *,
+    secondary: Nested[Any],
+    override_fn: Callable[[Any, Any], Any] = tree_merge_default_override_fn,
+) -> Nested[Any]:
+    """Merge `secondary` into `primary`. The result contains shallow copies of subtrees from both.
```
ruomingp, line 1188: Consider making them deep copies to be consistent with the jax tree utils. Deep copies are also less error prone by decoupling side effects.

axlearn/common/utils.py
```diff
@@ -1158,6 +1159,55 @@ def prune_tree(
...
+    determine which leaf is kept in the resulting tree.
+
+    The default `override_fn` choses the non-empty leaf. If both leaves are non-empty, an error
+    will be raised.
+    """
+    if isinstance(primary, dict) ^ isinstance(secondary, dict):
+        raise ValueError(f"Trying to merge incompatible subtrees: {primary=}, {secondary=}")
+    if not (isinstance(primary, dict) or isinstance(secondary, dict)):
+        return override_fn(primary, secondary)
+    # Use type() so that if primary is a VDict, out_tree is also a VDict.
```
ruomingp, line 1201: Can we be even more strict?
```suggestion
    if type(primary) != type(secondary):
        raise ValueError(...)
```

axlearn/common/utils.py
```diff
@@ -1158,6 +1159,55 @@ def prune_tree(
...
+    out_tree = type(primary)(primary)
+    for k in secondary:
+        if k in primary:
+            out_tree[k] = tree_merge(primary[k], secondary=secondary[k], override_fn=override_fn)
+        else:
+            out_tree[k] = secondary[k]
```
ruomingp, line 1207:
```suggestion
            out_tree[k] = copy.deepcopy(secondary[k])
```

axlearn/common/utils.py
```diff
@@ -1158,6 +1159,55 @@ def prune_tree(
...
+    path in `primary`. If there are identical path with different leaves, `override_fn` is used to
+    determine which leaf is kept in the resulting tree.
+
+    The default `override_fn` choses the non-empty leaf. If both leaves are non-empty, an error
+    will be raised.
+    """
+    if isinstance(primary, dict) ^ isinstance(secondary, dict):
+        raise ValueError(f"Trying to merge incompatible subtrees: {primary=}, {secondary=}")
+    if not (isinstance(primary, dict) or isinstance(secondary, dict)):
+        return override_fn(primary, secondary)
```
ruomingp, line 1200:
```suggestion
        return copy.deepcopy(override_fn(primary, secondary))
```

axlearn/common/utils.py
```diff
@@ -1158,6 +1159,55 @@ def prune_tree(
...
+        f"Encountered incompatible subtree leaves: {primary=}, {secondary=}. Specify "
+        "a custom override function to resolve incompatible subtree merges."
+    )
+
+
+def tree_merge(
+    primary: Nested[Any],
+    *,
+    secondary: Nested[Any],
+    override_fn: Callable[[Any, Any], Any] = tree_merge_default_override_fn,
```
ruomingp, line 1186:
```suggestion
    leaf_merge_fn: Callable[[Any, Any], Any] = tree_merge_default_override_fn,
```

ruomingp, line 1186: How about using a simpler function (always choose primary) as the default merge function?

```suggestion
    override_fn: Callable[[Any, Any], Any] = lambda x, y: x,
```

ruomingp (commented):
Thanks.

axlearn/common/trainer_config_modifier.py
hanzhi713: No. There is a tracer problem when plumbing the FP8 stats through as state updates, because they are produced in the backward pass of a custom vjp. To circumvent this problem, we produce them as gradient and treat them as state updates.

See the internal PR 806 and 816 for context and discussions.

axlearn/common/utils.py
hanzhi713: Done.

axlearn/common/utils.py
hanzhi713: Done.

axlearn/common/utils.py
hanzhi713: Done.

axlearn/common/utils.py
hanzhi713: Done.

axlearn/common/utils.py
hanzhi713: Done.

hanzhi713: I think the current default makes more sense. Always choose primary will discard subtree of secondary even when the primary tree has an empty leaf or None at the corresponding position.

hanzhi713:
@ruomingp Could you please take a look again?

ruomingp:
> @ruomingp Could you please take a look again?

Sorry about the delay. Will take a look soon.

axlearn/common/trainer_config_modifier.py
```diff
@@ -277,3 +286,51 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+        transformation = maybe_instantiate(cfg.learner.optimizer)
```
ruomingp, line 331: Usually we avoid instantiation when constructing configs to preserve the readability of configs. Can we defer the `maybe_instantiate` call to `OverrideInplaceUpdateTransformation.__init__`?

axlearn/common/trainer_config_modifier.py
```diff
@@ -277,3 +286,51 @@ def __call__(self, cfg: SpmdTrainer.Config) -> SpmdTrainer.Config:
...
+        )
+        update_cfg.rules = [
+            f".*/{x}"
+            for x in [
+                "input_scale",
+                "kernel_scale",
+                "output_grad_scale",
+                "input_amax_history",
+                "kernel_amax_history",
+                "output_grad_amax_history",
```
ruomingp, line 328: I wonder how we can keep this list in sync with quantized_dot_general/layers.py. Consider 
* Defining them as an enum in `quantized_dot_general/common.py` and enumerate the enum members here
* Move this class to `quantized_dot_general/update_transformation.py`
* Change quantized_dot_general/layers.py to use the enums instead of the string literals?

axlearn/common/update_transformation.py
```diff
@@ -263,3 +271,77 @@ def mask_tree(tree: dict, *, keep: dict, mask_value: Any) -> dict:
...
         is_leaf=lambda x: x is None,
     )
+
+
+class OverrideInplaceUpdateTransformation(WrappedPartitionedGradientTransformation):
+    """An update transformation that provides rules to override inplace updates.
+
+    This update transformation moves gradients that match rules in `delta_updates` to
+    `inplace_updates`, then applies `PartionedGradientTransformation.update`. Also, optimizer
+    states won't be created for parameters that match these rules.
```
ruomingp, line 281: Another way to look at this is that we are applying different optimizers to different params. Specifically, we apply a special optimizer for some params where the optimzier simply takes the "gradient" as the new param value.

If this makes sense, we can introduce the concept of a composite optimizer containing sub-optimizers, similar to https://github.com/apple/axlearn/blob/fde078db3a4733cfd4c3a67b83276329f2fb7716/axlearn/common/learner.py#L379-L388 and it can be configured accordingly.

Would this be a more general solution?

ruomingp (commented):
Apologies about the delay.

axlearn/common/trainer_config_modifier.py
hanzhi713: Done.

axlearn/common/trainer_config_modifier.py
hanzhi713: Done.

axlearn/common/update_transformation.py
hanzhi713: Unfortunately, this cannot be implemented in learner as a regular optimizer, as all optimizer updates are delta updates, meaning that the output of the optimizer will be added to the original value. Therefore, the implementation of the "replace" logic  must be done through changing "inplace_updates".

axlearn/common/learner_test.py
```diff
@@ -1512,6 +1519,158 @@ def _check_masking(self, tree: Nested[Any], rule: str):
...
+            optimizer.transformation = transformation
+            optimizer.rules = [
+                f".*{x}"
+                for x in [
+                    "input_scale",
+                    "kernel_scale",
+                    "output_grad_scale",
+                    "input_amax_history",
+                    "kernel_amax_history",
+                    "output_grad_amax_history",
```
ruomingp, line 1557: Use the constants?

axlearn/common/learner_test.py
```diff
@@ -1512,6 +1519,158 @@ def _check_masking(self, tree: Nested[Any], rule: str):
...
+        if use_override_inplace_update:
+            self.assertGreater(updated_params["input_amax_history"][0], 0.0)
+            self.assertGreater(updated_params["kernel_amax_history"][0], 0.0)
+            self.assertGreater(updated_params["output_grad_amax_history"][0], 0.0)
+        else:
+            # If using a regualr update transformation, learner will try to use adam rule for
+            # these parameters, resulting in negative values.
+            self.assertLess(updated_params["input_amax_history"][0], 0.0)
+            self.assertLess(updated_params["kernel_amax_history"][0], 0.0)
+            self.assertLess(updated_params["output_grad_amax_history"][0], 0.0)
```
ruomingp, line 1662: Use the constants from FP8_AMAX_HISTORY_PARAM_NAMES?

axlearn/common/quantized_dot_general/layers.py
```diff
@@ -63,6 +64,14 @@ class ClippingChoice(Enum):
     OUTPUT_ACTIVATION = 1
 
 
+FP8_SCALE_PARAM_NAMES = ["input_scale", "kernel_scale", "output_grad_scale"]
```
ruomingp, line 67: Consider using enum:

```suggestion
class Fp8ScaleParam(enum.Enum):
    INPUT_SCALE = "input_scale"
    ...
```

This avoids spelling mistakes.

axlearn/common/trainer_config_modifier_test.py
```diff
@@ -174,5 +178,42 @@ def test_chain_config_modifier(self):
...
+        self.assertIsInstance(cfg.learner.optimizer, OverrideInplaceUpdateTransformation.Config)
+        self.assertEqual(
+            cfg.learner.optimizer.rules,
+            [
+                ".*/input_scale",
+                ".*/kernel_scale",
+                ".*/output_grad_scale",
+                ".*/input_amax_history",
+                ".*/kernel_amax_history",
+                ".*/output_grad_amax_history",
```
ruomingp, line 212: Use the constant names?

axlearn/common/update_transformation.py
hanzhi713: Using that method is not very reliable when there's inf.

axlearn/common/utils.py
```diff
@@ -1158,6 +1159,63 @@ def prune_tree(
     return in_tree
 
 
+def tree_merge_default_leaf_merge_fn(primary: Any, secondary: Any):
```
ruomingp, line 1162: The behavior of this function still seem too complicated to be the default function for tree_merge. Can we ask the user to specify the merge function explicitly?

axlearn/common/learner_test.py
hanzhi713: Done.

axlearn/common/quantized_dot_general/layers.py
hanzhi713: Done.

axlearn/common/trainer_config_modifier_test.py
hanzhi713: Done.

axlearn/common/utils.py
hanzhi713: Done.

ruomingp (approved):

PR #993, opened 2025-02-14
Improvements and fixes to gradient accumulation

apoorvtintin:
- Fix to `with_minibatch_steps` decorator to generate correct primal outputs shapes.
- Improved `with_minibatch_steps` to shard the minibatch with same sharding as original input batch.

Misc:
- Enable gradient accumulation for Fuji 3B on TRN2


axlearn/experiments/text/gpt/fuji.py
```diff
@@ -352,6 +354,20 @@ def get_trainer_kwargs(
...
+                            GradientAccumulationModifier.default_config().set(
+                                grad_acc_steps=4,
+                                minibatch_partitioner=config_for_function(
+                                    partition_by_path_rank
+                                ).set(
+                                    path_rank_to_partition={
+                                        # Note: the batch axes are different here than in
+                                        # `cfg.batch_axis_names`,
+                                        # as we partition sequence dim over `seq`.
+                                        (None, 1): PartitionSpec(("data", "expert", "fsdp")),
```
kelvin-zou, line 366: I am wondering, if we have a default input partition with axis=0 on ("data", "expert", "fsdp") and axis=1 on  "seq", do we still need this?

axlearn/common/gradient_accumulation.py
```diff
@@ -73,23 +75,25 @@ def _make_scan_minibatch_inputs(
...
-            ),
-            inputs["input_batch"],
+    minibatch_input = jax.tree.map(
+        lambda x: jax.lax.dynamic_slice_in_dim(
+            x,
+            start_index=minibatch_index * minibatch_size,
+            slice_size=minibatch_size,
+            axis=0,
         ),
-        input_partition_spec(),
```
kelvin-zou, line 91: To me, it seems rather a hack than a proper solution, that is, we want to have a different `input_partition_spec()` than the default one, then we need this?

axlearn/experiments/text/gpt/fuji.py
apoorvtintin: Thanks for the quick review.
(None, 1) is for the `target_num_bytes` and (None, 2) is for the `input_ids` and `target_labels`, so we need both. Together they will work for most cases, but for the outliers where a specific sharding is required the ability to change sharding for the minibatches will be good to have. But let me know what you think.

Let me know if this answers your question.

axlearn/common/gradient_accumulation.py
apoorvtintin: Sorry I missed the default case, added it.

I think the below partition spec is good as a default, but the ability to change PartitionSpec might be good to have, what do you think?
```
(None, 1): PartitionSpec(("data", "expert", "fsdp")),
(None, 2): PartitionSpec(("data", "expert", "fsdp"), "seq"), 
```

axlearn/common/gradient_accumulation.py
```diff
@@ -57,39 +59,38 @@ def _make_scan_minibatch_inputs(
...
-    Note that this only preserves the input sharding if the `input_partition_spec`
-    returns the correct partition spec to shard the input slices with.
-
     Args:
         inputs: Same pytree as ForwardFn inputs.
         forward_key: The `forward_key` from the ForwardFn inputs
         param_noise_key: The `param_noise_key` from the ForwardFn inputs
         minibatch_size: Size of the minibatch.
         minibatch_index: Current scan minibatch index.
+        minibatch_partitioner: If not None, applies additional sharding constraints
```
apghml, line 76: Could we add unit tests that cover this argument?

apghml (changes requested):

axlearn/common/gradient_accumulation.py
apoorvtintin: Added unit tests covering the argument, both default and custom minibatch_partitioner are tested.

axlearn/common/gradient_accumulation.py
```diff
@@ -57,39 +59,38 @@ def _make_scan_minibatch_inputs(
     param_noise_key: Tensor,
     minibatch_size: int,
     minibatch_index: int,
+    minibatch_partitioner: Optional[InputPartitionFn],
```
apghml, line 62: Echoing Kelvin's comment, could you explain concretely why we need this functionality? If it's just something that might be useful, maybe we can wait until we are certain that we will need it?

axlearn/common/gradient_accumulation.py
```diff
@@ -134,16 +136,32 @@ def with_minibatch_steps(
...
+            additional sharding constraints on each minibatch created.
 
     Returns:
         Decorated ForwardFn.
     """
 
+    # Default partitioner for minibatches.
+    if not minibatch_partitioner:
+        minibatch_partitioner = partition_by_path_rank(
+            path_rank_to_partition={
```
apghml, line 159: Can we default this to the same sharding the input is already using along all non-batch axes?

axlearn/common/gradient_accumulation.py
apghml: If you haven't already, please take a look at:
https://github.com/apple/axlearn/blob/main/CONTRIBUTING.md

> The PR authors are expected to reply to each comment they have addressed, e.g., with "Done". However, they should refrain from resolving the comments -- instead, let the reviewers do so.

axlearn/common/gradient_accumulation.py
apoorvtintin: In the case where gradient accumulation is not enabled, the inputs to the graph are sharded as per the policy in [input_partitioner](https://github.com/apple/axlearn/blob/334f421f8b192414960e472a8cb3ebe7761a164a/axlearn/experiments/text/gpt/common.py#L686). This ensures the batch dimension is sharded on data, expert and fsdp axis while sequence dimension is replicated on model axis.

Gradient accumulation wraps the train steps in a scan loop, while the input_partitioner shards the input batch to  correctly at first. In the gradient accumulation wrapper the input batches are resharded/overridden by the function [_make_scan_minibatch_inputs](https://github.com/apple/axlearn/blob/334f421f8b192414960e472a8cb3ebe7761a164a/axlearn/common/gradient_accumulation.py#L91) and sharded along all axis available which is probably unexpected and inefficient. Minibatches should follow the same PartitionSpec as input_batches.

The addition of the minibatch_partitioner allows the minibatches to use the same sharding/PartitionSpec as input_partitioner provides in the input batches in the case gradient accumulation is not used.

axlearn/common/gradient_accumulation.py
apoorvtintin: Just confirming if I heard it correctly, we want to default to [input_partition_specs](https://github.com/apple/axlearn/blob/334f421f8b192414960e472a8cb3ebe7761a164a/axlearn/common/utils.py#L707) from `utils.py` and not what the [input_partitioner](https://github.com/apple/axlearn/blob/334f421f8b192414960e472a8cb3ebe7761a164a/axlearn/experiments/text/gpt/common.py#L686) sets.

apghml: Not exactly. I was envisioning that for all axes other than axis 0, we default to whatever sharding the input already has. For axis 0, ideally we could also keep whatever sharding the input already has too, although  I'm not sure that would work with logical batching.

axlearn/common/gradient_accumulation.py
apghml: If we just preserve the sharding the input already has, would that also address the concern about the input sharding being changed?

apoorvtintin: Yeah preserving sharding of the input and not having a `sharding_constraint` for minibatches would address the concern as well.

axlearn/common/gradient_accumulation.py
apoorvtintin: > For axis 0, ideally we could also keep whatever sharding the input already has too, although I'm not sure that would work with logical batching

I think preserving the sharding of the input would be perfect, logical batching already inserts the correct sharding constraint after squeezing out the padded batches

apoorvtintin:
Removed additional sharding constraints from gradient accumulation decorator, minibatches now should use the sharding spec created by the `input_partitioner`.

axlearn/common/gradient_accumulation.py
```diff
@@ -78,18 +75,16 @@ def _make_scan_minibatch_inputs(
...
-            inputs["input_batch"],
+    minibatch_input = jax.tree.map(
+        lambda x: jax.lax.dynamic_slice_in_dim(
+            x,
+            start_index=minibatch_index * minibatch_size,
+            slice_size=minibatch_size,
+            axis=0,
         ),
-        input_partition_spec(),
+        inputs["input_batch"],
```
apghml, line 85: Suppose we have a global input batch of size 100 running on 10 chips (so a per host batch size of 10) and we want to switch to doing 10 grad accumulation steps each with a global batch size of 10 (1 per host per accumulation step).

Suppose that the input is originally sharded evenly across the chips (first 10 on first chip, second 10 on second chip, etc). Then when we get the first slice of 10 for the first grad accumulation step, won't all these examples be on the same chip? Will that cause a problem? (E.g., if we worry XLA might not automatically reshard the examples across chips?)

Maybe we should reshard the batch axis only?

axlearn/common/gradient_accumulation.py
```diff
@@ -172,12 +167,26 @@ def fwd_helper(
                 otherwise None.
             """
             minibatch_size = _compute_minibatch_size(inputs["input_batch"], steps=steps)
+
+            # Create a sample minibatch for the carry buffer creation below
```
apghml, line 171: Could you explain in more detail why this is needed?

apoorvtintin: I saw broadcasting errors coming from the scan body, (example below), it complained that the carry buffer shape and the output of minibatch step are incompatible. 
`TypeError: add got incompatible shapes for broadcasting: (32, 4096, 3072), (8, 4096, 3072).`

The carry buffer initialization uses the full batch while creating the buffer, which does not match with the output of ministep which would use the shapes of minibatch.

The simple fix for this is to use a minibatch sample for creating carry buffer ensuring it's shapes are same as the minibatch step.

Let me know if I missed something.

apghml: Do we know why this issue wasn't causing errors before?

apoorvtintin: The unit test does not have any metric/output that relies on batch size which is why it does not catch this issue. I dug a bit deeper and found `output_collection/module_outputs/decoder/transformer/layer3/output` carries batch dimension as output - ref below.

`path (GetAttrKey(name='output_collection'), GetAttrKey(name='module_outputs'), DictKey(key='decoder'), DictKey(key='transformer'), DictKey(key='layer3'), DictKey(key='output')) shape (32, 4096, 3072)`

axlearn/common/gradient_accumulation.py
apoorvtintin: We can completely avoid the batch reshards using a reshape + transpose. I added it to the PR let me know if it addresses your concerns.

Using the same example as @apghml:
> Suppose we have a global input batch of size 100 running on 10 chips (so a per chip size of 10) and we want to switch to doing 10 grad accumulation steps each with a global batch size of 10 (1 per chip per accumulation step).
> Suppose that the input is originally sharded evenly across the chips (first 10 on first chip, second 10 on second chip, etc). Then when we get the first slice of 10 for the first grad accumulation step, won't all these examples be on the same chip? Will that cause a problem? (E.g., if we worry XLA might not automatically reshard the examples across chips?)

Rather than using first 10 batches available in the global batch array for the first iteration, we construct the minibatch using the first batch from every device that is minibatch 0 =>[0, 10, 20 ....]. This is achieved using the reshape and transpose. 

Essentially the logic here is to ensure each device uses local batches avoiding extra reshards.
This also scales well across multiple nodes as each node only runs a local reshape + transpose, also higher per device BS is also supported.

This should addresses the concerns around input batch reshards, let me know if there are still more concerns.

@kelvin-zou I can't think of a way to get size of a specific axis at runtime, but I do believe JAX should be able to give an informative error if the batch size != size of batch axes size.

apoorvtintin:
Addressed comments and added a test that checks if minibatch sharding is same as input batch sharding, This test fails for the original minibatch creation/sharding code.

axlearn/common/gradient_accumulation.py
```diff
@@ -172,12 +160,56 @@ def fwd_helper(
                 otherwise None.
             """
             minibatch_size = _compute_minibatch_size(inputs["input_batch"], steps=steps)
+
+            def reshape_for_scan(x: Tensor):
+                """Helper function that adds a minibatch dimension while evenly dividing
+                batches across gradient accumulation iterations.
+
+                Input dimension is [GBS, seq], this first reshaped to [MBS, steps, seq],
```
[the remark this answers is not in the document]
apoorvtintin: Fixed

apoorvtintin:
> Thanks for the explanation. Can you add a test that fails without this fix?

@apghml I added a test which would fail for the original case without the fix, let me know if it addresses your concerns.

axlearn/common/gradient_accumulation_test.py
```diff
@@ -1,15 +1,143 @@
...
+            """Simple ForwardFn."""
+            _check_equal_sharding(
+                input_batch=inputs["input_batch"],
+                expected=expected_minibatch_sharding,
+            )
+            loss = -jax.nn.log_softmax(model_params["w"] + model_params["b"])[1]
+            output_collection = new_output_collection()
+            output_collection.state_updates["w"] = model_params["w"] + 1
+            output_collection.state_updates["loss"] = WeightedScalar(loss, 1)
+            return ForwardOutputs(loss=loss, aux={}, output_collection=output_collection)
```
apghml, line 100: IIUC, the issue you fixed regarding the batch size in the output collection only happens if the output collection contains something with a leading `batch_size` dim, which I couldn't find here? Do the tests in the current version of the PR fail without that specific fix? If not, can we add a test that fails without that specific fix (or make sure one of the existing tests will fail without it)?

apoorvtintin: Added a check for this in [TestMinibatchSteps](https://github.com/apple/axlearn/pull/993/files#diff-a6b44b5d010f7278c598c45f37b8a1a8bdc265eecb890207fd307e2707704d5cR164). This test will now fail without the fixes made in this PR.

apoorvtintin: Added tests for both checking sharding for minibatches and for carry buffer creation. Thanks for the reviews @apghml, let me know if more changes are needed.

apoorvtintin:
Added tests for both checking sharding for minibatches and for carry buffer creation. Thanks for the reviews @apghml, let me know if more changes are needed.

axlearn/common/gradient_accumulation_test.py
```diff
@@ -28,11 +157,14 @@ def test_minibatch_steps_grads_and_loss(self, steps):
...
         def loss_fn(*, model_params, inputs) -> ForwardOutputs:
             """Simple ForwardFn."""
-            del inputs
             loss = -jax.nn.log_softmax(model_params["w"] + model_params["b"])[1]
             output_collection = new_output_collection()
             output_collection.state_updates["w"] = model_params["w"] + 1
             output_collection.state_updates["loss"] = WeightedScalar(loss, 1)
+            # This output_collection entry is used to check if the gradient accumulation decorator
+            # correctly handles outputs that depend on batch size during carry buffer creation.
+            output_collection.summaries["output_with_batch_dimension"] = inputs["input_batch"]
```
apghml, line 166: Do we have an assert statement somewhere that checks that does an equality / allclose check on the value returned in the output collection?

apoorvtintin: This is a potential problem, right now accumulation logic for weighted tensors just does a simple addition. This means currently if an output is based on global batch size (example [batch_size,  ...]) then for each minibatch the accumulation would be in a carry buffer dimension [minibatch, ...]. This would mean the output of metrics from non accumulated batch would be different from the same batch decorated by gradient accumulation.

apghml: Thanks, could you add this information to the docstring with a concrete example of a how a specific output tensor like [1,2,3,4] would be affected by using minibatching?

apoorvtintin: Added a docstring here https://github.com/apple/axlearn/pull/993/files#diff-096933a05bbb85e50805cead7cd719b7707d35b0dbd31b7fc6c0e9aed2d54ff9R125. Let me know if this is enough.

apghml (approved):

kelvinzou (approved):

apoorvtintin:
Fixed the skipping logic for the partitioner test. Can we please run the CI again?

Ruixuan (approved):

apoorvtintin:
Updated golden configs for Fuji.

apoorvtintin:
Rebased the PR, all tests pass with `run_tests.sh` script.

axlearn/common/gradient_accumulation.py
```diff
@@ -172,12 +167,54 @@ def fwd_helper(
...
+                    The reshaped tensor.
+                """
+                if x.shape[0] % minibatch_size != 0:
+                    raise ValueError(
+                        f"minibatch_size {minibatch_size} does not evenly divide "
+                        f"global batch size of {x.shape[0]}"
+                    )
+
+                x = x.reshape(minibatch_size, -1, *x.shape[1:])
+                return jnp.swapaxes(x, 0, 1)
```
markblee, line 196: I guess this retains any sharding annotations on global batch size to the mini batch size -- OOI what happens if the minibatch doesn't divide the global batch sharding? Is it silently ignored?

axlearn/common/gradient_accumulation.py
```diff
@@ -172,12 +167,54 @@ def fwd_helper(
...
+
+            inputs["input_batch"] = jax.tree_map(reshape_for_scan, inputs["input_batch"])
+
+            # Create a sample minibatch for the carry buffer creation below
+            (
+                sample_minibatch_inputs,
+                _,
+                _,
+            ) = _make_scan_minibatch_inputs(
```
markblee, line 205: nit -- 
```suggestion
            sample_minibatch_inputs, *_ = _make_scan_minibatch_inputs(
```

axlearn/common/gradient_accumulation.py
```diff
@@ -172,12 +167,54 @@ def fwd_helper(
                 otherwise None.
             """
             minibatch_size = _compute_minibatch_size(inputs["input_batch"], steps=steps)
+
+            def reshape_for_scan(x: Tensor):
+                """Helper function that adds a minibatch dimension while evenly dividing
+                batches across gradient accumulation iterations.
+
+                Input dimension is [Global logical Batch Size, Sequence], this first reshaped to
```
markblee, line 175: nit -- 
```suggestion
                Input dimension is [global_logical_batch_size, seq_len], this first reshaped to
```

ditto for below.

markblee (approved):
Feel free to go ahead w/ merge since @Ruixuan has already tested this PR internally.

axlearn/common/gradient_accumulation.py
apoorvtintin: JAX will throw an exception during the tracing stage if the minibatch size does not divide the global batch. This [test](https://github.com/apple/axlearn/pull/993/files#diff-a6b44b5d010f7278c598c45f37b8a1a8bdc265eecb890207fd307e2707704d5cR116) asserts the same.

PR #994, opened 2025-02-14
Add SystemCharacteristics for v5e-8

changlan:


axlearn/cloud/gcp/system_characteristics.py
```diff
@@ -334,6 +334,15 @@ class _SystemCharacteristics:
         "16x20x28", 2240, "tpu-v5p-slice", "ct5p-hightpu-4t", 4, AcceleratorType["TPU"], "v5p-17920"
     ),
     # v5litepod
+    "v5litepod-8": _SystemCharacteristics(
+        "2x4",
+        2,
+        "tpu-v5-lite-podslice",
+        "ct5lp-hightpu-4t",
+        8,
```
Ethanlm, line 342: Should this be 4 instead?

samos123: There is an 8t vm machine type too where a single vm has 8 chips: https://cloud.google.com/tpu/docs/v5e#v5e-types-for-serving

But I think those are meant for serving.

changlan: Good catch - will fix.

changlan: It is from https://github.com/AI-Hypercomputer/xpk/blob/main/src/xpk/core/system_characteristics.py#L1108.

samos123: I think xpk is wrong but let me ask Victor.

PR #995, opened 2025-02-15
Implemented sliding window attention to maintain KV cache only for the window size to enable infinite decoding.

ds-hwang:
Implemented sliding window attention to maintain KV cache only for the window size to enable infinite decoding.

Currently, when using `MultiheadAttention` or `GroupedQueryAttention` for
sliding window attention, the KV cache is kept for the full sequence length
(`seq_len`) instead of the window length (`window_len`).

For example, a model with `window_len=1k` and `seq_len=2M` keeps a KV cache
for the full 2M tokens. It then biases 1999k invalid KV tokens before
calculating attention, resulting in a computational complexity of **O(2M²)**
instead of the desired **O(1k²)**.

This issue persists even when using flash attention. Flash attention uses the
KV cache allocated in HBM as its input. While unnecessary blocks are discarded
during computation, the KV cache still occupies HBM inefficiently for the full
2M tokens.

To address this, when `MultiheadAttention` detects a sliding window mask, it
stores the key-value (KV) cache in a ring buffer inside the input linear layer.
As a result, downstream projects using `MultiheadAttention` automatically
benefit from efficient KV cache handling in `init_states` and `extend_step`.

Additionally, for use cases like local-global attention in LLMs, it is
recommended to use sliding window masks for even the global attention as well.
For example, if you want to train an LLM with a context length of 8k, you can
set the sliding window size to 8k during training. This enables functionally
infinite decoding during inference. Accuracy wouldn't be good tho.


axlearn/common/attention.py
```diff
@@ -735,34 +937,31 @@ def init_states(
...
-                value=jnp.zeros(
-                    shape=(batch, self.num_kv_heads, cfg.per_head_dim, max_len),
-                    dtype=dtype,
-                ),
-            )
+                kv_shape.update(batch_size=key.shape[0], kv_len=key.shape[1])
+            kv_shape = KVCache.Shape(**kv_shape)
+            if query.dtype is None:
+                dtype = jnp.float32
+                logging.warning("init_states dtype is not set, so fallback to float32")
```
ruomingp, line 949: When does this happen? Can we raise a ValueError instead?

axlearn/common/attention.py
```diff
@@ -843,44 +1042,32 @@ def extend_step(
...
-            source_len = cached_key.shape[-1]
-
-            # [B, T, N, H] --> [B, N, H, T].
-            k_proj = jnp.einsum("btnh->bnht", k_proj)
-            v_proj = jnp.einsum("btnh->bnht", v_proj)
-
-            # Create a dispatch matrix of shape [B, T=step, S].
-            oh_indices = jax.nn.one_hot(
-                time_step[:, None] + jnp.arange(num_query_steps), source_len, dtype=cached_key.dtype
+            with child_context("kv_cache_extend_step", module=self.kv_cache):
```
ruomingp, line 1050: Do we need this child_context?

ds-hwang:
@ruomingp could you review? From 967

axlearn/common/attention.py
ds-hwang: Yes, unfortunately. It's because `init_states()` calls here while `init_states()` already called `kv_cache.init_states()`

changlan (changes requested):
I might be a bit paranoid, but I wonder we could hold on until we resolve all the performance concerns in the internal PR? Thanks!

axlearn/common/attention.py
ds-hwang: No, too many caller doesn't set dtype in the query TensorSpec, because dtype of TensorSpec is optional...

AS-IS dtype was
```
# Default to base layer dtype for initialization if cache_dtype is None.
dtype = cfg.cache_dtype or self.dtype()
assert dtype is not None
```
which uses param's dtype for kv_cache (which is existing bug).

ds-hwang:
> I might be a bit paranoid, but I wonder if we could hold on until we resolve all the performance concerns in the internal PR? Thanks!

I provided benchmark results there. What is not resolved yet?

axlearn/common/attention.py
ds-hwang: Yes, now raise ValueError. Let me fix all downstream about it.

ruomingp:
Will wait for the internal PR to be approved.

changlan (approved):
LGTM. Thanks for reverting the performance impacting changes and rerunning the benchmarks.

ds-hwang:
@changlan Thank you for constructive review!
@ruomingp Could you review again?

ruomingp (approved):

ds-hwang:
Thank you for reivew!

PR #997, opened 2025-02-17
AOT compilation support for inference

changlan:
* Add optional `devices` init argument to InferenceRunner for passing fake devices during AOT compilation.
* Add more v5e slice types.


apghml:
Neat! Is there an easy way to run this with a script, similar to the existing run_aot_compilation.py?

PR #999, opened 2025-02-18
Fix flash decoding in GPU.

ds-hwang:
target_positions used to be time_step, but after PR #995, it now represents the actual target positions with shape [batch, step_len]. https://github.com/apple/axlearn/pull/995

Updating the GPU decoding code to align with this change.

CI did not cover GPU unit tests.

TEST=test_extend_step10 of axlearn/common/flash_attention/layer_test.py in GPU


markblee (commented):
Do we have an internal bump PR?

ds-hwang:
> Do we have an internal bump PR?

I just made it. 18796

markblee (approved):

PR #1004, opened 2025-02-18
Reverts sliding window attention changes.

markblee:


ds-hwang (approved):
Thank you for revert.

PR #1008, opened 2025-02-20
Partially reverts "gRPC Checkpointer (#1005)"

hanzhi713:
As discussed, gRPC Checkpointer is an experimental API. Let's first iterate this internally.


markblee (approved):
Let's also make sure internal bump passing before merge. Thanks!

PR #1016, opened 2025-02-25
Expose jax.lax.scan's unroll option to Repeat layer

changlan:


markblee (commented):
Thanks! Do we have an internal bump?

changlan:
Yes - 19082

markblee (approved):

PR #1017, opened 2025-02-25
Fix crash in log-mel frontend when waveform samples are integers.

ds-hwang:
After updating JAX, this existing hidden bug started causing CI failures. When the sample dtype is int32 (which is valid), `jnp.finfo` returns None, even though `jnp.iinfo` is available.
The previous JAX version seemed to handle this case more forgivingly.

```
../axlearn/axlearn/audio/frontend_utils.py:297: in linear_to_log_spectrogram
    return jnp.log(jnp.maximum(x, jnp.finfo(x.dtype).tiny))
```

### DIFF
--- a/axlearn/audio/frontend.py
+++ b/axlearn/audio/frontend.py
@@ -118,6 +118,19 @@ def _pre_emphasis(coeff: float) -> StageFn:
     return functools.partial(pre_emphasis, coeff=jnp.array(coeff))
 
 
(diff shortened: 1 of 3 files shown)


ds-hwang:
@markblee Could you take a look? From 1092

ruomingp (approved):

ds-hwang:
> test_dtype

Good point. Thank you. Done.

@ruomingp Could you approve it again?

markblee (approved):

PR #1018, opened 2025-02-25
Handle None bias in BiasAndResidual

ds-hwang:
Handle None bias in BiasAndResidual

### DIFF
--- a/axlearn/common/attention_bias.py
+++ b/axlearn/common/attention_bias.py
@@ -205,7 +205,9 @@ class BiasAndResidual(BaseAttentionBias, Generic[B]):
     residual: BaseAttentionBias
 
     def _value(self) -> Optional[Tensor]:
-        return CompositeAttentionBias([self.bias, self.residual]).value()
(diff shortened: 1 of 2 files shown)


ds-hwang:
@markblee could you review? From 974

markblee (approved):

axlearn/common/attention_bias.py
```diff
@@ -205,7 +205,10 @@ class BiasAndResidual(BaseAttentionBias, Generic[B]):
     residual: BaseAttentionBias
 
     def _value(self) -> Optional[Tensor]:
-        return CompositeAttentionBias([self.bias, self.residual]).value()
+        biases = [self.residual]
+        if self.bias is not None:
```
chunyang-wen, line 209:
```python
biases = [self.bias] if self.bias is not None else []
biases.append(self.residual)
```

ds-hwang: That's better. Thank you. Done.

markblee (approved):

PR #1020, opened 2025-02-25
Add priority_class as a launch flag

Ethanlm:


axlearn/cloud/gcp/jobset_utils.py
```diff
@@ -256,6 +256,14 @@ def define_flags(cls, fv: flags.FlagValues):
             "not all TPU types support this flag.",
             **common_kwargs,
         )
+        # Currently only supported in specific clusters with PriorityClass setup.
+        # TODO(ethanli): infer it from the JobMetadata.priority, and support it on all clusters.
```
markblee, line 260:
```suggestion
        # Only supported in clusters with PriorityClass setup.
        # TODO(ethanli): infer it from the JobMetadata.priority.
```

What's the behavior when specified and not supported? Is it silently ignored or raises an error?

markblee (approved):

axlearn/cloud/gcp/jobset_utils.py
Ethanlm: It will raise an error

Ethanlm:
squashed the commits

markblee (approved):

PR #1021, opened 2025-02-26
Emulate flash attentnion unittests on CPU.

ds-hwang:
utils.py codebase is not well covered by CI because it branches different backend.

This PR introduces new CPU test, utils_test.py.
This test is expected to run on CPU and is designed to validate GPU/TPU code from a CPU environment by fake mesh.
It allows quick verification in CI and local environments to ensure that code changes do not break GPU/TPU Flash Attention.


ds-hwang:
@ruomingp could you review it? From 1066

markblee (approved):

PR #1026, opened 2025-02-27
Flash Attention now explicitly checks whether it is in decoding mode.

ds-hwang:
Flash Attention now explicitly checks whether it is in decoding mode.

Currently, Flash Attention infers decoding implicitly based on circumstantial
evidence. This PR makes the check explicit.


ds-hwang:
@markblee Could you review? From 1084

markblee (approved):

PR #1027, opened 2025-02-28
[CI] Enable 4- and 8-device Flash Attention layers tests.

liboyue:
This PR enabled 4- and 8-device Flash Attention layers tests. These tests were actually skipped because there is only one device. Ideally we should use `pytest` marks to mark multi-device tests.


ruomingp (approved):

kelvinzou (changes requested):
I believe Boyue is filing this PR for google debugging purpose. CPU + simulated multi-device is having bug in pallas path and let's not check in this by mistake.

PR #1028, opened 2025-02-28
Refactor Mask in Attention

ds-hwang:
Currently, the attention code is **hardcoded** to handle either `causal_mask` or an arbitrary `mask_fn`.

To support **sliding window masks**, we previously used a **hack** by injecting the `_sliding_window_size` attribute into functions.

This refactor **makes the masking logic more flexible** by allowing arbitrary `MaskFnAttentionBias`.
If downstream requires a **new mask pattern**, they can simply:
  1. Implement a **subclass of `MaskFnAttentionBias`**.
  2. Set `attention.mask` accordingly.

### DIFF
--- a/axlearn/common/attention_bias.py
+++ b/axlearn/common/attention_bias.py
@@ -33,12 +33,13 @@ from typing import (
     final,
 )
 
+import einops
 import jax
 from jax import numpy as jnp
 from jax.sharding import PartitionSpec
 
 from axlearn.common import struct
-from axlearn.common.config import ConfigOr, maybe_instantiate
+from axlearn.common.config import ClassConfigBase, ConfigOr, config_for_class, maybe_instantiate
 from axlearn.common.utils import Tensor
 
 NEG_INF = -1e15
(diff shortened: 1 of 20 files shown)


axlearn/common/attention_bias.py
```diff
@@ -513,29 +519,15 @@ def _bool_value(self) -> Optional[Tensor]:
...
-            else:
-                raise NotImplementedError(f"Invalid value {target_positions.ndim=}.")
-            target_positions = target_positions[..., None]  # Shape: [batch, target_len, 1].
-
+        target_positions, source_positions = self.target_positions, self.source_positions
+        if target_positions.ndim != source_positions.ndim != 2:
+            raise ValueError(
+                f"{target_positions.shape=} or {source_positions.shape=} is not rank 2."
+            )
+        target_positions = einops.rearrange(target_positions, "b t -> b t 1")
```
changlan, line 529: fly-by comment: I wonder if we should do the same without introducing extra `einops` dependency.

ds-hwang: We could use the following code:
```
target_positions = target_positions[:, :, None]
```

However, I prefer using einops because it makes the meaning of shape transformations clearer in the code.

Since Axlearn and Ajax already have einops installed by default, we are increasingly using it.
Additionally, the G-model codebase also heavily relied on einops.

Would you prefer not to use einops?

changlan: I don't have a strong opinions - just from engineering perspective we might want to prefer native approach over 3rd party dependencies in case of unexpected surprises. But you can make the call if you think `einops` is stable enough.

markblee (approved):

axlearn/common/attention_bias.py
ds-hwang: Thank you for your concern.
I generally agree with your opinion. einops improves code readability, which I believe will contribute to maintaining a healthy codebase in the long run.
Also, as far as I remember, Google uses it in production, so I consider it to be a reliable thirdparty.

changlan: > Google uses it in production

fwiw I didn't see it in production codebase, but maybe "uses it in production" is a vague term after all.

ruomingp (commented):
LGTM. Thanks!

PR #1029, opened 2025-03-02
Supports arbitrary uniform partitioning in host-global array conversions.

markblee:


axlearn/common/utils.py
```diff
@@ -822,65 +837,93 @@ def make_gda(x, partition_spec):
...
 
-    return jax.tree.map(get_local_array, tree_paths(global_arrays), global_arrays)
+        Each key is a tuple of length `len(global_shape)`.
+        Each element of the tuple is a `(start, limit)` tuple, specifying the start and limit
+        indices of the shard along the global shape dim.
+        """
+        index_to_shard = []
+        for shard in shards:
+            index = tuple(
+                (s.start or 0, s.stop or global_shape[dim]) for dim, s in enumerate(shard.index)
```
ruomingp, line 874: Nit: maybe `slice` is more readable than `tuple`?

ruomingp (approved):

axlearn/common/utils.py
markblee: Unfortunately slices aren't hashable (useful for dedup).

PR #1030, opened 2025-03-03
Remove "einops==0.8.0" dependency in audio as core already has it.

ds-hwang:
Remove "einops==0.8.0" dependency in audio as core already has it.


ds-hwang:
@markblee could you take a look? From 1109

markblee (approved):

PR #1037, opened 2025-03-05
Enable to assign different parameters dtype during training

jialingt:
This PR enables `train_dtype` to accept both `jnp.dype` and also `PerParamFn[jnp.dtype]`:

1.  `jnp.dtype`, where both float inputs and model parameters will be cast to this dtype.

2. `ConfigOr[PerParamFn[jnp.dtype]]`, allowing different dtypes to be applied to different parameters during training.

### DIFF
--- a/axlearn/common/utils.py
+++ b/axlearn/common/utils.py
@@ -27,7 +27,17 @@ import traceback
 import types
 from collections.abc import Mapping, Sequence
 from enum import Enum
-from typing import Any, Callable, Literal, NamedTuple, Optional, Protocol, TypeVar, Union
+from typing import (
+    Any,
+    Callable,
+    Literal,
+    NamedTuple,
+    Optional,
+    Protocol,
+    TypeVar,
+    Union,
+    runtime_checkable,
+)
 
 import jax
 import numpy as np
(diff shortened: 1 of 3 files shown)


apghml (approved):

axlearn/common/utils.py
```diff
@@ -1072,6 +1091,150 @@ def cast(x: Union[Tensor, TensorSpec]) -> Union[Tensor, TensorSpec]:
...
 
 
+@runtime_checkable
+class PerParamFn(Protocol[T]):
+    """A callable that operates on each parameter."""
+
+    def __call__(self, params: Union[Nested[Tensor], Nested[TensorSpec]]) -> Nested[T]:
+        """This protocol requires a callable that accepts either a nested Tensor or
+        a nested TensorSpec as input and returns a processed value for each parameter.
+        Args:
```
markblee, line 1101: Looks like some docstring format got lost from copy/paste?

jialingt: Oh you are right! Good catch. 
Added all spaces back.

ds-hwang:
Hi, after this PR, CI is broken.

PR #1039, opened 2025-03-07
Implements TPU decoding in Pallas

hanzhi713:
See the docstring of `tpu_decoding.py`. This is to improve decoding performance before we have paged KV cache and paged kernel. 

Other changes:
1. The unit test of GPU FlashDecoding is moved and now tests both TPU and GPU decoding.
2. `mha_reference` now adopts GQA optimizations so it's more fair to compare the performance of custom decoding kernel and the reference implementation.
3. Fix `gpu_attention_benchmark.py` being broken for decoding after #962

### DIFF
--- a/axlearn/common/attention.py
+++ b/axlearn/common/attention.py
@@ -2091,6 +2091,44 @@ class MultiheadAttention(BaseLayer):
         return config_for_function(constant_scale_fn).set(value=1)
 
 
+def compute_gqa_logits(q_proj: Tensor, k_proj: Tensor) -> Tensor:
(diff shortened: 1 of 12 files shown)


ruomingp (approved):

hanzhi713:
@ruomingp Could you please re-approve? Thanks. A mistake was made when addressing internal comments. I've verified all TPU and GPU tests passed.

apghml (approved):

hanzhi713:
It looks like the merge check is failing with unrelated errors.

PR #1041, opened 2025-03-07
Refactor KV Cache Out of `QKVLinear`

ds-hwang:
Currently, **`QKVLinear` is overly complex** because it handles **both QKV computation and KV cache management**.

Although **QKVLinear’s role and KV cache strategy should be independent**, the current implementation **forces QKVLinear to manage KV cache**, making it necessary to **carefully maintain every QKVLinear subclass** (`FusedQKV`, `GroupedQKV`, `RoPEQKV`, etc.) to ensure they correctly handle KV cache.

**Key Changes in This PR**
This PR **removes KV cache logic from `QKVLinear`**, turning it into a **pure `forward`-only class** (such as QKV proj and RoPE) that **no longer needs to handle decoding**.

Instead, **Attention now owns the KV cache directly**, making it **more flexible for future KV cache strategies**.

Currently, **`QKVLinear` supports only one KV cache behavior**, which maintains a **fixed max length**. However, in the near future, we will introduce more **KV cache strategies**, such as:

- **Sliding Window Attention** → Requires a **sliding window KV cache**.
- **Sparse Attention** → Needs a KV cache that **dynamically selects sparse KV** (similar to DeepSeek). https://arxiv.org/abs/2502.11089

**Implementation Details**
A key aspect of this refactor is **how query positions and key positions are generated**.
Previously, the related logic was **scattered across multiple places**, but now, **positions are computed in a single place**:

- **Query positions** → Must be determined **before RoPE** since RoPE requires them. The **same query positions** are then **reused throughout the code**.
- **Key positions** → Only the **KV cache layer** can determine them **accurately** since **KV cache strategies** directly affect key positions. So, **KV cache is now responsible for generating key positions**.

In addition, **`KVState` now carries both KV values and key positions**.


axlearn/common/attention.py
```diff
@@ -168,10 +169,160 @@ class ForwardMode(enum.Enum):
...
+        The input k_proj/v_proj are generated by `i_proj.forward()` from the input tokens.
+        The output k_proj/v_proj will have the `kv_cache` concatenated and will be used
+        in the current attention operation.
+
+        Args:
+            cached_states: A `Nested[Tensor]` object containing KV cache such as key and value.
+            k_proj: [batch, source_length, num_kv_heads, per_head_dim], Projected key tensor.
+            v_proj: [batch, source_length, num_kv_heads, per_head_dim], Projected value tensor.
+            proj_positions: [batch, source_length], Positions of the k_proj, int32.
+            paddings: [batch, source_length], An optional paddings of the k_proj, k_proj dtype.
```
changlan, line 252: nit: Could you please add more details on the semantics of paddings?

axlearn/common/attention.py
```diff
@@ -168,10 +169,160 @@ class ForwardMode(enum.Enum):
...
+
+        The input k_proj/v_proj are generated by `i_proj.forward()` from the input tokens.
+        The output k_proj/v_proj will have the `kv_cache` concatenated and will be used
+        in the current attention operation.
+
+        Args:
+            cached_states: A `Nested[Tensor]` object containing KV cache such as key and value.
+            k_proj: [batch, source_length, num_kv_heads, per_head_dim], Projected key tensor.
+            v_proj: [batch, source_length, num_kv_heads, per_head_dim], Projected value tensor.
+            proj_positions: [batch, source_length], Positions of the k_proj, int32.
```
changlan, line 251: Is this same as `key_positions`? Why use different names?

axlearn/common/attention.py
```diff
@@ -2013,16 +2003,46 @@ def init_states(
...
+                kv_cache=self.kv_cache.init_states(shape=kv_shape, dtype=query.dtype),
+            )
+
+        if time_step is None:
+            # init_states without prefilling branch.
+            return init_states, None
+
+        # Prefill branch.
+        # TODO(dhwang2): Optimize it by passing only the valid parts of the query. Currently,
+        # prefill has a complexity of O(max_len²), but this can be easily reduced to
```
changlan, line 2031: Is this true when we use splash attention?

axlearn/common/attention.py
```diff
@@ -1758,6 +1734,7 @@ def _forward_for_mode(
             query: A Tensor or TensorSpec of shape [batch, target_length, target_dim].
             key:   An optional Tensor of shape [batch, source_length, source_dim].
             value: An optional Tensor of shape [batch, source_length, source_dim].
+            paddings: An optional Tensor of shape [batch, target_dim] for query paddings.
```
changlan, line 1737: nit: add more comments on the semantics of paddings?

axlearn/common/attention.py
ds-hwang: Done.

axlearn/common/attention.py
ds-hwang: it's changed to live_step_len and has long docstring.

axlearn/common/attention.py
ds-hwang: It's general explanation and it's not true on flash attention.

axlearn/common/attention.py
ds-hwang: it's changed to live_step_len and has long docstring.

ds-hwang:
@ruomingp @markblee could you review? From 1110

markblee (approved):

PR #1042, opened 2025-03-10
Fix KeyError of aot_model_analysis in GPU

ds-hwang:
On H100, XLA AOT analysis works well, but not all GPUs generate proper logs. Therefore, I wrapped the entire AOT log section with a try-exception block.

### DIFF
--- a/axlearn/common/trainer.py
+++ b/axlearn/common/trainer.py
@@ -1358,16 +1358,19 @@ def aot_model_analysis(compiled: jax.stages.Compiled) -> str:
                 + f"Code memory: {mb_or_gb(mem_stats.generated_code_size_in_bytes)}\n"
                 + f"Total HBM memory: {mb_or_gb(total_hbm)}\n"
             )
-        except AttributeError:
+        except AttributeError as e:
             # Some platforms may return different format.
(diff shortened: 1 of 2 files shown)


axlearn/common/trainer.py
```diff
@@ -1386,4 +1389,8 @@ def m_or_g(x, suffix=""):
             + f"  Load Balancing / Dispatch): {cost_stats.get('utilization8{}')}\n"
             + f"  Texture Units (or Rarely Used Compute Units): {cost_stats.get('utilization9{}')}"
         )
+    else:
+        # Some platforms may return different format unlike CPU, TPU (v5p) and GPU (H100).
+        analysis_results += f"{cost_stats}\n"
```
markblee, line 1394: nit -- Consider logging the warning anyway? Seems useful for users to know that this isn't the 'intended' output.

markblee (approved):

markblee (approved):

axlearn/common/trainer.py
ds-hwang: It has warning already in next line. :) Thank you for review!

PR #1043, opened 2025-03-11
Set megascale_grpc_enable_xor_tracer to false by default

Ethanlm:
megascale_grpc_enable_xor_tracer has caused two problems

* memory link on megascale coordinator (a bug before jax-0.5.0)
* extensive logs but doesn't seem to provide extra value to us

so disable it for now.


kelvin-zou (commented):
Do we have a good job that validated this change?

Ethanlm:
> Do we have a good job that validated this change?

Yes. Let me send you on slack

ruomingp (approved):

kelvin-zou (approved):
Thanks!

PR #1047, opened 2025-03-12
Generalizes checkpointer to support arbitrary PythonSavables.

markblee:


axlearn/common/checkpointer_orbax.py
```diff
@@ -126,7 +127,7 @@ def typestr(self) -> str:
 
         def _ckpt_dir(self, info: ocp.type_handlers.ParamInfo) -> str:
             # Each worker writes its grain checkpoints under a different path.
-            return os.path.join(info.parent_dir, f"grain_{jax.process_index()}")
+            return os.path.join(info.parent_dir, f"python_{jax.process_index()}")
```
ruomingp, line 130: This will break existing checkpoints. Is that OK?

ruomingp (approved):

axlearn/common/checkpointer_orbax.py
markblee: I had announced and didn't hear objections (and should be fairly easy to migrate the checkpoints). Happy to address follow-ups.

PR #1048, opened 2025-03-13
enable arbitrary head_dim when using gpu flash attention

Firenze11:


hanzhi713:
@Firenze11 Let's not merge before we know the performance is good.

ruomingp (approved):

Firenze11:
> @Firenze11 Let's not merge before we know the performance is good.

Performance is verified:
```
is_decode=False, use_bwd=False, bs=2, num_heads=32, num_kv_heads=None, seq_len=4096, sw_sz=-1
                                                  jax           axlearn       
per_head_dim=16                                   4.609120      0.538272      
per_head_dim=17                                   4.699008      0.792672
per_head_dim=32                                   4.659616      0.557440
per_head_dim=36                                   4.818976      0.719712
per_head_dim=63                                   4.890624      1.644256
per_head_dim=64                                   4.842208      0.691360
per_head_dim=72                                   4.973056      0.920128
per_head_dim=128                                  5.186784      0.895648
```
comparable to pre-PR records here https://github.com/apple/axlearn/blob/main/axlearn/common/flash_attention/gpu_attention_benchmark.py#L70-L75

Firenze11:
The tests have passed. @ruomingp @hanzhi713 could you take another look? Thanks!

hanzhi713 (approved):
Thanks. Please implement this for backward soon, otherwise this kernel will be broken for training if people use a non-power-of-2 head dim as the corresponding check is removed.

PR #1050, opened 2025-03-14
TPU flash attention: Allow custom mask.

ds-hwang:
TPU Flash Attention previously supported custom masks through splash_attention_mask.NumpyMask, but the following PR slightly disabled this functionality.
https://github.pie.apple.com/foundation-models/ajax/pull/19016

In production, only causal and sliding window masks are used, so TPU Flash Attention has been working without issues.

However, since there’s no need to remove support for something that was originally available, this PR re-enables custom mask support as before.

TEST=axlearn/common/flash_attention/tpu_attention_test.py on v5p


axlearn/common/flash_attention/tpu_attention.py
```diff
@@ -296,21 +296,21 @@ def check_tpu_splash_attention(
...
     if has_segment_ids:
         raise SplashAttentionUnsupportedError(
-            "The public API for SplashAttention that we "
-            "currently use does not support segment ids."
+            "The public API for SplashAttention that we currently use does not support segment ids."
+        )
+    if is_decoding:
+        # tpu_decoding.py covers this scenario.
+        raise SplashAttentionUnsupportedError(
+            "It does not support decoding because the query len must be a block size or bigger."
```
markblee, line 304:
```suggestion
            "Please use `tpu_decoding` for decoding."
```

markblee (approved):

axlearn/common/flash_attention/tpu_attention.py
ds-hwang: Thank you for suggestion. Done.

markblee (approved):

PR #1051, opened 2025-03-15
Fix Broken CPU Code Path in `run_aot_compilation.py`

ds-hwang:
`run_aot_compilation.py` was designed to run on **CPU** when `--topology=None`, but this **code path was broken**. This PR fixes it to ensure proper execution on **CPU**.

Now, --topology is a required flag.
If you want to perform AOT compilation using the JAX CPU library, you must specify --topology=cpu-256 (or a similar cpu-<digit> format). This is necessary because, even for CPU execution, we need to determine the logical device number to estimate activation memory usage.


ds-hwang:
@ruomingp Could you review it? From 1125

ruomingp (approved):

PR #1053, opened 2025-03-17
The dtype of paddings becomes jnp.bool.

ds-hwang:
Previously, it was assumed to have the same dtype as the input. However, since paddings only store 0/1 values, this restriction is unnecessary.

There is a need to save memory by using a much smaller dtype instead of float32/bfloat16.
We prefer to use jnp.bool because

Backend | Boolean Storage Size (jnp.bool)
-- | --
CPU (XLA:CPU) | 8 bits (1 byte per bool)
GPU (XLA:GPU, CUDA, ROCm) | 8 bits (1 byte per bool, no native bit-packing)
TPU (XLA:TPU) | 1 bit per bool (native bit-packing)

### DIFF
--- a/axlearn/common/convolution.py
+++ b/axlearn/common/convolution.py
@@ -445,11 +445,11 @@ class Conv1DWithPadding(Conv1D):
 
         Args:
             x: A Tensor of shape [batch_size, seq_len, frequency, input_dim].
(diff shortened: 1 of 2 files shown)


ds-hwang:
@ruomingp, @markblee Could you review it? From 1135

markblee (approved):

PR #1058, opened 2025-03-18
Refactor and optimize FlashAttention dispatch

hanzhi713:
The flash attention dispatch logic is convoluted and overly restrictive. This PR refactors kernel support check into separate backend files, and carefully re-examined the conditions in each kernel.

The following changes are made in this PR:
1. All attention kernels are refactored to separate classes having `is_supported` and `__call__` methods. This also unified the call signature of all kernels which makes it possible to greatly simplify unit tests and benchmarks.
2. The common checks for all attention kernels (e.g. checking if k and v shapes are equal) are lifted to the `BaseFlashAttention` class to reduce duplication.
3. Detailed logging is added to explain why some kernels aren't chosen for easy debug.
4. Some problems that exist within the previous attention kernel dispatch are solved:
    * Splash supports q seq != kv seq, but not enabled
    * Splash supports GQA optimization, but not enabled
    * Splash supports segment id, but not enabled
    * GPU Pallas Flash uses TPU block size for divisibility check. This is overly restrictive.
    * cuDNN supports any even sequence length, no block size restriction, but we use TPU block size for divisibility check.
    * cuDNN supports any head dim <= 128 and head_dim % 8 == 0, but we may still choose the Pallas kernel and error out.
    * No fallback to cuDNN when mask_fn or segments ids are used but seq_len and head_dim are not supported by Palls GPU kernel.
    * cuDNN (partially) supports sliding window, but not enabled.

### DIFF
--- a/axlearn/common/flash_attention/common.py
+++ b/axlearn/common/flash_attention/common.py
@@ -184,12 +184,14 @@ class BaseFlashAttention(Configurable):
     ) -> Tensor:
         """Computes attention context.
 
+        Warning: The dtype of key and value may differ from the dtype of query.
+
         Args:
             query: Query of shape [batch_size, target_length, num_heads, per_head_dim].
             key: Key of shape [batch_size, source_length, num_kv_heads, per_head_dim].
             value: Value of shape [batch_size, source_length, num_kv_heads, per_head_dim].
             bias: Attention bias to apply.
-            prng_key: PRNG key for dropout.
+            prng_key: PRNG key for dropout. Only needed when dropout_rate > 0.0.
 
         Returns:
             The context tensor of shape [batch_size, target_length, num_heads, per_head_dim].
--- a/axlearn/common/flash_attention/common.py
+++ b/axlearn/common/flash_attention/common.py
@@ -205,6 +205,12 @@ class BaseFlashAttention(Configurable):
 class BaseSingleStepDecoding(BaseFlashAttention):
     """Wraps the common checks for single step decoding kernels."""
 
+    @classmethod
+    def default_config(cls) -> BaseFlashAttention.Config:
+        cfg: BaseFlashAttention.Config = super().default_config()
+        cfg.is_decoding = True
+        return cfg
+
     def is_supported(
         self, *, query: Tensor, key: Tensor, value: Tensor, bias: BaseAttentionBias
     ) -> bool:
--- a/axlearn/common/flash_attention/common.py
+++ b/axlearn/common/flash_attention/common.py
@@ -2,7 +2,7 @@
 """Common utilities across backends."""
 
 from functools import partial
-from typing import NamedTuple, Optional
+from typing import Literal, NamedTuple, Optional
 
 import jax
 import jax.numpy as jnp
@@ -125,7 +125,17 @@ class BaseFlashAttention(Configurable):
         """Returns the class name."""
         return self.__class__.__name__
 
-    def _log_unsupported(self, reason: str):
+    def _log_unsupported(self, reason: str) -> Literal[False]:
+        """Logs this class is unsupported with `reason`.
+
+        The log message will be formatted as `Not using {self.name()} because {reason}`.
+
+        This method also conveniently returns False so it could be used like this in `is_supported`
+        ```
+        if ...:
+            return self._log_unsupported(...)
+        ```
+        """
         logging.warning("Not using %s because %s", self.name(), reason)
         return False
 
--- a/axlearn/common/flash_attention/common.py
+++ b/axlearn/common/flash_attention/common.py
@@ -138,7 +138,7 @@ class BaseFlashAttention(Configurable):
         return True
 
     def is_supported(
-        self, query: Tensor, key: Tensor, value: Tensor, bias: BaseAttentionBias
+        self, *, query: Tensor, key: Tensor, value: Tensor, bias: BaseAttentionBias
     ) -> bool:
         """Returns whether the attention kernel supports the given configuration.
 
@@ -175,6 +175,8 @@ class BaseFlashAttention(Configurable):
             )
         return True
 
+    # Note: Positional arguments are used since some use cases require positional-only args,
+    # such as functional transformations.
     def __call__(
         self,
         query: Tensor,
@@ -203,8 +205,11 @@ class BaseFlashAttention(Configurable):
 class BaseSingleStepDecoding(BaseFlashAttention):
     """Wraps the common checks for single step decoding kernels."""
 
-    def is_supported(self, query, key, value, bias):
-        if not super().is_supported(query, key, value, bias):
+    def is_supported(
+        self, *, query: Tensor, key: Tensor, value: Tensor, bias: BaseAttentionBias
+    ) -> bool:
+        """See `BaseFlashAttention.is_supported`."""
+        if not super().is_supported(query=query, key=key, value=value, bias=bias):
             return False
         if not self.cfg.is_decoding:
             return self._log_unsupported("is_decoding=False.")
--- a/axlearn/common/flash_attention/common.py
+++ b/axlearn/common/flash_attention/common.py
@@ -127,6 +127,7 @@ class BaseFlashAttention(Configurable):
 
     def _log_unsupported(self, reason: str):
         logging.warning("Not using %s because %s", self.name(), reason)
+        return False
 
     def _check_block_size(self, *, query: Tensor, key: Tensor, block_size: int) -> bool:
         q_seq_len = query.shape[1]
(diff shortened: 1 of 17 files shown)


axlearn/common/flash_attention/common.py
```diff
@@ -86,3 +91,178 @@ def query_iterator_indices(block_mask_map: np.ndarray, *, padding: int = 0) -> K
...
+        Attributes:
+            is_decoding: Whether we're in decoding/inference mode.
+            softmax_scale: Scale factor to apply to QK.
+            dropout_rate: Dropout rate for attention probs.
+            interpret: Whether to use interpret mode for Pallas kernels.
+            tpu_block_size: Block size for TPU pallas kernels.
+            gpu_block_size: Block size for GPU pallas kernels.
+        """
+
+        is_decoding: bool = False
```
markblee, line 112: Keep some of these as required? `is_decoding=False` doesn't seem like a universal default.

axlearn/common/flash_attention/common.py
```diff
@@ -86,3 +91,178 @@ def query_iterator_indices(block_mask_map: np.ndarray, *, padding: int = 0) -> K
...
+        # Keep a typed copy of self.config.
+        self.cfg: BaseFlashAttention.Config = self.config
+
+    def name(self) -> str:
+        """Returns the class name."""
+        return self.__class__.__name__
+
+    def _log_unsupported(self, reason: str):
+        logging.warning("Not using %s because %s", self.name(), reason)
+        return False
```
markblee, line 130:
```suggestion
```

axlearn/common/flash_attention/common.py
```diff
@@ -86,3 +91,178 @@ def query_iterator_indices(block_mask_map: np.ndarray, *, padding: int = 0) -> K
...
+    def _check_block_size(self, *, query: Tensor, key: Tensor, block_size: int) -> bool:
+        q_seq_len = query.shape[1]
+        k_seq_len = key.shape[1]
+        if q_seq_len % block_size != 0 or k_seq_len % block_size != 0:
+            self._log_unsupported(f"{q_seq_len=} or {k_seq_len=} is not divisible by {block_size=}")
+            return False
+        return True
+
+    def is_supported(
+        self, query: Tensor, key: Tensor, value: Tensor, bias: BaseAttentionBias
```
markblee, line 141:
```suggestion
        self, *, query: Tensor, key: Tensor, value: Tensor, bias: BaseAttentionBias
```

axlearn/common/flash_attention/common.py
```diff
@@ -86,3 +91,178 @@ def query_iterator_indices(block_mask_map: np.ndarray, *, padding: int = 0) -> K
...
+            )
+        if query.shape[2] % key.shape[2] != 0:
+            raise ValueError(
+                f"Expects query num heads {query.shape[2]} to be divisible by num key heads "
+                f"{key.shape[2]}"
+            )
+        return True
+
+    def __call__(
+        self,
```
markblee, line 179:
```suggestion
        self,
        *,
```
(If kwargs are prohibited somewhere, please leave a comment.)

axlearn/common/flash_attention/common.py
```diff
@@ -86,3 +91,178 @@ def query_iterator_indices(block_mask_map: np.ndarray, *, padding: int = 0) -> K
...
+            bias: Attention bias to apply.
+            prng_key: PRNG key for dropout. Only needed when dropout_rate > 0.0.
+
+        Returns:
+            The context tensor of shape [batch_size, target_length, num_heads, per_head_dim].
+        """
+        raise NotImplementedError()
+
+
+class BaseSingleStepDecoding(BaseFlashAttention):
```
markblee, line 203: Here we can define a `default_config` which sets `is_decoding=True`.

axlearn/common/flash_attention/common.py
```diff
@@ -86,3 +91,178 @@ def query_iterator_indices(block_mask_map: np.ndarray, *, padding: int = 0) -> K
...
+    """Wraps the common checks for single step decoding kernels."""
+
+    def is_supported(self, query, key, value, bias):
```
markblee, line 206: Retain typing/returns and add a docstring referring users to the parent.

(Please apply elsewhere.)

axlearn/common/flash_attention/gpu_attention.py
```diff
@@ -772,63 +789,168 @@ def call_kernel(
...
-        key: Key of shape [batch_size, source_length, num_heads, per_head_dim].
-        value: Value of shape [batch_size, source_length, num_heads, per_head_dim].
-        bias: Optional logit biases of shape [batch_size, num_heads, target_length, source_length].
-        mask: Optional logit mask of shape [batch_size, num_heads, target_length, source_length].
-        softmax_scale: Optional scale to apply to softmax. Defaults to 1.
-        seed: Random seed for dropout.
-        dropout_rate: Dropout rate.
-        qkv_layout: Layout string, with supported formats being BTNH, BNTH, BSNH,
-                    BNSH. Now it only supports BTNH.
+    _allow_explicit_bias = False
```
markblee, line 803: Why do we need the attribute?

axlearn/common/flash_attention/gpu_attention.py
```diff
@@ -772,63 +789,168 @@ def call_kernel(
...
+
+    def is_supported(self, query, key, value, bias):
+        if not super().is_supported(query, key, value, bias):
+            return False
+        if self.cfg.is_decoding:
+            if query.shape[1] > 1:
+                return self._log_unsupported("multi-step decoding is not supported.")
```
markblee, line 810: nit --
```suggestion
                self._log_unsupported("Multi-step decoding is not supported.")
                return False
```

axlearn/common/flash_attention/gpu_attention.py
```diff
@@ -772,63 +789,168 @@ def call_kernel(
...
+            if not key.shape[1] % 2 == 0:
+                return self._log_unsupported(f"key sequence length {key.shape[1]} is not even.")
+        else:
+            if not self._check_block_size(query=query, key=key, block_size=2):
```
markblee, line 814: Comment on block size in relation to gpu_block_size?

axlearn/common/flash_attention/layer.py
```diff
@@ -177,7 +177,7 @@ def _compute_attention(
             backend=backend,
             softmax_scale=1.0,
             is_decoding=is_decoding,
-            block_size=cfg.tpu_block_size,
+            tpu_block_size=cfg.tpu_block_size,
```
markblee, line 180: Can we leave a TODO to decouple the naming from specific platforms?

axlearn/common/flash_attention/neuron_attention.py
```diff
@@ -214,3 +218,37 @@ def _mha_backward(
...
 
 flash_attention.defvjp(_mha_forward, _mha_backward)
+
+
+class NeuronFlashAttention(BaseFlashAttention):
+    """Wraps the Neuron attention kernel."""
+
+    def is_supported(self, query, key, value, bias):
+        # TODO(hanzhi-zhou): neuron may error out for unsupported sequence length and head size.
+        # Should we add checks for them and fallback to XLA?
```
markblee, line 228: What is the reason not to do so?

markblee (commented):
Nice! This looks much cleaner.

axlearn/common/flash_attention/neuron_attention.py
hanzhi713: I don't know the exact kernel compatibility. I need to get my hand on some neuron devices to add proper checks. For now, this PR maintains the behavior of flash before this PR.

axlearn/common/flash_attention/common.py
hanzhi713: This return False is convenient since we can just use `return self._log_unsupported(...)` without the need to add `return False` after that.

axlearn/common/flash_attention/gpu_attention.py
hanzhi713: It's there to allow creating two classes, one allowing explicit bias, one disallowing it.

axlearn/common/flash_attention/gpu_attention.py
hanzhi713: It's in lowercase since the log message will be prepended with "Not using xxx because "

axlearn/common/flash_attention/common.py
hanzhi713: Left a comment

axlearn/common/flash_attention/common.py
hanzhi713: Done.

axlearn/common/flash_attention/common.py
hanzhi713: Done.

axlearn/common/flash_attention/gpu_attention.py
hanzhi713: Done.

axlearn/common/flash_attention/layer.py
hanzhi713: Done.

axlearn/common/flash_attention/common.py
hanzhi713: It's less verbose to have this set to False by default since majority of the use cases is decoding=False.

axlearn/common/flash_attention/common.py
hanzhi713: Done.

hanzhi713:
@markblee Could you please take a look again?

axlearn/common/flash_attention/common.py
markblee: It's slightly odd to rely on the return value of a logging statement. Should we at least add a docstring or return annotation so it's obvious that it's intentional?

markblee (approved):

markblee (approved):

PR #1061, opened 2025-03-19
TPU flash attention: Allow custom mask. 2/2

ds-hwang:
This is follow-up of https://github.com/apple/axlearn/pull/1050

`mask.bool_value()` internally uses `mask.target_positions` tensor, so it cannot be used inside `with jax.ensure_compile_time_eval():`, even if its shape is statically determined.

Before the changes in the previous PR, https://github.com/apple/axlearn/pull/1028 `mask.bool_value()` worked because mask did not store any tensors.

To resolve this, this PR reimplemented the same logic without using `jax.ensure_compile_time_eval()`.

As mentioned earlier, since the shape is statically determined, `target_positions` can be converted into a NumPy array at runtime.


ds-hwang:
@markblee Could you review? From 1145

ruomingp (approved):

PR #1066, opened 2025-03-24
Enabling Google Cloud Monitoring for Goodput and Step Deviation

Perseus14:
- Upgrade `ml-goodput-measurement` library to `v0.0.10` which now supports asynchronous upload of goodput, badput and step_deviation metrics to GCM
- Instrument several measurement events in `common/trainer.py` for goodput and badput calculation
- Modify `cloud/gcp/measurement.py` to accomodate for the new library changes specifically flags for pushing goodput and step deviation data to GCM.
- Added a guide for Goodput Monitoring.


docs/05-Goodput-Monitoring.md
```diff
@@ -0,0 +1,115 @@
...
+Goodput is the metric that measures the efficiency of model training jobs, i.e. productive time spent on training progress proportional to the total time spent by the workload. It is an actionable way for users to monitor where they can improve to get the most value from their accelerators.
+
+### What is Badput
+Badput is the metric that measures time that a workload spent on anything that is not productive training proportional to the total time spent by the workload. For example, the time spent in accelerator initialization, training preparation, program startup, data loading, portions of checkpointing, disruptions and wasted progress since the last checkpoint etc. all contribute to Badput.
+
+The ML Goodput Measurement library exposes Badput Breakdown. Further details of each bucket can be found [here](https://github.com/AI-Hypercomputer/ml-goodput-measurement?tab=readme-ov-file#badput-breakdown-details)
+
+### What is Step Time Deviation
+
+Step Time Deviation is the metric that measures deviation of step time from ideal step time.
```
jiya-zhang, line 16: Can we specify if this is in ms or seconds or other units?

axlearn/cloud/gcp/measurement.py
```diff
@@ -52,9 +68,28 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
...
+        - enable_gcp_goodput_metrics: Whether to push Goodput metrics to Google Cloud
+            Monitoring.
+        - enable_gcp_step_deviation_metrics: Whether to push step deviation metrics to Google
+            Cloud Monitoring.
         """
         cfg: measurement.Recorder.Config = cls.default_config()
-        cfg = maybe_set_config(cfg, **parse_kv_flags(fv.recorder_spec, delimiter="="))
+        kwargs = parse_kv_flags(fv.recorder_spec, delimiter="=")
+
+        boolean_flags = [
```
dipannita08, line 83: Should we add one to pick between interval monitoring and cumulative monitoring for GCP?

dipannita08 (commented):
LGTM, I'm good to loop in Ethan and team :)

axlearn/cloud/gcp/measurement.py
```diff
@@ -109,18 +144,35 @@ def start_monitoring(self, *args, **kwargs):
...
+                gcp_options = goodput_monitoring.GCPOptions(
+                    enable_gcp_goodput_metrics=cfg.enable_gcp_goodput_metrics,
+                    enable_gcp_step_deviation_metrics=cfg.enable_gcp_step_deviation_metrics,
+                )
+                self._monitor = goodput_monitoring.GoodputMonitor(
+                    job_name=cfg.name,
+                    logger_name=f"goodput_logger_{cfg.name}",
+                    tensorboard_dir=cfg.upload_dir,
+                    upload_interval=int(cfg.upload_interval),
+                    monitoring_enabled=(jax.process_index() == 0),
```
Ethanlm, line 165: Looks like this can just be `true`

axlearn/cloud/gcp/measurement.py
```diff
@@ -109,18 +144,35 @@ def start_monitoring(self, *args, **kwargs):
...
+        cfg: GoodputRecorder.Config = self.config
         if jax.process_index() == 0:
+            if self._monitor is None:
+                gcp_options = goodput_monitoring.GCPOptions(
+                    enable_gcp_goodput_metrics=cfg.enable_gcp_goodput_metrics,
+                    enable_gcp_step_deviation_metrics=cfg.enable_gcp_step_deviation_metrics,
+                )
+                self._monitor = goodput_monitoring.GoodputMonitor(
+                    job_name=cfg.name,
+                    logger_name=f"goodput_logger_{cfg.name}",
```
Ethanlm, line 162: not necessary a problem/blocker to this PR - but I think asking users to provide a unique name can be confusing. It would be better to construct a name that is unique by default

docs/05-Goodput-Monitoring.md
```diff
@@ -0,0 +1,115 @@
...
+To enable Goodput recording and monitoring on AXLearn, follow the example below.
+
+```bash
+   axlearn gcp gke start --instance_type=tpu-v5litepod-16 \
+        --bundler_type=artifactregistry --bundler_spec=image=tpu \
+        --bundler_spec=dockerfile=Dockerfile \
+        --name=<jobset-name> \
+        -- python3 -m ...training-config... \
+        --recorder_type=axlearn.cloud.gcp.measurement:goodput \
+        --recorder_spec=name=<jobset-name> \
```
Ethanlm, line 60: It might be worth to call out that a unique name should be used, otherwise the goodput value might not be correct

axlearn/cloud/gcp/measurement.py
```diff
@@ -37,10 +41,22 @@ class Config(measurement.Recorder.Config):
...
+              Google Cloud Monitoring.
+            enable_gcp_step_deviation_metrics: Whether to push step deviation
+              metrics to Google Cloud Monitoring.
         """
 
         upload_dir: Required[str] = REQUIRED
         upload_interval: Required[int] = REQUIRED
+        include_step_deviation: bool = True  # Default to True
+        step_deviation_interval_seconds: int = 30  # Default to 30 seconds
+        enable_gcp_goodput_metrics: bool = True  # Default to True
```
Ethanlm, line 58: Whether it is true or false, the goodput value will still show up on tensorboard, is it correct?

axlearn/common/trainer.py
```diff
@@ -853,7 +854,7 @@ def _prepare_training(self, prng_key: Tensor) -> bool:
 
         # Attempt to restore the latest checkpoint, which may contain a saved `_input_iter`.
         self.restore_checkpoint(restore_step=None)
-
+        self._maybe_record_event(measurement.Event.START_TRAINING_PREPARATION)
```
Ethanlm, line 857: Should this be moved to before `self.restore_checkpoint`?

axlearn/common/trainer.py
```diff
@@ -872,7 +873,7 @@ def _prepare_training(self, prng_key: Tensor) -> bool:
 
             with fs.open(os.path.join(cfg.dir, "model_analysis.txt"), "w") as f:
                 f.write(model_analysis)
-
+        self._maybe_record_event(measurement.Event.END_TRAINING_PREPARATION)
```
Ethanlm, line 876: Should this be moved to the bottom of this function?

axlearn/common/trainer.py
```diff
@@ -909,6 +910,7 @@ def restore_checkpoint(self, restore_step: Optional[int] = None) -> Optional[int
             restore_input_iter = cfg.save_input_iterator
             try:
                 # Try to restore with `input_iter`.
+                self._maybe_record_event(measurement.Event.START_DATA_LOADING)
```
Ethanlm, line 913: Does "DATA_LOADING" include restoring checkpoints? It looks a little bit confusing to me

axlearn/common/trainer.py
```diff
@@ -253,7 +253,7 @@ def __init__(
         self._per_param_train_dtype = maybe_instantiate(
             canonicalize_per_param_dtype(cfg.train_dtype)
         )
-
+        self._maybe_record_event(measurement.Event.START_ACCELERATOR_INIT)
```
Ethanlm, line 256: Is there a guideline on this event? I am not very sure where `START_ACCELERATOR_INIT` should be added

Ethanlm (commented):
Thanks for the work. Left some comments

axlearn/cloud/gcp/measurement.py
Ethanlm: Do we need to choose between interval monitoring and cumulative monitoring? Can it all be supported?

dipannita08: Cumulative Metrics at time T1 - Cumulative Metrics at time T0 is **not** the same as Interval Metrics between time [T0, T1), as a result these two sets of APIs provide different computations for the metrics. (I can highlight some examples to show why this is true if you are interested). 

Therefore depending on which set of APIs is used, the consumption of these metrics should be informed as such.

Each API provides a single metric for each type of Goodput or Badput, since we have **one** set of GCM metric to write to, the user code needs to make a determination of which API it is interested in.

My suggestion was to bootstrap both, with an option to pick the desired option with a single flag when running the workload.

Details on the underlying APIs should be documented in the README.

axlearn/cloud/gcp/measurement.py
dipannita08: We cannot construct a unique job name under the hood since, the library cannot differentiate between real disruptions (due to infrastructure, hardware or user errors) and manual disruptions (the user restarts the workload with same job name). If we force re-construct a unique job name when the workload restarts from a real disruption, we will lose valuable information from before the disruption.

Therefore the decision to re-use a previous job_name (I don't see the use-case for this, but if there is ever such a a use-case) should lie with the (informed) user.

Note that **when a manual disruption is needed due to errors in getting the workload started, a new unique job_name is not required**, since logs from the failed run are not uploaded, therefore do not affect future computations.

My suggestion:
Since it sounds like we always want to use a unique job_name, we can bail on Goodput Monitoring with an error message if a job_name (with existing logs) is accidentally re-used such that GoodputMonitoring does not proceed to upload potentially incorrect metrics.

docs/05-Goodput-Monitoring.md
dipannita08: This should already be in the README, however, we can also bail on GoodputMonitoring or log a warning if a previous name is re-used.

axlearn/cloud/gcp/measurement.py
dipannita08: Alternatively, logging a warning might be a better idea. Will leave it to you to make a decision.

axlearn/common/trainer.py
dipannita08: Its likely that restore_checkpoint might need to be wrapped around the checkpoint logging.

Since you have maximum context of underlying operations, if you could review & provide guidance on the location of each of the Recorder API calls for accelerator initialization, training preparation and data loading, that would be great.

axlearn/common/trainer.py
dipannita08: Data loading should include just that; the loading of each batch of data pre-training or during/overlapped with training.

axlearn/common/trainer.py
dipannita08: The design doc I shared with you previously should provide more insight on this. (Specifically, the *Badput Breakdown, Accelerator Initialization Time* section)

dipannita08:
> Thanks for the work. Left some comments

Thanks for your review, Ethan! Responded, PTAL :)

docs/05-Goodput-Monitoring.md
dipannita08: Actually, unique job_names and the consequence of re-using them on manual restarts is documented in the ml-goodput-measurement library README, we can bring that note over to this README or link to it.

axlearn/cloud/gcp/measurement.py
Perseus14: Done!.Thanks!

axlearn/cloud/gcp/measurement.py
Perseus14: Added details in the doc. Preferred approach is to use cumulative and apply PromQL logic over it for interval needs. This can be done on the GCM dashboard with a dynamic $time_span.

If the above approach doesn't provide the needed value, rather than reusing the same endpoint, we can create new metric endpoints on GCM and push interval data to it.

docs/05-Goodput-Monitoring.md
Perseus14: Highlighted in the doc that unique readable name is required

Perseus14:
Addressed the comments  and updated the `measurement.py` and `docs/05-Goodput-Monitoring.md`

axlearn/cloud/gcp/measurement.py
Ethanlm: I was actually thinking how do we enable this for jobs. Jobs can use the same name on different clusters. Writing Goodput to the same logger will cause incorrect result. 

On the other hand, we don't want to create a unique name every time the job auto-restarts. It should use the same name throughout the lifetime.

But also provide an option so users can choose to use the same logger in case of manual restart

axlearn/common/trainer.py
Ethanlm: checkpoint restore is part of training preparation. So I think it makes sense to include it.
We should just properly map "TRAINING_PREPARATION" to  "_prepare_training" function. 

This is also aligned with the design doc you shared earlier

> This is the time spent on the creation of checkpoint managers, checkpoint loading, running
mesh and model optimizers etc. 


Separately,  it would be nice to add "checkpoint" specific events in the near future.

But does Badput support nested structure? What happens if we record checkpoint_restore related event, which is part of "training_preparation", will calculation be messed up?

axlearn/common/trainer.py
Ethanlm: right, so should we remove it from "restore_checkpoint" here?

axlearn/common/trainer.py
Ethanlm: Thanks. It is probably just much easier to understand if we treat the whole "__init__" function as "_ACCELERATOR_INIT" event.

Imagine someone wants to add some more code in `__init__`. They will need to think if they should add it before `self._maybe_record_event(measurement.Event.START_ACCELERATOR_INIT)` or after.

Ethanlm:
cc @markblee @kelvin-zou to take a look as well

README.md
```diff
@@ -11,6 +11,7 @@
 | [Concepts](docs/02-concepts.md) | Core concepts and design principles. |
 | [CLI User Guide](docs/03-cli.md) | How to use the CLI. |
 | [Infrastructure](docs/04-infrastructure.md) | Core infrastructure components. |
+| [Goodput Monitoring](docs/05-Goodput-Monitoring.md) | How to monitor goodput. |
```
markblee, line 14: Thanks for adding docs! Can we link to it from `04-infrastructure.md` instead of here?

axlearn/cloud/gcp/measurement.py
```diff
@@ -37,10 +41,22 @@ class Config(measurement.Recorder.Config):
         Attributes:
             upload_dir: Directory to store metrics for the monitor.
             upload_interval: Time interval (seconds) for monitoring uploads.
+            include_step_deviation: Whether to include step deviation
+              metrics in the upload.
```
markblee, line 45: Clarify how 'step deviation' is defined and when one would want to disable it?

BTW, we use 100 line-length and 4 space tabs. Can we fix accordingly? Thanks!

axlearn/cloud/gcp/measurement.py
```diff
@@ -37,10 +41,22 @@ class Config(measurement.Recorder.Config):
...
+            step_deviation_interval_seconds: Time interval (seconds) for step
+              deviation metrics uploads.
+            enable_gcp_goodput_metrics: Whether to push Goodput metrics to
+              Google Cloud Monitoring.
+            enable_gcp_step_deviation_metrics: Whether to push step deviation
+              metrics to Google Cloud Monitoring.
```
markblee, line 51: Do we need separate flags for these two behaviors? What's the use-case for pushing one but not the other? 

In general, we try hard to avoid booleans in APIs: https://github.com/apple/axlearn/blob/main/docs/ml_api_style.md#avoid-booleans-in-apis

For example, in this case it's unclear what the interaction between these two flags is -- e.g., one would expect 'step deviation metrics' to be considered part of 'Goodput metrics', so what happens if you enable one but disable the other?

axlearn/cloud/gcp/measurement.py
```diff
@@ -52,9 +68,28 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
          - upload_dir: The directory to write Tensorboard data to.
          - upload_interval: The time interval in seconds at which to query and upload data
            to Tensorboard.
+         - include_step_deviation: Whether to include step deviation
+            metrics in the upload.
+        - step_deviation_interval_seconds: Time interval (seconds) for step
```
markblee, line 73: Instead of using two flags `include_step_deviation` and `step_deviation_interval_seconds`, one could infer that setting `step_deviation_interval_seconds = -1` means "don't upload to GCM".

axlearn/cloud/gcp/measurement.py
```diff
@@ -52,9 +68,28 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
...
+        kwargs = parse_kv_flags(fv.recorder_spec, delimiter="=")
+
+        boolean_flags = [
+            "include_step_deviation",
+            "enable_gcp_goodput_metrics",
+            "enable_gcp_step_deviation_metrics",
+        ]
+        for flag in boolean_flags:
+            if flag in kwargs:
+                kwargs[flag] = kwargs[flag].lower() == "true"
```
markblee, line 90: Let's try to avoid this change if possible (see above). An additional downside is that we now need to keep this list in sync w/ the config names and their types.

markblee (commented):
Thanks for the PR! Left some initial comments on measurement.

README.md
Perseus14: Sure! I have added a section in Infrastructure doc and linked it to the Goodput Monitoring doc. Let me know if that's what you were looking for.

axlearn/cloud/gcp/measurement.py
Perseus14: I have added some content on the Goodput Monitoring doc on how we are defining and calculating step deviation.

Also modfied to follow the 100 line-length and 4 space tabs rule!

axlearn/cloud/gcp/measurement.py
Perseus14: The goodput library by design separates out goodput and step deviation calculation. It spins up separate threads to calculate & report goodput and step deviation. Additionally it can push data to both tensorboard and GCM.

The current implementation offers AXLearn users the flexibility to either push data to TB or both TB and GCM. Additionally it also offers flexibility to push either goodput data or both goodput and step deviation data.

By default, we calculate and push both goodput and step deviation data to TB and GCM. The flags make it so that users have choice to disable any of them.

axlearn/cloud/gcp/measurement.py
Perseus14: Done!

Removed the boolean flag `include_step_deviation`. Values less than or equal to 0 for `step_deviation_interval_seconds` will be considered as 'don't calculate step_deviation and don't push to TB and GCM'.

Also converted the boolean flags `enable_gcp_goodput_metrics` and `enable_gcp_step_deviation_metrics` to int. Value set to 1 indicates that it's enabled, any other value indicates that it's disabled.

The current implementation gives users two types of choices.

- Whether to send it to Tensorboard only or both Tensorboard and GCM.
- Whether to send goodput metric only or both goodput and step deviation.

Let me know if these are fine or should be simplified further.

axlearn/cloud/gcp/measurement.py
Perseus14: Done!

axlearn/cloud/gcp/measurement.py
```diff
@@ -4,15 +4,19 @@
...
         --bundler_type=artifactregistry --bundler_spec=image=tpu \
         --bundler_spec=dockerfile=Dockerfile \
         -- python3 -m my_training_job \
         --recorder_type=axlearn.cloud.gcp.measurement:goodput \
         --recorder_spec=name=my-run-with-goodput \
         --recorder_spec=upload_dir=my-output-directory/summaries \
-        --recorder_spec=upload_interval=30
+        --recorder_spec=upload_interval=30 \
+        --recorder_spec=include_step_deviation=True \
+        --recorder_spec=step_deviation_interval_seconds=30 \
```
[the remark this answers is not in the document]
Perseus14: We have separate threads for step_deviation and goodput so we can have separate upload intervals for both.

If there is no scenario where the intervals need to be different, we can modify to have step_deviation interval to use the uplod_interval value

axlearn/common/trainer.py
Perseus14: Done

axlearn/cloud/gcp/measurement.py
Perseus14: I stand corrected. Please ignore my previous suggestion. 

For the moment, we do not support interval query. We defined GCM system metric (Productive and Non-Productive Time) for cumulative data, we cannot use it for interval data.

Will need new GCM system metrics to support interval query, so hopefully this can be supported in the coming iterations of the library.

axlearn/cloud/gcp/measurement.py
Ethanlm: > Also converted the boolean flags enable_gcp_goodput_metrics and enable_gcp_step_deviation_metrics to int. Value set to 1 indicates that it's enabled, any other value indicates that it's disabled.

Sorry I know you changed from true/false to 1/others for `enable_gcp_step_deviation_metrics`, but I personally think using True/False for enable_gcp_step_deviation_metrics is easier to read. 

I understand @markblee had some comments about true/false. 
cc @markblee , do we prefer int 1/others over True/False?

axlearn/cloud/gcp/measurement.py
markblee: Thanks, how about this question?

> What's the use-case for pushing one but not the other?

Is it just a matter of cost?

axlearn/cloud/gcp/measurement.py
markblee: Maybe there was some miscommunication -- we shouldn't need `enable_gcp_step_deviation_metrics` anymore since it can be inferred from `step_deviation_interval_seconds`. 

I also wonder why we would want to disable one but not the other and left a question here: https://github.com/apple/axlearn/pull/1066#discussion_r2035785864

Is there a specific reason to disable it on a per-job basis?

Perseus14: Idea was to make it flexible so that users can have the choice to push to TB and _not_ to GCM. If that's not a concern, I can remove that flag.

Currently I expect most of the jobs to use the following command, this would push data to TB and GCM

`python3 -m my_training_job \
        --recorder_type=axlearn.cloud.gcp.measurement:goodput \
        --recorder_spec=name=my-run-with-goodput \
        --recorder_spec=upload_dir=my-output-directory/summaries \
        --recorder_spec=upload_interval=30 \
        --recorder_spec=step_deviation_interval_seconds=30`
        
In cases where user wants to push to tensorboard  but not to GCM, the following command can be used       

`python3 -m my_training_job \
        --recorder_type=axlearn.cloud.gcp.measurement:goodput \
        --recorder_spec=name=my-run-with-goodput \
        --recorder_spec=upload_dir=my-output-directory/summaries \
        --recorder_spec=upload_interval=30 \
        --recorder_spec=step_deviation_interval_seconds=30 \
        --recorder_spec=enable_gcp_goodput_metrics=0 \
        --recorder_spec=enable_gcp_step_deviation_metrics=0`
        
If these scenarios are not needed, I can remove the flags.

markblee: Perhaps we can introduce flags later as needed? It might be better to keep the initial APIs simple. WDYT?

axlearn/cloud/gcp/measurement.py
```diff
@@ -37,10 +40,20 @@ class Config(measurement.Recorder.Config):
         Attributes:
             upload_dir: Directory to store metrics for the monitor.
             upload_interval: Time interval (seconds) for monitoring uploads.
+            step_deviation_interval_seconds: Time interval (seconds) for step
+              deviation metrics uploads. -1 to disable step deviation uploads.
```
markblee, line 44:
```suggestion
                deviation metrics uploads. -1 to disable step deviation uploads.
```
nit -- we use 4-space tabs.

axlearn/cloud/gcp/measurement.py
```diff
@@ -37,10 +40,20 @@ class Config(measurement.Recorder.Config):
...
+              Google Cloud Monitoring. 1 to enable and anything else to disable.
+            enable_gcp_step_deviation_metrics: Whether to push step deviation
+              metrics to Google Cloud Monitoring. 1 to enable and anything
+              else to disable.
         """
 
         upload_dir: Required[str] = REQUIRED
         upload_interval: Required[int] = REQUIRED
+        step_deviation_interval_seconds: int = 30  # Default to 30 seconds
+        enable_gcp_goodput_metrics: int = 1  # Default to True
```
markblee, line 55: As mentioned in the other comment, replacing bools with 1/else still conceptually results in the same behavior as True/False. Maybe we can remove these for now and add them later as needed.

axlearn/common/trainer.py
```diff
@@ -240,6 +240,7 @@ def __init__(
         self._device_monitor = maybe_instantiate(cfg.device_monitor)
         self._recorder = maybe_instantiate(cfg.recorder)
         self._is_initialized: bool = False
+        self._maybe_record_event(measurement.Event.START_ACCELERATOR_INIT)
```
markblee, line 243: Is accelerator init intended to include jax.distributed.init time? FWIW, that happens in `launch.setup`.

axlearn/common/trainer.py
```diff
@@ -567,8 +568,10 @@ def run(
             )
 
             # Prepare training.
+            self._maybe_record_event(measurement.Event.START_TRAINING_PREPARATION)
             if not self._prepare_training(prng_key):
                 return None
+            self._maybe_record_event(measurement.Event.END_TRAINING_PREPARATION)
```
markblee, line 574: Is it problematic for this to be unpaired if we return in L573?

axlearn/common/trainer.py
```diff
@@ -853,7 +869,6 @@ def _prepare_training(self, prng_key: Tensor) -> bool:
 
         # Attempt to restore the latest checkpoint, which may contain a saved `_input_iter`.
         self.restore_checkpoint(restore_step=None)
-
```
markblee, line 856: nit -- are these formatting changes relevant?

docs/05-Goodput-Monitoring.md
```diff
@@ -0,0 +1,163 @@
...
+Please use a unique workload name, unless you intend to monitor cumulative
+Goodput/Badput metrics of a previous workload along with your current workload.
+
+### How to Monitor Goodput and Badput
+
+To enable Goodput recording and monitoring on AXLearn, follow the example below.
+
+
+```bash
+    axlearn gcp gke start --instance_type=tpu-v5litepod-16 \
```
markblee, line 86: The commands here have recently migrated:
```suggestion
    axlearn gcp launch run --instance_type=tpu-v5litepod-16 \
```

axlearn/cloud/gcp/measurement.py
Perseus14: Done!

axlearn/cloud/gcp/measurement.py
Perseus14: Removed `enable_gcp_goodput_metrics` and `enable_gcp_step_deviation_metrics` flags. Default value is True for these variables.

axlearn/common/trainer.py
Perseus14: IIUC, in the current implementation we use `jax.process_index()` to identify the coordinator process (process_id=0) for recording logs to avoid duplicates. This requires jax to be initialized beforehand.

@dipannita08 Thoughts?

axlearn/common/trainer.py
Perseus14: Moved the events to inside the function to avoid unpaired logs

axlearn/common/trainer.py
Perseus14: My bad! Removed them.

docs/05-Goodput-Monitoring.md
Perseus14: Done!

axlearn/common/trainer.py
dipannita08: Yes, JAX initialization is required before making the Goodput record calls. ACC_INIT is meant to capture device init, slice building, device reinit/reset, security initializations, initialization of pre-mapped buffers etc.

axlearn/common/trainer.py
dipannita08: Application layer does not need to worry about unpaired record calls if it exits early, the library will take care of this under the hood. However, if we can determine more accurate start and end times for the event, then let's move the events to envelope them as tightly as possible.

axlearn/cloud/gcp/measurement.py
dipannita08: I'm not convinced that we cannot use the existing system metrics to write interval data. As long as the user knows which API was used, the metrics should be interpreted accordingly. Echoing my previous comment: "depending on which set of APIs is used, the consumption of these metrics should be informed as such."

The library support computing, storing and exporting Interval data - but we only have one set of system metrics to write to, so I think it is reasonable to let the user decide which type of metrics to export.

axlearn/cloud/gcp/measurement.py
dipannita08: >On the other hand, we don't want to create a unique name every time the job auto-restarts. It should use the same name throughout the lifetime.

>But also provide an option so users can choose to use the same logger in case of manual restart

Isn't this how it stands today? New unique names *are not* created on auto-restarts - job uses the same name throughout its lifetime, and user *can* choose a different or same logger name when they manually restart through the `name` flag.

dipannita08 (commented):
Thanks for the updates, left a couple more comments.

axlearn/cloud/gcp/measurement.py
```diff
@@ -109,18 +115,42 @@ def start_monitoring(self, *args, **kwargs):
...
-                tensorboard_dir=cfg.upload_dir,
-                upload_interval=int(cfg.upload_interval),
-                monitoring_enabled=(jax.process_index() == 0),
-                include_badput_breakdown=True,
-            )
 
+        Default behavior is to push metrics to Google Cloud Monitoring.
+        This behavior can be overridden by setting following the variables to False:
+        - enable_gcp_goodput_metrics=False
+        - enable_gcp_step_deviation_metrics=False
```
markblee, line 122:
```suggestion
        This behavior can be overridden by configuring `goodput_monitoring.GCPOptions`.
```

axlearn/common/trainer.py
markblee: Hm, I see. I wonder whether it makes sense to decouple the goodput metrics library from jax.

axlearn/cloud/gcp/measurement.py
```diff
@@ -109,18 +115,42 @@ def start_monitoring(self, *args, **kwargs):
...
+        """
+        cfg: GoodputRecorder.Config = self.config
+        enable_gcp_goodput_metrics = True
+        enable_gcp_step_deviation_metrics = True
+        include_step_deviation = True
```
markblee, line 127: Do we need all of these variables? For variables that are only used once, we can just inline them, as it's more readable.

markblee (approved):
Thanks, mostly LGTM.

axlearn/cloud/gcp/measurement.py
Perseus14: Makes sense!

axlearn/cloud/gcp/measurement.py
Perseus14: Done!

axlearn/common/trainer.py
dipannita08: It is a very good point. We don't have a better way to filter out a single worker in a pod from the application layer at the moment,  but we do want to eventually come up with a more generalized solution for this for future versions of the package.

dipannita08 (commented):
Quick FYI, we have another update to the package coming: to measure Badput from synchronous and training overlapped events such as "Eval Step" and "SDC Compilation". Might be easier to integrate the latest in this PR as well - should just be a couple extra lines of code to integrate!

cc @Ethanlm @markblee @ruomingp

axlearn/cloud/gcp/measurement.py
dipannita08: Since we have an increasing requirement to monitor interval data instead of cumulative, can we update the default to use Interval APIs instead?

Let me know if you have any concerns @Perseus14 .

Ethanlm:
> Quick FYI, we have another update to the package coming: to measure Badput from synchronous and training overlapped events such as "Eval Step" and "SDC Compilation". Might be easier to integrate the latest in this PR as well - should just be a couple extra lines of code to integrate!
> 
> cc @Ethanlm @markblee @ruomingp

We can also merge this first if that makes next PR's review easier

Ethanlm (approved):

findmyway (approved):

findmyway (approved):
Tested it internally and works well.

findmyway:
I guess the CI should pass if you sync the latest main branch.

Perseus14:
Does this need another approval to trigger the workflows? @findmyway

findmyway:
> Does this need another approval to trigger the workflows? @findmyway

I manually triggered it. All passed now ;)

findmyway:
@Perseus14 By the way, do you know the release strategy of `ml-goodput-measurement`? I notice that `0.0.9` is published to pypi, but it is not reflected in the github repo. It would help us bump the version more confidently if we could see the differences across versions easily at this stage.

axlearn/common/trainer.py
```diff
@@ -1026,6 +1042,9 @@ def _get_compiled_train_step_fn(
             )
             return self._compiled_train_step
         logging.log_first_n(logging.INFO, "Compiling XSC train step.", 1)
+        self._maybe_record_event(
+            measurement.Event.START_CUSTOM_BADPUT_EVENT, custom_badput_event_type="SDC_CHECK"
+        )
```
dipannita08, line 1047: We should check if the [xsc_policy_check](https://github.com/apple/axlearn/blob/121f43206facb2c9a9a524dbf5774c7d07bfdffb/axlearn/common/trainer.py#L170) is enabled before recording this. Perhaps inside the compile function would be best.

Ethan confirmed that it is not enabled by default.

Thanks for the pointers @Ethanlm!

Ethanlm: @dipannita08 Would it be easier to revert the recent change, get this PR merged first, and make a separate PR for SDC related change?

dipannita08: It would be easier (overall faster) to get the bigger change and the SDC + Eval Step support merged in together.

But if we are looking to restart the job over the weekend, then I'm good to revert the latest changes, merge this PR and create a new one for the SDC + Eval changes!

Ethanlm: That's fine. We are not going to add this to the job. It is going to finish pretty soon. I asked mainly for easier reviews. Thanks

dipannita08: Sounds good!

Perseus14:
Reverted to previous approved changes.

Upgraded `ml-goodput-measurement` library to v0.0.10 as it fixes few bugs

@Ethanlm @findmyway @markblee

findmyway (approved):

dipannita08 (approved):
Thank you! 

@markblee, @Ethanlm would you be able to PTAL and add this to the merge queue?

Ethanlm (approved):

Ethanlm:
Since @markblee approved the change before, I went ahead and clicked the merge

PR #1070, opened 2025-03-25
Changed job submitter flags

muyangyuapple:


markblee (approved):

markblee (approved):

axlearn/cloud/gcp/job_flink.py
```diff
@@ -390,8 +391,9 @@ def _build_job_submission_deployment(
             cfg.builder.accelerator.num_replicas * system.vms_per_slice * system.chips_per_vm
         )
         user_command += (
-            f" --flink_master_address={job_manager_ip}"
-            f" --flink_parallelism={flink_parallelism}"
+            f" --flink_master={job_manager_ip}"
+            f" --parallelism={flink_parallelism}"
```
markblee, line 395: What's the motivation of changing `--flink_parallelism` to `--parallelism`? Is there testing for this change?

muyangyuapple: Beam's PipelineOption has an constructor to take in k,v value in a list format. People usually do this:
```
args, pipeline_args = argparse.parse_known_args()

pipeline_option = PipelineOption(pipeline_args)
```

But the flag key has to match exactly what's declare in FlinkRunner. So I change the flag names to match what's declare there.

This line in unit test covers these three flags together: https://github.com/apple/axlearn/pull/1070/files#diff-8e07cb3662b05c6fca899373d2f52d1e28fbb97ea00a1d670db1086e326d734cR354

And I also launched a job to verify it e2e.

muyangyuapple: By doing this, users can avoid setting the options one by one.

For future reference, FlinkRunner's flag names are here: https://beam.apache.org/documentation/runners/flink/

markblee: How was it working before? Was the old flag being ignored?

markblee (approved):

axlearn/cloud/gcp/job_flink.py
muyangyuapple: It was used by the keyworded argument constructor like this:
https://github.pie.apple.com/foundation-models/ajax/blob/master/ajax/tools/beam_inference/utils.py#L875

muyangyuapple: It used to required users to declare those flags explicitly and pass the flag values to this util function to create a pipeline option.

After this change, it's up to users to chose this util function or the constructor that takes in a list w/o declaring those flags if they use argparser. It's more a convenience improvement.

PR #1071, opened 2025-03-26
KV cache update respects cached_dtype

ds-hwang:
The newly added dynamic_update_slice_in_dim path was ignoring cached_dtype. This fixes that issue.


ds-hwang:
@markblee Could you review? From 1170

markblee (approved):

PR #1076, opened 2025-03-28
Make `attention_logit_biases` in init_states/extend_step Optional

ds-hwang:
This is legacy. Make it optional. Use cfg.mask to avoid unnecessary O(T^2) biases. Currently, it's used by only T5 model.


ds-hwang:
@markblee Could you review? From 1166

markblee (approved):

PR #1077, opened 2025-03-28
Flash attention fallbacks to standard attention, not ReferenceMHA.

ds-hwang:
`ReferenceMHA` is essentially a duplication of standard attention’s `_compute_attention`.
Instead of duplication, we now call `_compute_attention` directly.

Another drawback of `ReferenceMHA` is that it **doesn’t apply gradient checkpointing**, which can lead to **significantly higher memory usage**. For example, training a **600M Conformer model** downstream can cause memory usage to jump from **24.89 GB to 157.58 GB**.


ds-hwang:
@markblee Could you review? From 1172

markblee (approved):

PR #1081, opened 2025-03-31
Splash Attention creates a NumPy mask inside jax.ensure_compile_time_eval()

ds-hwang:
This is because mask.mask can use jnp ops (e.g., jnp.logical_and). If that happens during tracing, it produces tracer objects, which can’t be converted to NumPy arrays.

To fix this, we restore the use of jax.ensure_compile_time_eval() around the custom mask creation.


ds-hwang:
@markblee Could you review? From 1161

markblee (approved):

PR #1083, opened 2025-04-01
Remove assert_equal_shape((k_proj, v_proj)) in attention.

ds-hwang:
They don't need to be same shape. e.g. MLA


ds-hwang:
@markblee Could you review? From 1181

markblee (approved):

PR #1092, opened 2025-04-02
Fix GPU decoding dtype and flash tests

hanzhi713:
Similar fix as in https://github.com/apple/axlearn/pull/1049 for mixed qkv dtypes but for GPU decoding kernel.

Also fixed flash attention tests problems introduced in https://github.com/apple/axlearn/pull/1081


markblee (approved):
Thanks!

PR #1093, opened 2025-04-03
Clarify that the k_proj/v_proj's dtype maybe changed by the kv_cache.

ds-hwang:
This PR introduces no logic change, since low-precision floats are automatically upcast when operated on with higher-precision floats. The only difference is that the upcasting is now made explicit in the code.


ds-hwang:
@ruomingp Could you review? From 1192

ruomingp (approved):

PR #1094, opened 2025-04-03
Add 7B sigmoid

WeersProductions:
- Adds 7B Sigmoid configs + baselines.


axlearn/experiments/text/gpt/common.py
```diff
@@ -254,9 +254,9 @@ def model_config(
         z_loss_scale: The scalar weight for the z-loss to encourages the cross-entropy loss
             normalizer to be well-behaved.
         ffn_structure: The inner structure of the feedforward layer.
-            Options: [prenorm, postnorm, hybridnorm].
+            Options: See `SupportedNormStructure`.
```
markblee, line 257: Consider updating the typing appropriately and move the definitions to common.

markblee (approved):

PR #620, opened 2024-08-02
remove nodeSelector provisioner-nodepool

samos123:
fix #619

PR #1096, opened 2025-04-04
fix nodeSelectors when not using bastion

### LINKED ISSUE #621: Unable to use TPU on GKE using on-demand quota

samos123:
Removes the spot and provisioner-node-pool-id selector when bastion is not in-use. This allows submitting Jobset on both on-demand, reserved and spot capacity and improves UX for non-Bastion use cases.

Fixes #621

This approach is preferred over #622 and #620

### DIFF
--- a/axlearn/cloud/gcp/jobs/gke_runner.py
+++ b/axlearn/cloud/gcp/jobs/gke_runner.py
@@ -278,7 +278,7 @@ class GKERunnerJob(GCPJob):
                 **custom_jobset_kwargs(),
             )
 
-            tier = os.environ.get("BASTION_TIER", 0)
+            tier = os.environ.get("BASTION_TIER", None)
             reservation = _infer_reservation(resp["spec"])
             if runner_utils.should_recreate_job(tier, reservation):
                 return GKERunnerJob.Status.RESCHEDULED
@@ -396,7 +396,7 @@ class GKERunnerJob(GCPJob):
             node_pool_config = node_pool.get("config", {})
             reservation_affinity = node_pool_config.get("reservationAffinity", {})
             taints = node_pool_config.get("taints", [])
-            tier = os.environ.get("BASTION_TIER", 0)
+            tier = os.environ.get("BASTION_TIER", None)
             has_reservation = (
                 reservation_affinity.get("key") == "compute.googleapis.com/reservation-name"
                 and len(reservation_affinity.get("values", [])) > 0
--- a/axlearn/cloud/gcp/jobset_utils.py
+++ b/axlearn/cloud/gcp/jobset_utils.py
@@ -570,7 +570,7 @@ class TPUReplicatedJob(SingleReplicatedJob):
             if cfg.reservation_project is not None:
                 selector.update({"cloud.google.com/reservation-project": cfg.reservation_project})
             labels.update({"bastion-tier": "reserved"})
-        else:
+        elif tier is not None:
             logging.info("Found tier=%s in env. Using spot quota", tier)
             selector.update({"cloud.google.com/gke-spot": "true"})
             tolerations.append(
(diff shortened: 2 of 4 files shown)


axlearn/cloud/gcp/jobs/gke_runner.py
```diff
@@ -278,7 +278,7 @@ def _get_status(self) -> Status:
                 **custom_jobset_kwargs(),
             )
 
-            tier = os.environ.get("BASTION_TIER", 0)
+            tier = os.environ.get("BASTION_TIER", None)
```
markblee, line 281: Instead of silently disabling it if it's not present, maybe it's safer to allow users to opt-out by setting it to something like "none". This can avoid accidentally disabling it on the bastion codepaths.

axlearn/cloud/gcp/jobset_utils.py
```diff
@@ -570,7 +570,7 @@ def _build_pod(self) -> Nested[Any]:
             if cfg.reservation_project is not None:
                 selector.update({"cloud.google.com/reservation-project": cfg.reservation_project})
             labels.update({"bastion-tier": "reserved"})
-        else:
+        elif tier is not None:
```
markblee, line 573:
```suggestion
        elif tier != "none":
```

samos123:
Tested by using this:
```
#!/usr/bin/env bash

set -xe

export RANDOM_CHARS=$(LC_CTYPE=C openssl rand -base64 12 | tr -dc 'a-z0-9' | head -c 3 ; echo)
export CLUSTER=${CLUSTER:-$USER-axlearn2}
export NAME=$USER-$RANDOM_CHARS
export BASTION_TIER=disabled
export DEFAULT_PROJECT_ID=$(gcloud config get project)
export PROJECT_ID=${PROJECT_ID:-$DEFAULT_PROJECT_ID}

axlearn gcp bundle --name=$NAME \
        --bundler_spec=allow_dirty=True \
        --bundler_spec=dockerfile=Dockerfile --bundler_spec=target=tpu

axlearn gcp gke start --cluster=$CLUSTER --name=$NAME \
        --instance_type=tpu-v6e-16 \
        --num_replicas=1 \
        --bundler_spec=allow_dirty=True \
        --bundler_spec=dockerfile=Dockerfile --bundler_spec=target=tpu \
        -- python3 -m axlearn.common.launch_trainer_main \
        --module=text.gpt.c4_trainer --config=fuji-7B-v2-flash-single-host \
          --trainer_dir=gs://$PROJECT_ID-axlearn/$USER-v6e-7b-1/ \
          --data_dir=gs://axlearn-public/tensorflow_datasets  \
          --jax_backend=tpu \
          --mesh_selector=tpu-v6e-16 \
          --trace_at_steps=3
```

samos123:
I will add a test case to explicitly test for nodeSelectors being gone.

axlearn/cloud/gcp/jobs/gke_runner.py
samos123: fixed, but changed to "disabled" instead of "none" since it's more explicit in what it does and not to be confused with None python value.

axlearn/cloud/gcp/jobset_utils.py
samos123: fixed but used "disabled" instead.

samos123:
Thanks Mark for your feedback and approving. Do we need approval from @ruomingp as well?

markblee:
> Thanks Mark for your feedback and approving. Do we need approval from @ruomingp as well?

Will do a quick test internally and merge soon.

PR #1097, opened 2025-04-04
Add GKE A3 Mega, A3 Ultra, and A4 High instance support

andersensam:
This PR adds support for A3 Mega, A3 Ultra, and A4 High (B200) on GKE.


Dockerfile
```diff
@@ -108,4 +108,4 @@ COPY . .
 # Final target spec.                                                           #
 ################################################################################
 
-FROM ${TARGET} AS final
+FROM ${TARGET} AS final
```
samos123, line 111: probably better to revert this since it's not changing anything

axlearn/common/flash_attention/gpu_attention.py
```diff
@@ -66,6 +66,13 @@
 from axlearn.common.layers import get_dropout_mask
 from axlearn.common.utils import Tensor
 
+### A4 platform specific code ###
```
samos123, line 69: Is this only temporary until triton has a fix? Maybe we can add a link to the upstream bug here with a TODO to remove once upstream bug is fixed?

pyproject.toml
```diff
@@ -197,4 +197,4 @@ junit_family="xunit2"
 
 [tool.isort]
 line_length = 100
-profile = "black"
+profile = "black"
```
samos123, line 200: revert

axlearn/cloud/gcp/jobset_utils.py
```diff
@@ -898,18 +986,258 @@ def _build_a3_init_container(self) -> Nested[Any]:
...
+            },
+        ]
+
+        command = [
+            "bash",
+            "-c",
+            "set -ex; chmod 755 /fts/entrypoint_rxdm_container.sh; \n\
+            /fts/entrypoint_rxdm_container.sh --num_hops=2 --num_nics=8 \
+                --uid= --alsologtostderr &\n\
+            while [ ! -f /run/tcpx/terminated ]; do sleep 10; done;",
```
samos123, line 1086: for potential follow up, with k8s native sidecars, we should be able to remove this `while [ ! -f /run/tcpx/terminated ]; do sleep 10; done;"` and then launch the sidecar process in foreground.

samos123 (commented):
Left some comments.

Dockerfile
andersensam: Fixed

pyproject.toml
andersensam: Fixed

axlearn/common/flash_attention/gpu_attention.py
andersensam: Relabeled with TODO. Asking Triton team if we have an external tracker for this workaround

samos123:
@kelvin-zou @Ethanlm since we couldn't get a3 ultra PR merged in-time we now have a bigger PR that includes both A3 Ultra and A4. Could you give it another review? We want to get his merged ASAP since we have a lot more follow up work and this PR just keeps getting bigger.

Note that this code path is solely for being able to test AXLearn from GKE directly using `axlearn gke start` for GPU training benchmarks.

PR #1098, opened 2025-04-04
Significantly improved performance for GPU FP32 FlashAttention

hanzhi713:
Previously, FP32 code path for the GPU FlashAttention uses pure FP32 arithmetic, which disabled the use of TensorCore, making the kernel more than **100x** slower than expected. We fixed this by using the TF32 preset rather than `jax.lax.Precision.HIGHEST`. This is also consistent with the default precision of FP32 matmul in Jax GPU.

Additionally, we try to use `jax.lax.DotAlgorithmPreset` when possible, as it's more clear to the reader what the actual compute and accumulation precision is. We only fallbacks to `jax.lax.Precision` when it's not possible to use `DotAlgorithmPreset`.

FP32 benchmarks are added for GPU FlashAttention.

This PR is proved to speed up some FP32 eval workload from 14hr to 35min.

Forward pass numbers with pure FP32 (unit=ms):
```
is_decode=False, use_bwd=False, num_heads=32, num_kv_heads=None, seq_len=4096, per_head_dim=128, sw_sz=-1
                                                  jax           axlearn       
bs=2,dtype=<class 'jax.numpy.float32'>            9.690944      420.148224    
bs=4,dtype=<class 'jax.numpy.float32'>            19.689793     821.556946    
bs=8,dtype=<class 'jax.numpy.float32'>            39.802753     1622.739136   
```

Forward pass numbers with TF32:
```
is_decode=False, use_bwd=False, num_heads=32, num_kv_heads=None, seq_len=4096, per_head_dim=128, sw_sz=-1
                                                  jax           axlearn
bs=2,dtype=<class 'jax.numpy.float32'>            9.706688      3.208640
bs=4,dtype=<class 'jax.numpy.float32'>            19.720287     6.164672
bs=8,dtype=<class 'jax.numpy.float32'>            39.786495     12.005504
```

### DIFF
--- a/axlearn/common/flash_attention/common.py
+++ b/axlearn/common/flash_attention/common.py
@@ -293,7 +293,11 @@ class ReferenceMHA(BaseFlashAttention):
 
 
 def get_cpu_dot_precision(dtype) -> jax.lax.DotAlgorithmPreset:
-    """Get the suitable DotAlgorithmPreset for the given dtype."""
+    """Get the suitable DotAlgorithmPreset for the given dtype for CPU backend.
+
+    CPU doesn't support different compute and accumulation precision. This should only be used
+    for CPU emulation and unit tests.
+    """
     if dtype == jnp.float32:
         return jax.lax.DotAlgorithmPreset.F32_F32_F32
     if dtype == jnp.float16:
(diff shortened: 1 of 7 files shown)


hanzhi713:
@liboyue The kernel itself is 100x slower than the TF32 one. I have some hypothesis below

1. Using TensorCore should significantly reduce the register pressure. TensorCore has dedicated units for load/store to/from TensorCore. This avoids expensive register spilling. 
2. (A minor reason) When using TensorCore matmul, other non matmul flops can be offloaded to CUDA cores. When not using TensorCore matmul, we lose this opportunity for overlapping.

hanzhi713:
@markblee Could you please take a look?

markblee:
> @markblee Could you please take a look?

Please see internal comments.

markblee (approved):

PR #1103, opened 2025-04-09
Remove einops dependency by reimplementing minimal functionality for JAX

ds-hwang:
Reimplement einops.rearrange and einops.repeat specifically for JAX, covering only the functionality we use.

While einops (https://openreview.net/pdf?id=oapKSVM2bcj) significantly improves code readability, we decided to remove the dependency and use internal equivalents due to the following concerns:
* Thread safety issues:
  * Backend initialization is not thread-safe. See https://github.com/arogozhnikov/einops/issues/372.
  * It uses global variables internally, so we are not sure if it's thread-safe.
* External dependency: As a personal project, einops isn't what we want to rely on in production.

Roughly speaking, rearrange() is implemented as: reshape → transpose → reshape And repeat() as: reshape → tile → reshape
If any of the ops are identity, they are skipped.

### DIFF
--- a/axlearn/audio/encoder_asr.py
+++ b/axlearn/audio/encoder_asr.py
@@ -11,7 +11,6 @@ from math import prod
 from typing import Any, Optional
 
 import jax.numpy as jnp
-from einops import rearrange
 
 from axlearn.audio.frontend import LogMelFrontend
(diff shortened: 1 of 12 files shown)


ds-hwang:
@ruomingp Could you take a look? From 1194

PR #1106, opened 2025-04-11
[Jax API update] Update jax.core.Primitive to jax.extend.core.Primitive

Steboss:


axlearn/common/decoding_test.py
```diff
@@ -827,6 +827,7 @@ def tokens_to_scores(tokens, cache):
         prefix_merger=[None, _TokenSumPrefixMerger()],
         brevity_penalty=[None, decoding.brevity_penalty_fn(bp_type="hf", alpha=1.0)],
     )
+    # pylint: disable=R0917
```
apghml, line 830: Nit: use the human-readable name?

Steboss: Modified :)

axlearn/common/decoding_test.py
```diff
@@ -827,6 +827,7 @@ def tokens_to_scores(tokens, cache):
         prefix_merger=[None, _TokenSumPrefixMerger()],
         brevity_penalty=[None, decoding.brevity_penalty_fn(bp_type="hf", alpha=1.0)],
     )
+    # pylint: disable=too-many-arguments
```
apghml, line 830: Can we use `disable-next` or reenable it afterward?

axlearn/common/optimizers.py
```diff
@@ -704,6 +704,7 @@ def sgd_optimizer(
         )
 
 
+# pylint: disable=R0913
```
apghml, line 707: I think there was a comment previously about replacing these with the human-readable names? Also are we sure we actually need them? Can you make sure your local pylint version matches the version pinned in pyproject.toml and see if you can remove them? (Also previous comment about not disabling this globally.)

Steboss: You're right, there's been a version mismatch. I fixed this and re-run `pylint`. Let me know if it's ok now, thank you

apghml (approved):

axlearn/common/array_serialization.py
```diff
@@ -596,7 +596,7 @@ async def _run_deserializer():
             byte_limiter = serialization._LimitInFlightBytes(concurrent_bytes)
             h2d_limiter = serialization._LimitInFlightBytes(_get_premapped_buffer_size())
 
-            future_arrays = jax.tree_util.tree_map(
+            future_arrays = jax.tree.map(
```
Ethanlm, line 599: Is `jax.tree.map` preferred over `jax.tree_util.tree_map`? I saw below message on my test, but it is not clear which one is preferred 

```
>>> import jax
>>> jax.tree.map(lambda x: x + 1, {"x": 7, "y": 42})
{'x': 8, 'y': 43}
>>> jax.tree_map(lambda x: x + 1, {"x": 7, "y": 42})
<stdin>:1: DeprecationWarning: jax.tree_map is deprecated: use jax.tree.map (jax v0.4.25 or newer) or jax.tree_util.tree_map (any JAX version).
{'x': 8, 'y': 43}
>>> 

```

Ethanlm (commented):
Otherwise api deprecation, do we expect any functionality or performance change from the new APIs?

axlearn/common/array_serialization.py
olupton: https://docs.jax.dev/en/latest/changelog.html#jax-0-4-26-april-3-2024 says
> `jax.tree_map()` is deprecated; use `jax.tree.map` instead, or for backward compatibility with older JAX versions, use [`jax.tree_util.tree_map()`](https://docs.jax.dev/en/latest/_autosummary/jax.tree_util.tree_map.html#jax.tree_util.tree_map).

sounds like a mild preference for `jax.tree.map`, but I'm not aware of it being a big deal (unless you need to support JAX older than 0.4.25)

Ethanlm (approved):

Steboss:
@Ethanlm or @apghml could anyone of you merge this PR please?

PR #1110, opened 2025-04-15
Set the logmel upper bound to the Nyquist frequency based on the sample rate.

ds-hwang:
The previous value 7600 was hard-coded, corresponds to the Nyquist rate for 16kHz. For 24kHz, the upper bound should be different. This PR fixes 24kHz's the upper bound while no change for 16kHz.

There are currently no production models using 24kHz logmel. Experimental models will need to be retrained.

Authored by m_padinjaruveetti@apple.com

### DIFF
--- a/axlearn/audio/frontend.py
+++ b/axlearn/audio/frontend.py
@@ -86,9 +86,22 @@ class BaseFrontend(BaseLayer):
 
 
 def _log_mel_spectrogram(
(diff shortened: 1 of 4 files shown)


ds-hwang:
@markblee Could you review? From 1214

markblee (approved):

PR #1114, opened 2025-04-18
repro PR for test_spmd_evaler fail with jax 0.5.3

matthew-e-hopkins:
PR for test failure of EvalerTest.test_spmd_evaler0 EvalerTest.test_spmd_evaler1

Test checks that `eval_step` only runs once but current failing state shows that it is now running twice.


kelvin-zou:
Close it since issue has been identified and addressed

PR #1117, opened 2025-04-18
Upgrade grain version.

markblee:
The update allows us to run grain tests locally on macos. 

`packing.SingleBinPackIterDataset` is removed in grain, so we remove `trim_and_pack_dataset`. (Not currently used.)


RsEnts (approved):
Thanks Mark! LGTM.

ruomingp (approved):

PR #1118, opened 2025-04-19
Makes logic in optimizers use 1-based steps.

ruomingp:
Resolves inconsistent definitions of "step" between SpmdTrainer and optimizers.

# Background

SpmdTrainer.step is used for summaries and checkpoints and defined as:
* step=0: the state after initialization. A checkpoint and summaries may be saved at step 0 to reflect the initial state before the first training step;
* step=n: summaries during the n'th training step and checkpoint (if any) saved after n steps

In optimizers, the use of steps are inconsistent. Most use 0-based steps, where the first training step uses the schedule value computed from count=0:
* `scale_by_schedule` (used for learning rate schedules)
* `add_decayed_weights`
* `param_ema` (different from `ema`)
* `skip_and_clip_by_global_norm` uses 0-based steps to compute the drop_norm threshold
* `decay_bias_correction` assumes a 0-based step

# Resolution

After discussion, we decided to change the optimizer logic to be consistent with SpmdTrainer.step. Specifically, we change `decay_bias_correction`, `adafactor_decay_rate`, `scale_by_schedule`, `add_decayed_weights`, `param_ema`, and `skip_and_clip_by_global_norm` to assume that the first step is 1.

This will make optimizer steps to be consistent with summary steps, e.g., the summary at step N will reflect the learning rate schedule computed for step N.

### DIFF
--- a/axlearn/common/factorized_rms.py
+++ b/axlearn/common/factorized_rms.py
@@ -133,6 +133,8 @@ def scale_by_factored_rms(
         """Apply gradient transformation."""
         if params is None:
             raise ValueError("param is None")
+        # Step starts from 1.
+        count_inc = optax.safe_int32_increment(state.count)
 
         def _update(grad, v_row, v_col, v, param, step):
             grad = grad.astype(jnp.float32)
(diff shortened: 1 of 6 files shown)


ruomingp:
This PR is ready for review now. Thanks!

markblee (approved):

PR #1122, opened 2025-04-21
Set `prevent_cse` to False for most fuji meshes

hanzhi713:
Since we're using `RepeatedTransformer`, `prevent_cse` should set to False. 

Neuron backend is an exception here, since it uses `StackedTransformer`


markblee (approved):

hanzhi713:
@markblee Could you re-approve this PR? Rebased master to resolve CI flakiness.

hanzhi713:
@markblee

markblee (approved):

PR #1131, opened 2025-04-24
Bump minimum transformers version to 4.51.3

yuanliu939:


apghml (commented):
Hi was this pr approved internally first? Also, is there an internal bump PR?

apghml (approved):

PR #1132, opened 2025-04-25
Fix Splash Attention unittest for sliding window attention.

ds-hwang:
This is the same skip to
https://github.com/apple/axlearn/blob/89b5345f0159baebe5d1720d997d92fc8c4ecb66/axlearn/common/flash_attention/layer_test.py#L551

Note: The bias indicates which keys each query is allowed to attend to. If a query is not allowed to attend to any keys, then its output can technically be any arbitrary value. Splash attention returns NaN for those, while reference attentions returns avg value of values.

@lgayathrikolluru reported this issue.

TEST=test_forward_and_backward194


ds-hwang:
@markblee Could you review it? From 1251.

markblee (approved):

PR #1133, opened 2025-04-25
The dtype of `paddings` and `mask` is now `jnp.bool`.

ds-hwang:
Previously, it was typically `int32`, but since `paddings` only store 0/1 values, that's redundant. We prefer to use `jnp.bool`, which uses only 1 byte. For reference, paddings are currently used inconsistently across the codebase; some places use int32, others float32, and elsewhere bool. This PR standardizes all of them to bool.

Additionally, **XLA may store booleans as 1 bit per value on TPU via bit-packing** (though this is not officially documented).

When switching paddings from `int32` to `bool`, the following improvements in memory and runtime were observed:
- **HBM usage**: 86,242.00 MiB → 86,232.91 MiB
- **Step time** (measured 3 times): 1,345,702,414 ± 128,038 ns → 1,345,565,230 ± 136,309 ns

* int32 paddings: 
image

* bool paddings:
image

### DIFF
--- a/axlearn/audio/decoder_asr.py
+++ b/axlearn/audio/decoder_asr.py
@@ -498,7 +498,7 @@ class CTCDecoderModel(BaseASRDecoderModel):
         max_decode_len = paddings.shape[-1] + 1
         beam_search_outputs = beam_search_decode(
             inputs=jnp.zeros_like(paddings),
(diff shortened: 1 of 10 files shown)


ds-hwang:
@markblee Could you review it? From 1247.

apghml (approved):

markblee (approved):

Issue #1126, opened 2025-04-23
Support for JAX versions > 0.5.4 (current AXLearn pinned to 0.4.38)


Steboss:
AXLearn currently supports JAX 0.4.38, while JAX has now progressed to 0.6.0. When attempting to use AXLearn with newer JAX releases (≥ 0.5.4), we see several compatibility issues:
1. `jax.core.Primitive` is updated to `jax.extend.core.Primitive`
2. `jax.tree_map` is now `jax.tree.map`. (These first two points have been addressed in these PR #1082 and #1106 )
3. In `axlearn/common/utils.py`, the call to 'jax._src.tree_util' references `_registry_with_keypaths`, which no longer exists. I’ve proposed a fix in my fork [here](https://github.com/Steboss/axlearn/blob/main/axlearn/common/utils.py#L1858)
4. The config option `jax_spmd_mode` has been removed in JAX (see [here](https://github.com/jax-ml/jax/commit/7634230cdcd2d3cb42d1093f6ab255f47f9869d5) ). This causes the following error:
```bash
File "/opt/axlearn/axlearn/common/launch.py", line 115, in setup
    setup_spmd(
File "/opt/axlearn/axlearn/common/utils_spmd.py", line 47, in setup
     jax.config.update("jax_spmd_mode", "allow_all")
File "/opt/jaxlibs/jax/jax/_src/config.py", line 111, in update
     raise AttributeError(f"Unrecognized config option: {name}")
AttributeError: Unrecognized config option: jax_spmd_mode
```
 
I can contribute patches to address these errors. However, additional optimizations or performance checks might be required before merging. 
Let me know how nvidia team can collaborate with you on this. Thank you

Steboss:
It looks like there's a new error we should be aware of: 
```
File "/opt/axlearn/axlearn/common/config.py", line 282, in validate_config_field_value
raise InvalidConfigValueError(
 axlearn.common.config.InvalidConfigValueError: Invalid config value type <class 'jax._src.partition_spec.PartitionSpecImpl'> for value "PartitionSpec(('data', 'expert', 'fsdp', 'seq'),)". Consider registering a custom validator with `register_validator`.
```

[...]

PR #1136, opened 2025-04-28
[Jax API update] Remove `jax_spmd_mode` 

Steboss:
This PR is a continuation of #1106, and it refers to issue #1126 
In particular here the major change is in `axlearn/common/utils_spmd.py`, where the config doesn't need to be updated anymore, as `jax_spmd_mode` has been removed in JAX 0.6.0 ([here](https://github.com/jax-ml/jax/commit/7634230cdcd2d3cb42d1093f6ab255f47f9869d5))


axlearn/common/utils_spmd.py
```diff
@@ -43,8 +43,6 @@ def setup(
     """
     # Use a GSPMD-friendly PRNG implementation.
     jax.config.update("jax_default_prng_impl", "rbg")
-    # This allows replicated jax.Arrays to be used for computation on the host.
-    jax.config.update("jax_spmd_mode", "allow_all")
```
apghml, line 47: The behavior that is always active in 0.6.0 is equivalent to the previous allow_all behavior?

axlearn/common/test_utils.py
```diff
@@ -469,14 +469,14 @@ def replace_keys(v, mapping):
 
     # Complete the param_init_specs to match the params treedef.
     # Replace with Nones so that jax doesn't treat them as leaves.
-    params_with_nones = jax.tree_map(
+    params_with_nones = jax.tree.map(
```
apghml, line 472: Is the previous pr ready to merge? If so, you can comment asking someone with access to merge it and then we can rebase this pr.

axlearn/common/utils.py
```diff
@@ -31,28 +31,29 @@
...
 import jax
-import jax.flatten_util
 import numpy as np
 from absl import logging
 from jax import numpy as jnp
 from jax._src.ad_checkpoint import name_p
 from jax._src.lax import lax as lax_internal
 from jax._src.mesh import thread_resources
-from jax._src.tree_util import KeyEntry, KeyPath
+from jax._src.tree_util import KeyEntry, KeyPath, flatten_one_level_with_keys
```
apghml, line 53: Any chance `flatten_one_level_with_keys` or similar functionality could be made public?

axlearn/common/test_utils.py
Steboss: Ready to be merged

axlearn/common/utils.py
Steboss: I can check on this, if there's another way to solve it

axlearn/common/utils_spmd.py
Steboss: The behaviour is equivalent to the previous `allow_all`. 
I further tested this code on `fuji-3B-v3-flash-attention` and results are in line with what we were expecting (from previous runs).

axlearn/common/utils.py
Steboss: Unfortunately, I can't find any public alternative to this solution

Steboss:
@apghml rebased the PR with the current `main` :)

axlearn/common/utils.py
```diff
@@ -1853,7 +1854,7 @@ def thread_stack_traces() -> Sequence[Sequence[str]]:
     return grouped_lines
 
 
-def pytree_children(node: Any) -> Sequence[tuple[KeyEntry, Any]]:
+def pytree_children(node: Any) -> List[Tuple[KeyEntry, Any]]:
```
apghml, line 1857:
```suggestion
def pytree_children(node: Any) -> list[tuple[KeyEntry, Any]]:
```

axlearn/common/utils.py
```diff
@@ -1866,23 +1867,13 @@ def pytree_children(node: Any) -> Sequence[tuple[KeyEntry, Any]]:
...
-
-    key_handler = registry_with_keypaths.get(type(node))
-    if key_handler:
-        key_children, _ = key_handler.flatten_with_keys(node)
-        return key_children
-
-    flat = jax.tree_util.default_registry.flatten_one_level(node)
-    if flat is None:
+    try:
+        # pylint: disable-next=protected-access
```
apghml, line 1871: Do we still need this pylint directive?

axlearn/common/utils.py
```diff
@@ -1866,23 +1867,13 @@ def pytree_children(node: Any) -> Sequence[tuple[KeyEntry, Any]]:
         assert pytree_children(dict(a=[1,2])) == [(DictKey('a'), [1,2])]
```
apghml, line 1867: Is the new implementation still referencing `jax._src.tree_util.generate_key_paths()`? If not, update the docstring?

Also, doe the returned children still match those returned by  `jax.tree_util.default_registry.flatten_one_level()`?

Steboss: > Is the new implementation still referencing jax._src.tree_util.generate_key_paths()? If not, update the docstring?
You're right, I'll update the docstring accordingly 
> Also, doe the returned children still match those returned by jax.tree_util.default_registry.flatten_one_level()?
Good point. I ran a test and I could see children match, however, do you have a test that I could use to be 100% sure about this?

axlearn/common/utils.py
Steboss: `pylint` can be removed here, done

axlearn/common/utils.py
Steboss: @apghml I can confirm that the returned children still match `jax.tree_util.default_registry.flatten_one_level()`

apghml: Thanks for checking. Thinking about this a bit more, does it make sense to use the (public?) API `jax.tree_util.default_registry.flatten_one_level_with_keys()` instead of

Steboss: @apghml  I owe you an apology for this. Made the call with the public API, and check it still match: 

```python
        assert pytree_children(dict(a=[1,2])) == [(DictKey('a'), [1,2])]
```
thanks for your help

axlearn/common/utils.py
```diff
@@ -1853,36 +1853,23 @@ def thread_stack_traces() -> Sequence[Sequence[str]]:
...
-    if key_handler:
-        key_children, _ = key_handler.flatten_with_keys(node)
-        return key_children
-
-    flat = jax.tree_util.default_registry.flatten_one_level(node)
-    if flat is None:
+    try:
+        key_child_pairs, _ = default_registry.flatten_one_level_with_keys(node)
+        return list(key_child_pairs)
+    except ValueError:
```
apghml, line 1870: Do we still need the `try/except` statement now?

Steboss: I've run a test by training a fuji model, and I couldn't spot a `ValueError`.
I wanted to be consistent with what we had before. Are there more tests that I could run just to be 100% sure we can avoid having `try/except`? thank you

apghml: The original private implementation of `jax._src.tree_util.flatten_one_level_with_keys` is the following:

```
def flatten_one_level_with_keys(
    tree: Any,
) -> tuple[Iterable[KeyLeafPair], Hashable]:
  """Flatten the given pytree node by one level, with keys."""
  out = default_registry.flatten_one_level_with_keys(tree)  # type: ignore
  if out is None:
    raise ValueError(f"can't tree-flatten type: {type(tree)}")
  else:
    return out
```

So maybe we should do a None check and return an empty list in that case?

Steboss: Oh yes absolutely, I'll add a none check, and we should be fine. Thank you @apghml

Steboss: Ok i've removed the `try/except` 
We don't really need a `None` check actually. If `node` is `None`, `key_child_pairs=()`, so this will be returned as an empty list.

apghml: Do we know why the check is done in the original private implementation? What if `node` is something like `object()`?

Steboss: I can run a fuji model and examine the content of `node`, so we'll get some more clarity on this. Unfortunately, I don't know what was the original rational for this. @kelvin-zou any idea on this? thank you

apghml: I think the try/except is for cases like this:
```
import jax
result = jax.tree_util.default_registry.flatten_one_level_with_keys(object())
print(result)  # None
```

So we should return an empty list in that case.

Steboss:
hey @apghml 
So I investigated over the `pytree_children` function, so that it could match exactly the previous implementation. For simplicity, i'll copy and paste here a part of the initial output only. This is the current implementation 

```bash
I0502 06:21:33.000848 140737350427392 utils.py:1878] ((GetAttrKey(name='output_collection'), OutputCollection(summaries={}, state_updates={}, module_outputs={})), (GetAttrKey(name='parent'), None), (GetAttrKey(name='prng_key'), Traced<ShapedArray(uint32[4])>with<DynamicJaxprTrace>), (GetAttrKey(name='state'), {'optimizer': (EmptyState(), (ScaleByAdamState(count=Traced<ShapedArray(int32[])>with<DynamicJaxprTrace>, mu={'decoder': {'emb': {'dropout': {}, 'token_emb': {'weight': Traced<ShapedArray(float32[131072,3072])>with<DynamicJaxprTrace>}}, 'output_dropout': {}, 'output_norm': {'scale': Traced<ShapedArray(float32[3072])>with<DynamicJaxprTrace>}, 'transformer': {'repeat': VDict({'layer': {'feed_forward': {'dropout1': {}, 'dropout2': {}, 'linear1_0': {'weight': Traced<ShapedArray(float32[28,3072,8192])>with<DynamicJaxprTrace>}, 'linear1_1': {'weight': Traced<ShapedArray(float32[28,3072,8192])>with<DynamicJaxprTrace>}, 'linear2': {'weight': Traced<ShapedArray(float32[28,8192,3072])>with<DynamicJaxprTrace>}, 'norm': {'scale': Traced<ShapedArray(float32[28,3072])>with<DynamicJaxprTrace>}, 'stochastic_depth': {}}, 'self_attention': {'attention': {'dropout': {}, 'i_proj': {'i_proj': {'qkv_proj': {'weight': Traced<ShapedArray(float32[28,3072,40,128])>with<DynamicJaxprTrace>}}, 'rope_pos_emb_layer': {}}, 'kv_cache': {}, 'o_proj': {'weight': Traced<ShapedArray(float32[28,3072,24,128])>with<DynamicJaxprTrace>}, 'scale_key': {}, 'scale_query': {}}, 'dropout': {}, 'norm': {'scale': Traced<ShapedArray(float32[28,3072])>with<DynamicJaxprTrace>}, 'stochastic_depth': {}}}})}}, 'metrics': {'aux': {}, 'lm': {}}}, nu={'decoder': {'emb': {'dropout': {}, 'token_emb': {'weight': Traced<ShapedArray(float32[131072,3072])>with<DynamicJaxprTrace>}}, 'output_dropout': {}, 'output_norm': {'scale': Traced<ShapedArray(float32[3072])>with<DynamicJaxprTrace>}, 'transformer': {'repeat': VDict({'layer': {'feed_forward': {'dropout1': {}, 'dropout2': {}, 'linear1_0': {'weight': Traced<ShapedArray(float32[28,3072,8192])>with<DynamicJaxprTrace>}, 'linear1_1': {'weight': Traced<ShapedArray(float32[28,3072,8192])>with<DynamicJaxprTrace>}, 'linear2': {'weight': Traced<ShapedArray(float32[28,8192,3072])>with<DynamicJaxprTrace>}, 'norm': {'scale': Traced<ShapedArray(float32[28,3072])>with<DynamicJaxprTrace>}, 'stochastic_depth': {}}, 'self_attention': {'attention': {'dropout': {}, 'i_proj': {'i_proj': {'qkv_proj': {'weight': Traced<ShapedArray(float32[28,3072,40,128])>with<DynamicJaxprTrace>}}, 'rope_pos_emb_layer': {}}, 'kv_cache': {}, 'o_proj': {'weight': Traced<ShapedArray(float32[28,3072,24,128])>with<DynamicJaxprTrace>}, 'scale_key': {}, 'scale_query': {}}, 'dropout': {}, 'norm': {'scale': Traced<ShapedArray(float32[28,3072])>with<DynamicJaxprTrace>}, 'stochastic_depth': {}}}})}}, 'metrics': {'aux': {}, 'lm': {}}}), ScaleByScheduleState(count=Traced<ShapedArray(int32[])>with<DynamicJaxprTrace>), AddDecayedWeightsState(count=None), ScaleByScheduleState(count=Traced<ShapedArray(int32[])>with<DynamicJaxprTrace>), EmptyState()))}))


I0502 06:21:33.000930 140737350427392 utils.py:1888] [(GetAttrKey(name='summaries'), {}), (GetAttrKey(name='state_updates'), {}), (GetAttrKey(name='module_outputs'), {})]
```
and this is the output of the new implementation 
```
I0502 09:17:33.186500 140737350427392 utils.py:1893] [(GetAttrKey(name='output_collection'), OutputCollection(summaries={}, state_updates={}, module_outputs={})), (GetAttrKey(name='parent'), None), (GetAttrKey(name='prng_key'), Traced<uint32[4]>with<DynamicJaxprTrace>), (GetAttrKey(name='state'), {'optimizer': (EmptyState(), (ScaleByAdamState(count=Traced<int32[]>with<DynamicJaxprTrace>, mu={'decoder': {'emb': {'dropout': {}, 'token_emb': {'weight': Traced<float32[131072,3072]>with<DynamicJaxprTrace>}}, 'output_dropout': {}, 'output_norm': {'scale': Traced<float32[3072]>with<DynamicJaxprTrace>}, 'transformer': {'repeat': VDict({'layer': {'feed_forward': {'dropout1': {}, 'dropout2': {}, 'linear1_0': {'weight': Traced<float32[28,3072,8192]>with<DynamicJaxprTrace>}, 'linear1_1': {'weight': Traced<float32[28,3072,8192]>with<DynamicJaxprTrace>}, 'linear2': {'weight': Traced<float32[28,8192,3072]>with<DynamicJaxprTrace>}, 'norm': {'scale': Traced<float32[28,3072]>with<DynamicJaxprTrace>}, 'stochastic_depth': {}}, 'self_attention': {'attention': {'dropout': {}, 'i_proj': {'i_proj': {'qkv_proj': {'weight': Traced<float32[28,3072,40,128]>with<DynamicJaxprTrace>}}, 'rope_pos_emb_layer': {}}, 'kv_cache': {}, 'o_proj': {'weight': Traced<float32[28,3072,24,128]>with<DynamicJaxprTrace>}, 'scale_key': {}, 'scale_query': {}}, 'dropout': {}, 'norm': {'scale': Traced<float32[28,3072]>with<DynamicJaxprTrace>}, 'stochastic_depth': {}}}})}}, 'metrics': {'aux': {}, 'lm': {}}}, nu={'decoder': {'emb': {'dropout': {}, 'token_emb': {'weight': Traced<float32[131072,3072]>with<DynamicJaxprTrace>}}, 'output_dropout': {}, 'output_norm': {'scale': Traced<float32[3072]>with<DynamicJaxprTrace>}, 'transformer': {'repeat': VDict({'layer': {'feed_forward': {'dropout1': {}, 'dropout2': {}, 'linear1_0': {'weight': Traced<float32[28,3072,8192]>with<DynamicJaxprTrace>}, 'linear1_1': {'weight': Traced<float32[28,3072,8192]>with<DynamicJaxprTrace>}, 'linear2': {'weight': Traced<float32[28,8192,3072]>with<DynamicJaxprTrace>}, 'norm': {'scale': Traced<float32[28,3072]>with<DynamicJaxprTrace>}, 'stochastic_depth': {}}, 'self_attention': {'attention': {'dropout': {}, 'i_proj': {'i_proj': {'qkv_proj': {'weight': Traced<float32[28,3072,40,128]>with<DynamicJaxprTrace>}}, 'rope_pos_emb_layer': {}}, 'kv_cache': {}, 'o_proj': {'weight': Traced<float32[28,3072,24,128]>with<DynamicJaxprTrace>}, 'scale_key': {}, 'scale_query': {}}, 'dropout': {}, 'norm': {'scale': Traced<float32[28,3072]>with<DynamicJaxprTrace>}, 'stochastic_depth': {}}}})}}, 'metrics': {'aux': {}, 'lm': {}}}), ScaleByScheduleState(count=Traced<int32[]>with<DynamicJaxprTrace>), AddDecayedWeightsState(count=None), ScaleByScheduleState(count=Traced<int32[]>with<DynamicJaxprTrace>), EmptyState()))})]

I0502 09:17:33.186563 140737350427392 utils.py:1876] [(GetAttrKey(name='summaries'), {}), (GetAttrKey(name='state_updates'), {}), (GetAttrKey(name='module_outputs'), {})]

```
I tested by running `fuji-3B-v3-flash`, c4 dataset, ICI FSDP=8, global batch size 16, sequence length 4096. The only difference in the output is the print for `Traced<ShapedArray(int32[])>` that, now, with JAX is `Traced<int32[]>` - so nothing to be worried about. 

I've updated the test suite, so that we have: 
- a custom object with fields `_fields`, like `OutputCollection(summaries={}, state_updates={}, module_outputs={}))` that we see as input `node` 
- and a test with `object()`: 
```python 
obj = object()
self.assertSequenceEqual(pytree_children(obj), [])
```
The test on `pytree_children` works fine. 

Finally, I checked the performance, related to the removal of `spmd_mode`, and the overall results are the same: 

| Metrics | This PR implementation |  Previous AXLearn implementation |
| --- | --- | --- |
| Tokens per sec per gpu | 9288 | 8904 |
| Seqs per sec per gpu | 2.26 | 2.17 |
| Average time step | 0.88 | 0.91 |
| TFLOPS per sec per GPU | 218.80 | 209.74 |

axlearn/common/utils_test.py
```diff
@@ -334,6 +334,25 @@ class TestUnstructured:
...
+            _fields = ("a", "b", "c")
+
+            def __init__(self, a, b, c):
+                self.a = a
+                self.b = b
+                self.c = c
+
+        tree = CustomWithFields(**original_tree)
+        self.assertSequenceEqual(
+            pytree_children(tree),
```
apghml, line 348: Hm..., this seems to not match the behavior of `jax.tree.leaves()`?

```
import jax

class CustomWithFields:
    _fields = ("a", "b", "c")

    def __init__(self, a, b, c):
        self.a = a
        self.b = b
        self.c = c

print(jax.tree.leaves(CustomWithFields(1,2,3)))  # Prints the entire object as a single leaf.
```

axlearn/common/utils.py
```diff
@@ -1853,35 +1853,38 @@ def thread_stack_traces() -> Sequence[Sequence[str]]:
...
     """
-    # pylint: disable-next=protected-access
-    registry_with_keypaths = jax._src.tree_util._registry_with_keypaths
-
-    key_handler = registry_with_keypaths.get(type(node))
-    if key_handler:
-        key_children, _ = key_handler.flatten_with_keys(node)
-        return key_children
+    # If node is a NamedTuple
+    if isinstance(node, tuple) and hasattr(node, "_fields"):
```
apghml, line 1868: Do we need this? It looks like the output is already correct?

```
from typing import NamedTuple

import jax

class C(NamedTuple):
    a: int

print(jax.tree_util.default_registry.flatten_one_level_with_keys(C(1)))
```

axlearn/common/utils.py
```diff
@@ -1853,35 +1853,38 @@ def thread_stack_traces() -> Sequence[Sequence[str]]:
...
+    # Node is Sequence
     flat = jax.tree_util.default_registry.flatten_one_level(node)
     if flat is None:
         return []
 
-    if isinstance(node, tuple) and hasattr(node, "_fields") and flat[1] == type(node):
-        # Handle namedtuple as a special case, based on heuristic.
-        return [(jax.tree_util.GetAttrKey(s), getattr(node, s)) for s in node._fields]
-    return [(jax.tree_util.FlattenedIndexKey(i), c) for i, c in enumerate(flat[0])]
+    return [(jax.tree_util.FlattenedIndexKey(i), child) for i, child in enumerate(flat[0])]
```
apghml, line 1887: Overall I'm a bit nervous that the various changes in this function may cause us to deviate from what jax does, even if it has the same behavior in existing axlearn code. Was there a reason we needed to change this function? If so, can we reference however jax's source code is generating the key paths in jax.tree.map_with_path()` to see how they do it and exactly mirror that logic?

axlearn/common/utils.py
apghml: Also why did we move this earlier in the function body? IIUC, this will change the behavior in the case that someone creates a NamedTuple subclass and registers a custom tree flattening handler for it?

axlearn/common/utils.py
```diff
@@ -1853,35 +1853,38 @@ def thread_stack_traces() -> Sequence[Sequence[str]]:
...
+
+    # If node is not a NT but exposes a public `_fields` attribute
+    if hasattr(node, "_fields") and not isinstance(node, tuple):
+        return [(jax.tree_util.GetAttrKey(name), getattr(node, name)) for name in node._fields]
+    # Standard JAX
+    try:
+        key_child_pairs, _ = default_registry.flatten_one_level_with_keys(node)
+        if key_child_pairs:
+            return list(key_child_pairs)
+    except (ValueError, TypeError):
```
apghml, line 1879: Could you explain why we still need to catch these errors?

Steboss:
@apghml thanks for your comments. 
let's to do in this way: 
- i'll create a PR only for `jax_spmd_mode` removal 
- and i'll create a new PR for `utils`, so I'll check if we really need to restructure the `utils.py` `pytree_children` function completely, or if JAX can give us some alternatives
What do you think?

axlearn/common/utils.py
```diff
@@ -1853,35 +1853,38 @@ def thread_stack_traces() -> Sequence[Sequence[str]]:
     return grouped_lines
 
 
-def pytree_children(node: Any) -> Sequence[tuple[KeyEntry, Any]]:
+def pytree_children(node: Any) -> list[tuple[KeyEntry, Any]]:
```
apghml, line 1856: Does this example fail with the new implementation?
```
import jax.numpy as jnp
import jax.tree
from jax.tree_util import GetAttrKey


class MyContainer:

    def __init__(self):

        self.x = jnp.zeros(2)
        self.y = jnp.ones(2)



def flatten_with_keys(obj):

    children = [(GetAttrKey('x'), obj.x),
                (GetAttrKey('y'), obj.y)]  # children must contain arrays & pytrees



    aux_data = ()  # aux_data must contain static, hashable data.

    return children, aux_data

def unflatten(aux_data, children):
    obj = object.__new__(MyContainer)

    obj.x, obj.y = children

    obj.size, = aux_data

    return obj

jax.tree_util.register_pytree_node(MyContainer, flatten_with_keys, unflatten)

pytree_children(MyContainer())
```

PR #1138, opened 2025-04-28
Supports general per-feed processing.

markblee:


axlearn/common/input_grain.py
```diff
@@ -692,17 +693,26 @@ def pad_for_evaluation(
     return ds
 
 
-def _set_read_config_recursively(source: Dataset, **kwargs) -> bool:
-    """Sets **kwargs on all `shard_dataset` in `source`."""
-    if isinstance(source, _ShardDataset):
+@runtime_checkable
+class PerFeedDataset(Protocol):
```
apghml, line 697: It seems kind of unusual to support changing the dispatch config after the input source has already been instantiated. Do we need that capability? E.g., could we recursively set the dispatch config in the actual Config instead?

axlearn/common/input_base.py
```diff
@@ -229,9 +231,14 @@ def batches(self, it: Iterator[Nested[Tensor]]) -> Iterator[Nested[Tensor]]:
                 def check_per_feed_batch(x: Tensor):
                     expected = self.input_dispatcher.feed_logical_batch_size
                     actual = x.shape[0]
-                    if expected != actual:
+                    # The actual batch size is allowed to be smaller than feed batch size in the
+                    # case where global batch sizes differ across paths in the batch.
```
apghml, line 235: What would be a concrete example where the global batch size differs across paths in the batch? Is that something we support elsewhere in axlearn?

axlearn/common/input_grain.py
markblee: It's mainly because the dispatch configs (i.e. the number of feeds) is not known until instantiation (because it depends on how many processes you're running on). We do a similar thing in `input_tf_data`.

axlearn/common/input_base.py
markblee: Responded in internal comment.

apghml (approved):

PR #1141, opened 2025-04-30
Change wait_for_stop to be Optional for backwards compatibility

Ethanlm:
related PR: https://github.com/apple/axlearn/pull/1139

### DIFF
--- a/axlearn/cloud/gcp/jobs/launch.py
+++ b/axlearn/cloud/gcp/jobs/launch.py
@@ -233,7 +233,8 @@ class BaseBastionManagedJob(FlagConfigurable):
         # Resources used by the job.
         resources: Callable[[ConfigBase], ResourceMap[int]] = infer_resources
         # If True, wait for a job to stop when cancelling.
-        wait_for_stop: Required[bool] = REQUIRED
+        # Default to None for backwards compatibility.
+        wait_for_stop: Optional[bool] = None
 
     @classmethod
     def define_flags(cls, fv: flags.FlagValues):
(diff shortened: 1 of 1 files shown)


axlearn/cloud/gcp/jobs/launch.py
```diff
@@ -347,7 +347,9 @@ def __init__(self, cfg: Config):
 
     def cancel(self):
         """Submits a cancel request to bastion."""
-        self._bastion_dir.cancel_job(self.config.name, wait_for_stop=self.config.wait_for_stop)
+        # Default wait_for_stop to True.
+        wait_for_stop = self.config.wait_for_stop if self.config.wait_for_stop is not None else True
+        self._bastion_dir.cancel_job(self.config.name, wait_for_stop=wait_for_stop)
```
markblee, line 352: nit -- 
```suggestion
        cfg = self.config
        # Default wait_for_stop to True, unless explicitly set to False.
        self._bastion_dir.cancel_job(cfg.name, wait_for_stop=cfg.wait_for_stop != False)
```

markblee (approved):

axlearn/cloud/gcp/jobs/launch.py
Ethanlm: Thanks. Updated and squashed commits.

Ethanlm: Used `is not False` due to:

```
 [C0121(singleton-comparison), BaseBastionManagedJob.cancel] Comparison 'cfg.wait_for_stop != False' should be 'cfg.wait_for_stop is not False' if checking for the singleton value False, or 'bool(cfg.wait_for_stop)' if testing for truthiness

```

markblee (approved):

PR #1144, opened 2025-05-01
Differentiate SUCCEED and FAILED at the end of the Flink-Beam job

jinglu1:


markblee (changes requested):
Sorry, I missed this PR internally. Please see comments. Thanks!

PR #1145, opened 2025-05-01
Multihead pathways setup for multi-host infefence use case

Ethanlm:


axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -87,7 +83,40 @@ class PathwaysReplicatedJob(BaseReplicatedJob):
 
     @config_class
     class Config(BaseReplicatedJob.Config):
+        """Configures PathwaysReplicatedJob.
+
+        Attributes:
+            inner: The wrapped TPUReplicatedJob configuration.
+            pathways_head_cpu: CPU request for pathways-head container.
+            pathways_head_mem: Memory request for pathways-head container.
```
markblee, line 91: Maybe we don't need the pathways prefix since it's already defined in the pathways job.

markblee (approved):

axlearn/cloud/gcp/pathways_utils.py
Ethanlm: I added "pathways-head" in the prefix so it is more clear this cpu and memory requirement is about pathways-head, since there are other components like pathways-worker (even though they run on TPUs, it looks more clear that these two flags are only about pathways-head)

axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -47,20 +49,14 @@
...
 
-
-def get_pathways_head_address(job_name: str) -> str:
-    """Returns the address of the pathways head pod."""
-    # There will be only one pathways-head pod, so it is already 0-0.
-    # First 0 means the first replicatedJob of pathways-head,
-    # the second 0 means the first pod in the replicatedJob.
-    return f"{job_name}-{_PATHWAYS_HEAD_REPLICATED_JOB_NAME}-0-0.{job_name}"
+# Add node-selector for cpu workload to avoid sharing nodes with system services.
+_PATHWAYS_HEAD_NODE_POOL_SELECTOR_KEY = "axlearn/nodepool_type"
```
samos123, line 58: I think we should default to not having any nodeSelector to make this more easily usable for users outside of Apple.

PR #1152, opened 2025-05-02
Use lambda instead of functools.partial to create single-host config maker

### LINKED ISSUE #1151: Single-host config maker rejected by `config_class_for_function`
I encountered the following error when trying to run the Fuji 3B model locally:
```
Traceback (most recent call last):
  File "fuji.py", line 339, in <module>
    run_main()
  File "fuji.py", line 219, in run_main
    app.run(lambda _: main(parsed_args))
  File "venv/lib/python3.12/site-packages/absl/app.py", line 308, in run
    _run_main(main, args)
  File "venv/lib/python3.12/site-packages/absl/app.py", line 254, in _run_main
    sys.exit(main(argv))
             ^^^^^^^^^^
  File "fuji.py", line 219, in <lambda>
    app.run(lambda _: main(parsed_args))
                      ^^^^^^^^^^^^^^^^^
  File "fuji.py", line 274, in main
    trainer_config: SpmdTrainer.Config = config_for_function(config_fn).fn()
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "axlearn/axlearn/common/config.py", line 918, in config_for_function
    config_cls = _config_class_for_function(fn)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "axlearn/axlearn/common/config.py", line 891, in _config_class_for_function
    name=f"config_for_function({fn.__module__}.{fn.__qualname__})",
                                                ^^^^^^^^^^^^^^^
AttributeError: 'functools.partial' object has no attribute '__qualname__'
```
This is apparently caused by https://github.com/apple/axlearn/blob/608e344a44ab54b5afc80328aa9f8026ff36ed98/axlearn/common/config.py#L891
as its input in this case is https://github.com/apple/axlearn/blob/608e344a44ab54b5afc80328aa9f8026ff36ed98/axlearn/experiments/text/gpt/fuji.py#L1014
which is per se not a complete function object due to `functools.partial`.

yhtang:
Fixes #1151 

An alternative solution is to use `functools.wraps` but that requires a custom config validator:
```
axlearn.common.config.InvalidConfigValueError: Invalid config value type <class 'functools.partial'> for value "functools.partial(<function trainer_configs.<locals>.make_single_host_config at 0x722900936160>, 'fuji-3B-v3-flash')". Consider registering a custom validator with `register_validator`.
```

### DIFF
--- a/axlearn/experiments/text/gpt/fuji.py
+++ b/axlearn/experiments/text/gpt/fuji.py
@@ -1011,7 +1011,6 @@ def trainer_configs(
                 return cfg
 
             # Make single-host config
-            make_single_host_config_func = functools.partial(make_single_host_config, config_name)
-            config_map[f"{config_name}-single-host"] = make_single_host_config_func
+            config_map[f"{config_name}-single-host"] = lambda: make_single_host_config(config_name)
 
     return config_map


axlearn/experiments/text/gpt/fuji.py
```diff
@@ -1011,7 +1011,6 @@ def make_single_host_config(base_config_name: str) -> SpmdTrainer.Config:
                 return cfg
 
             # Make single-host config
-            make_single_host_config_func = functools.partial(make_single_host_config, config_name)
-            config_map[f"{config_name}-single-host"] = make_single_host_config_func
+            config_map[f"{config_name}-single-host"] = lambda: make_single_host_config(config_name)
```
apghml, line 1014: IIUC, this won't work as is due to how closures work in Python?
```suggestion
            config_map[f"{config_name}-single-host"] = lambda config_name=config_name: make_single_host_config(config_name)
```

apghml (changes requested):

axlearn/experiments/text/gpt/fuji.py
yhtang: That should work just fine, like how we captured things in the idiomatic way of defining decorators.
Here is a simple test that captures the pattern:
```
def f():
  x = 1
  return lambda: print(x)

g = f()
g()
```
which correctly prints 1 when we called `g()`.

Ethanlm:
> > I have not encountered this issue. How can I reproduce it?
> 
> This is only triggered if using one of the *-single-host configs, e.g. `fuji-3B-v3-flash-single-host`

Yeah I have been using single-host config for my testing on TPU recently.

axlearn/experiments/text/gpt/fuji.py
apghml: Maybe I missed something, but why doesn't it exhibit the same problem as the following code:

```
fns = []
for i in range(3):
    fns.append(lambda: i)
print( fns[0]() ) # Incorrectly prints 2 instead of 0.
```

apghml: (The code in the PR I mean. It makes sense that the snippet you posted in the comment doesn't have this issue.)

ruomingp (commented):
Will defer to @apghml for approval.

PR #1153, opened 2025-05-03
[AudioSummary] Avoid batch axis slicing/indexing

yuanliu939:


yuanliu939:
@markblee Could you review? From 1277

markblee (approved):

PR #1154, opened 2025-05-04
fix: Scale QK in vision/attention

changlan:
`WindowAttention` layer in vision/attention.py overrides the forward calls directly so we need to scale the QK before computing logits.

### DIFF
--- a/axlearn/vision/attention.py
+++ b/axlearn/vision/attention.py
@@ -214,6 +214,18 @@ class WindowedAttention(MultiheadAttention):
         self.vlog(3, "atten.q_proj=%s", q_proj.sum())
         self.vlog(3, "atten.k_proj=%s", k_proj.sum())
         self.vlog(3, "atten.v_proj=%s", v_proj.sum())
+
+        # Scale query and key.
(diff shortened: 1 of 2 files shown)


axlearn/vision/attention.py
```diff
@@ -214,6 +214,16 @@ def forward(
...
         self.vlog(3, "atten.v_proj=%s", v_proj.sum())
+
+        # Scale query and key.
+        key_positions = jnp.arange(k_proj.shape[1])[None, :]
+        q_proj, k_proj = self._scale_qk(
+            q_proj=q_proj,
+            k_proj=k_proj,
+            query_positions=query_positions,
+            key_positions=key_positions,
+        )
```
ruomingp, line 225: Is there a unittest that can be added?

ruomingp (commented):
Thanks! Please also see my comments on the internal PR.

axlearn/vision/attention.py
changlan: Done - I added a unit test.

changlan:
Gentle ping @ruomingp

ruomingp (approved):

PR #1155, opened 2025-05-05
Goodput Monitoring | Add support for custom badput events

Perseus14:
- Goodput monitoring supports custom badput types that are synchronous and overlapping.
- Axlearn can now monitor two custom badput events (SDC_CHECK and EVAL)

### DIFF
--- a/axlearn/common/trainer.py
+++ b/axlearn/common/trainer.py
@@ -1037,14 +1037,23 @@ class SpmdTrainer(Module):
             mesh_shape=cfg.mesh_shape, mesh_axis_names=cfg.mesh_axis_names, device_kind=device_kind
         )
         if not with_xsc:
+            self._maybe_record_event(
+                measurement.Event.START_CUSTOM_BADPUT_EVENT,
+                custom_badput_event_type="COMPILATION_NO_XSC",
+            )
             self._compiled_train_step = self.compile_train_step(
                 trainer_state=trainer_state, input_batch=input_batch, compiler_options=options
+            )
+            self._maybe_record_event(
+                measurement.Event.END_CUSTOM_BADPUT_EVENT,
+                custom_badput_event_type="COMPILATION_NO_XSC",
             )
             return self._compiled_train_step
         logging.log_first_n(logging.INFO, "Compiling XSC train step.", 1)
 
         self._maybe_record_event(
-            measurement.Event.START_CUSTOM_BADPUT_EVENT, custom_badput_event_type="SDC_COMPILATION"
+            measurement.Event.START_CUSTOM_BADPUT_EVENT,
+            custom_badput_event_type="COMPILATION_WITH_XSC",
         )
         compiled_jit_train_step_fn = self.compile_train_step(
             trainer_state=trainer_state,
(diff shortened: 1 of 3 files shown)


axlearn/common/trainer.py
```diff
@@ -1043,6 +1043,9 @@ def _get_compiled_train_step_fn(
             return self._compiled_train_step
         logging.log_first_n(logging.INFO, "Compiling XSC train step.", 1)
 
+        self._maybe_record_event(
+            measurement.Event.START_CUSTOM_BADPUT_EVENT, custom_badput_event_type="SDC_CHECK"
```
Ethanlm, line 1047: Here is more about compiling with SDC check. How can we tell if the event is specific to SDC check?

dipannita08: My idea was to check if the [xsc_policy_check](https://github.com/apple/axlearn/blob/121f43206facb2c9a9a524dbf5774c7d07bfdffb/axlearn/common/trainer.py#L170) is enabled before recording this. Perhaps inside `compiled_jit_train_step_fn`?

dipannita08 (approved):
Approving. @Ethanlm, would you be able to help find the best placement for the recorder API calls for SDC compilation? Thank you!

We have internally tested functionality E2E.

axlearn/common/trainer.py
Perseus14: Thanks for the info @Ethanlm. To identify SDC check, we need to identify and record when it starts and ends as custom badput events. Looks like we might have to refactor the existing code to separate out `with_xsc` and `not with_xsc` training to identify SDC? Any recommendation here?

Ethanlm: sorry for the confusion. I meant this change looks like a recording of compilation time spent on compilation with SDC check enabled. But not the time spent on SDC check itself. 

If we are looking at compilation itself, does it make sense to use `SDC_CHECK_COMPILATION` instead?

dipannita08: I don't think its possible to determine exact start and end times of the SDC check from the application layer; we might need platform signals for that. The goal here is to measure "compilation with SDC enabled", so we should update the name to reflect that imo.

dipannita08: Thanks @Perseus14! @Ethanlm, PTAL :)

dipannita08 (approved):

axlearn/common/trainer.py
Ethanlm: My apologies, I think `COMPILATION_WITH_XSC` is probably a better name. In next PRs, we can also add `COMPILATION_NO_XSC` which is without XSC. What do you think?

Perseus14: The `COMPILATION_NO_XSC` would come under `if not with_xsc` portion of the code ride? I can add it in this PR

Ethanlm: yes thanks

findmyway (approved):

Ethanlm (approved):

Perseus14:
@markblee Could you also PTAL?

markblee (approved):

PR #1163, opened 2025-05-06
Pathways: pass XLA flags correctly

samos123:
Pathways requires XLA flags to be passed to the pathways-proxy and MXLA flags need to be set on each worker.

It also allows overriding or adding more XLA flags by passing `--pathways_xla_flags="xla_tpu_x=24,megascale_y=true"`.

Pathways currently also only supports passing the flags as command line arguments instead of passing them as environment variables.

I verified the XLA flags are effective by looking at step time for Fuji v2 7B on v6e-16:
* Steptime in pathways without XLA flags being passed correctly: 2.259768711319994 seconds
* Steptime in pathways with XLA flags being passed correctly: 2.1042977840899404 seconds
* Steptime in McJax with XLA flags being passed correctly:: 2.087459141649306 seconds

Also tested flags are being set correct by checking the head pod pathways-proxy container manually:
```
axlearn gcp launch run --cluster=$CLUSTER \
        --runner_name gke_tpu_pathways --pathways_xla_flags=xla_jf_crs_combiner_threshold_count=11 \
        --name=$USER \
        --instance_type=tpu-v6e-16 \
        --num_replicas=1 \
        --bundler_spec=allow_dirty=True \
        --bundler_spec=dockerfile=Dockerfile --bundler_spec=target=tpu \
        -- sleep infinity;
```


axlearn/common/compiler_options.py
```diff
@@ -163,6 +163,18 @@ def default_xla_options(
...
+def get_megascale_options(
+    xla_options: dict[str, Union[str, bool, int]]
+) -> dict[str, Union[str, bool, int]]:
+    return {k: v for k, v in xla_options.items() if k.startswith("megascale")}
+
+
+def get_xla_options(
+    xla_options: dict[str, Union[str, bool, int]]
+) -> dict[str, Union[str, bool, int]]:
+    return {k: v for k, v in xla_options.items() if not k.startswith("megascale")}
```
muyangyuapple, line 175: It's safer to write this condition as:

```
return {k: v for k, v in xla_options.items() if not k.startswith("xla") or k.startwith("tpu")}
```

Otherwise if one new none-xla flags are added as "foo_bar", they will be added to this dictionary too.

samos123: Updated and also removed `tpu_use_continuations="true",` so the if statement is cleaner. The default value is tpu_use_continuations="true" so there is no point in setting it in compiler_options.py.

muyangyuapple (approved):

axlearn/cloud/gcp/pathways_utils.py
ruomingp, line 1: Is there a test that can be added for the changes to this file?

ruomingp (commented):
Thanks!

axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -133,6 +139,15 @@ def __init__(self, cfg: BaseReplicatedJob.Config, *, bundler: Bundler):
...
         self._is_single_head = True
+        xla_and_mxla_options = default_xla_options(
+            instance_type=self._tpu_type,
+            num_slices=cfg.inner.accelerator.num_replicas,
+            backend="tpu",
+        )
+        # Needs to be passed to pathways-proxy.
+        self._xla_options = get_xla_options(xla_and_mxla_options)
+        # Needs to be passed as command arguments to each pathways-worker.
+        self._mxla_options = get_megascale_options(xla_and_mxla_options)
```
Ethanlm, line 150: Will megascale flags work even if in a single-slice setup?

Ethanlm (commented):
Thanks

axlearn/cloud/gcp/pathways_utils.py
samos123: It doesn't look like we set any megascale flags when using single slice so it's essentially a no-op for that use case. It would return an empty list and no new arguments are added.

samos123: I take that back, on v6e we set megascale flags irrelevant of whether it's single slice or multi-slice:
```
    if version == "v6e":
        options.update(
            # Change to 16GB. The default is 4GB which is too small for larger models. This
            # cause the step time to be double. You should increase this
            # further if you see "Allocator failed to allocate". A feature
            # to dynamically allocate may come later: b/380514965
            megascale_grpc_premap_memory_bytes=17179869184,
```

note that I verified that this doesnt cause any issues when using v6e-16

axlearn/cloud/gcp/pathways_utils.py
samos123: done

Ethanlm (approved):

samos123:
I think the error from CI is not related to my PR? https://github.com/apple/axlearn/actions/runs/14895076691/job/41907295505?pr=1163#step:8:9480

```
#22 475.0 ==================================== ERRORS ====================================
#22 475.0 _______ ERROR collecting axlearn/open_api/metrics/code_contests_test.py ________
#22 475.0 ImportError while importing test module '/root/axlearn/open_api/metrics/code_contests_test.py'.
#22 475.0 Hint: make sure your test modules/packages have valid Python names.
#22 475.0 Traceback:
#22 475.0 /usr/local/lib/python3.10/importlib/__init__.py:126: in import_module
#22 475.0     return _bootstrap._gcd_import(name[level:], package, level)
#22 475.0 axlearn/open_api/metrics/code_contests_test.py:18: in <module>
#22 475.0     from axlearn.open_api.metrics import code_contests
#22 475.0 axlearn/open_api/metrics/code_contests.py:14: in <module>
#22 475.0     from axlearn.open_api.common import (
#22 475.0 axlearn/open_api/common.py:27: in <module>
#22 475.0     from openai.types.chat.chat_completion_message import ChatCompletionMessage
#22 475.0 E   ModuleNotFoundError: No module named 'openai'
#22 475.0 -------------- generated xml file: /root/test-results/testing.xml --------------
```

@Ethanlm Thanks for reviewing and approving! Do we need anyone else to approve this?

axlearn/common/compiler_options.py
```diff
@@ -123,7 +123,6 @@ def default_xla_options(
             xla_should_allow_loop_variant_parameter_in_chain="true",
             xla_should_add_loop_invariant_op_in_chain="true",
             xla_tpu_use_enhanced_launch_barrier="true",
-            tpu_use_continuations="true",
```
markblee, line 126: Intended?

axlearn/common/compiler_options.py
```diff
@@ -163,6 +162,18 @@ def default_xla_options(
     return options
 
 
+def get_megascale_options(
+    xla_options: dict[str, Union[str, bool, int]]
+) -> dict[str, Union[str, bool, int]]:
```
markblee, line 167: Add a docstring for public fns here and below?

markblee (approved):

axlearn/common/compiler_options.py
samos123: yes intended, the default has always been true since it was introduced in 2020 and this flag doesn't follow the expected naming scheme that XLA flags start with `xla`. So there is no point in keeping it around. This allows the get_megascale_options and get_xla_options functions to be simpler.

axlearn/common/compiler_options.py
samos123: done for both

samos123:
@markblee could you review again please? I've added the docstring and responded to your comment about why one of the XLA flags was removed intentionally.

markblee (approved):

muyangyuapple (approved):

raytomato (approved):

Ethanlm:
I will test it in our internal env as a final validation

samos123:
Chatted with @Ethanlm the issue with the current PR is that it doesn't allow users to modify XLA flags in axlearn code. Please hold off with merging for now. I will discuss with Shaurya on how to address.

samos123:
@Ethanlm can you please review the PR again now that I've added the ability to add and override any of the existing XLA flags?

axlearn/common/compiler_options.py
```diff
@@ -163,6 +162,49 @@ def default_xla_options(
...
+
+    Returns:
+        A dictionary containing only Megascale-related XLA options
+        (those starting with 'megascale').
+    """
+    return {k: v for k, v in xla_options.items() if k.startswith("megascale")}
+
+
+def get_xla_options(
+    xla_options: dict[str, Union[str, bool, int]]
```
lukebaumann, line 181: nit: rename to `xla_and_mxla_options`

samos123: tried to be as specific as possible since people may not know that mxla refers to flags that have megascale in them. I originally did know that to be the case. I prefer using full name unless it gets too long.

lukebaumann: This is already done in [axlearn/cloud/gcp/pathways_utils.py](https://github.com/apple/axlearn/pull/1163/files/23e4d9df4c6a86c07c7a477438d3bfaf3883b044#diff-39e195f06f7b4ff901b248d73fbbb65b7aa44dd226a557413017fcb0adf7a243) and I am suggesting naming the argument with both. If I had no context, I might wonder why I need to call `get_xla_options(xla_options)`.

samos123: this function only returns the xla_options though and `default_xla_options` returns both XLA and MXLA flags. The issue is that I probably shouldn't rename `default_xla_options` which I expect to cause downstream issues to existing AXLearn users who may rely in private model repos on `default_xla_options`.

@Ethanlm would it be safe to rename `default_xla_options` to `default_xla_and_mxla_options`?

Ethanlm: yes, I think we can change it. 
But our internal repo does depend on this change. Maybe we do it in a separate PR, so it is easier to revert if anything goes wrong

Ethanlm (approved):
Internal CI passed. I also did a sanity check within our internal env with small jobs. They worked as expected.

samos123:
@markblee could I get a final review from you so we can get it merged?

axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -127,12 +146,36 @@ def __init__(self, cfg: BaseReplicatedJob.Config, *, bundler: Bundler):
...
+            num_slices=cfg.inner.accelerator.num_replicas,
+            backend="tpu",
+        )
+        pathways_xla_flags = {}
+        for flag in pathways_cfg.pathways_xla_flags:
+            if "=" in flag:
+                k, v = flag.split("=", 1)
+                # Users may accidentally pass flags as args
+                k = k.lstrip("--")
+                pathways_xla_flags[k] = parse_xla_flag_value(v.strip())
```
markblee, line 167: We have a util `parse_kv_flags` that does more checks (like making sure that we use maxsplit=1 so that `a=b=c` doesn't break here).

axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -92,13 +99,25 @@ class Config(BaseReplicatedJob.Config):
...
         super().define_flags(fv)
         common_kwargs = dict(flag_values=fv, allow_override=True)
+        # XLA flags and megascale flags have to be passed at jobset creation time.
+        # The XLA flags automatically get passed to the pathways proxy and the
+        # Megascale flags get passed to pathways workers. A single flag is used since
+        # the implementation details could change later.
+        flags.DEFINE_list(
+            "pathways_xla_flags",
+            [],
+            "Set XLA and Megascale flags. Defaults are set by compiler_options.py. "
```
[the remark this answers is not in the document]
samos123: You can also set new flags that are not set by default that's why I slightly prefer the Set XLA and Megascale flags wording. No hard opinion though, happy to use your wording.

axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -127,12 +146,36 @@ def __init__(self, cfg: BaseReplicatedJob.Config, *, bundler: Bundler):
         super().__init__(cfg, bundler=bundler)
         self._bundler = bundler
         self._inner: TPUReplicatedJob = cfg.inner.instantiate(bundler=self._bundler)
+        pathways_cfg: PathwaysReplicatedJob.Config = self.config
```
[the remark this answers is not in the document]
samos123: the issue is that `cfg: BaseReplicatedJob.Config` is passed as an arg to `__init__(`. So that's why we have to alias.

axlearn/cloud/gcp/pathways_utils.py
samos123: Nice catch. Didn't know AXLearn had that library. Fixed.

axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -92,13 +99,25 @@ class Config(BaseReplicatedJob.Config):
         """
 
         inner: Required[TPUReplicatedJob.Config] = REQUIRED
+        pathways_xla_flags: Optional[list[str]] = []
```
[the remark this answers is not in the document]
samos123: Fixed

samos123:
@markblee could you review once more please? I've addressed all your comments.

axlearn/cloud/gcp/pathways_utils.py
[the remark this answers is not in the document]
markblee: Oh, I meant the alias of the variable name. Also, I think the annotation is wrong in `__init__`. Anyway, not blocking on this since it's not due to this PR.

axlearn/common/compiler_options.py
```diff
@@ -163,6 +162,49 @@ def default_xla_options(
...
+        A dictionary containing only XLA-specific options (those starting with 'xla').
+    """
+    return {k: v for k, v in xla_options.items() if k.startswith("xla")}
+
+
+def parse_xla_flag_value(value: str) -> Union[int, bool, str]:
+    """
+    Attempts to convert an XLA flag string value to int, then bool.
+    If conversion fails, returns the original string (stripped).
+    """
```
markblee, line 198:
```suggestion
    """Attempts to convert an XLA flag string value to int, then bool.
    If conversion fails, returns the original string (stripped).
    """
```

Do we need this util public here? Can we move it to `pathways_utils` for now as `_parse_xla_flag_value` to minimize the public API?

markblee (approved):
Mostly LGTM.

axlearn/common/compiler_options.py
samos123: done

markblee:
> @markblee could you please review once more? I moved the get_xla_options and get_megascale_options to pathways_util.py. Nothing else changed. I also re-ran my manual test.

Thanks, I still see them in `compiler_options`, were they accidentally copied?

axlearn/common/compiler_options.py
```diff
@@ -163,6 +162,20 @@ def default_xla_options(
     return options
 
 
+def parse_xla_flag_value(value: str) -> Union[int, bool, str]:
```
samos123, line 165: removing this function as well

samos123:
Good catch! Apologies for somehow messing up the deletion in compiler_options.py.

I reviewed the final changes and also moved the parse_xla_flag function to pathways_utils.py. Please review again once more.

axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -59,6 +64,20 @@
 _PATHWAYS_HEAD_NODE_POOL_SELECTOR_VALUE = "workload"
 
 
+def parse_xla_flag_value(value: str) -> Union[int, bool, str]:
+    """
+    Attempts to convert an XLA flag string value to int, then bool.
+    If conversion fails, returns the original string (stripped).
+    """
```
ruomingp, line 71:
```suggestion
    """Attempts to convert an XLA flag string value to int, then bool.
    
    If conversion fails, returns the original string (stripped).
    """
```

axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -78,6 +97,35 @@ def get_pathways_tpu_version(gke_machine_type: str) -> str:
...
+) -> dict[str, Union[str, bool, int]]:
+    """Filters XLA options for those starting with 'xla_'.
+
+    Args:
+        xla_options: A dictionary of XLA options.
+
+    Returns:
+        A dictionary containing only XLA-specific options (those starting with 'xla').
+    """
+    return {k: v for k, v in xla_options.items() if k.startswith("xla")}
```
ruomingp, line 126:
```suggestion
    return {k: v for k, v in xla_options.items() if k.startswith("xla_")}
```

ruomingp (commented):
Thanks! Will defer to @markblee for approval.

axlearn/cloud/gcp/pathways_utils.py
samos123: done

axlearn/cloud/gcp/pathways_utils.py
samos123: done

markblee (approved):

samos123:
Seems the error is unrelated to my PR: https://github.com/apple/axlearn/actions/runs/15166874891/job/42647986834?pr=1163#step:8:5949
```
#22 155.3 /opt/venv/lib/python3.10/site-packages/transformers/integrations/tensor_parallel.py:465: in __init__
#22 155.3     self.input_layouts = (input_layouts or Replicate(),)
#22 155.3 E   NameError: name 'Replicate' is not defined
```

samos123:
Merged with latest main in the hopes that it would fix the CI issue. Could you re-trigger? 

Specifically #1205 may fix the CI failure in my run.

samos123:
Failure seems unrelated to my PR:
```
The hosted runner: GitHub Actions 314 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.
```
@markblee not sure if there is anything else I should do.

PR #1165, opened 2025-05-07
[SummaryWriter] Fix escaped unit test

yuanliu939:
One of the unit test escaped after https://github.com/apple/axlearn/pull/1153. Since we no longer log a list of Audio, the unit test needs to be adjusted accordingly.


apghml (commented):
Could you update the pr description to explain what issue this fixes and why some of the other asserts are deleted?

apghml (approved):

ds-hwang (approved):

ds-hwang:
CI unrelated failure in axlearn/open_api/metrics/code_contests_test.py.
Olli is going to fix it.

PR #1168, opened 2025-05-08
Apply logits modifier during inference to match training behavior

ds-hwang:
This PR ensures that the logits_modifier is consistently applied during inference, aligning it with the behavior used during training. Without this, the model's output distribution at inference time may deviate from what it was optimized for, particularly when the modifier introduces biases, token masking, or length conditioning. This change preserves the intended semantics of the model across both training and evaluation.

For example, in https://arxiv.org/abs/2203.03466, Table 8&9, there is a multiplier on the output weights. It should be applied for all uses of the output weights.

### DIFF
--- a/axlearn/common/decoder.py
+++ b/axlearn/common/decoder.py
@@ -554,6 +554,8 @@ class Decoder(BaseLayer):
             # Reuse the token embedding.
             with child_context("emb_attend", module=self.emb):
                 logits = self.emb.attend(x)
+        if self._output_logits_modifier is not None:
+            logits = self._output_logits_modifier(logits)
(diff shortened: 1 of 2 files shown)


ds-hwang:
@ruomingp could you review it? From 1282

apghml (approved):

ds-hwang:
CI unrelated failure in axlearn/open_api/metrics/code_contests_test.py.
Olli is going to fix it.

ds-hwang:
@markblee @ruomingp Could you take a look?

ds-hwang:
@markblee @ruomingp Could you take a look?

markblee (approved):

PR #1169, opened 2025-05-08
Update Dockerfile to use ubuntu:22.04

andersensam:
This PR aims to update the Dockerfile to use a base image compatible with CUDA 12.8 and eliminate `pip` caching, greatly reducing the size of GPU and TPU targets.

In testing, when compared to using NVIDIA base images, this approach saves ~7GB per image for GPU targets.

Libraries only required for GPU, such as `libnvrtc` are installed as part of the GPU target only.

Switching to `uv` from `pip` greatly speeds up dependency resolution and further modernizes this image.


Dockerfile
```diff
@@ -26,11 +26,10 @@ COPY pyproject.toml pyproject.toml
 RUN mkdir axlearn && touch axlearn/__init__.py
 # Setup venv to suppress pip warnings.
 ENV VIRTUAL_ENV=/opt/venv
-RUN python -m venv $VIRTUAL_ENV
+RUN python3 -m venv $VIRTUAL_ENV
```
samos123, line 29: Would it be benificial to manage the `venv` with `uv` as well? I think it could be confusing since we now may have 2 different venvs? one managed by uv within `axlearn/.venv` and one `/opt/venv`.

Some users may run `python` directly within the container and expect that to point to the uv managed venv. It would be nice if the `uv` managed venv becomes the python env for everything within the container.

andersensam: Using `uv` here doesn't automatically create another virtual environment and since we have installed `uv` inside the `venv` already, it's accessible inside the container

samos123: Oh nice! I didn't know `uv` could be used within an existing venv like that. Please close this conversation.

samos123 (approved):

Dockerfile
```diff
@@ -101,7 +100,12 @@ FROM base AS gpu
 
 # TODO(markblee): Support extras.
 ENV PIP_FIND_LINKS=https://storage.googleapis.com/jax-releases/jax_cuda_releases.html
-RUN pip install .[core,gpu]
+# Enable the CUDA repository and install the required libraries (libnvrtc.so)
+RUN curl -o cuda-keyring_1.1-1_all.deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/cuda-keyring_1.1-1_all.deb && \
+    dpkg -i cuda-keyring_1.1-1_all.deb && \
+    apt-get update && apt-get install -y cuda-libraries-dev-12-8 ibverbs-utils && \
+    apt clean -y
```
markblee, line 107: Why was this previously not necessary?

markblee (commented):
Thanks! Mostly LGTM, pending internal testing.

Dockerfile
andersensam: Newer versions of Jax (like 0.5.3) have some requirements on `libnvrtc.so` that were not present previously. It seems that there isn't any pip package that includes the library, so manual install is needed. Using the official CUDA image gets around this, but then the image is much larger

andersensam:
Updated to use `UV_FIND_LINKS` instead of `PIP_FIND_LINKS` and enabled `--prerelease=allow` to account for `libtpu` nightly versions

markblee (approved):

ruomingp (approved):

PR #1171, opened 2025-05-09
Use `remat` in `scan_in_context`

ds-hwang:
Applying `scan(remat(scan_fn))` is a well-known JAX idiom. It enables gradient checkpointing at scan function boundaries, which significantly reduces memory usage during training.

We measured the impact on a 2.8B model (using `RepeatedTransformer`). The change saved a substantial amount of memory with minimal effect on training speed:

* **Before (as-is)**
  * `train_step`: 19,558.35 MiB, 1112 ms
  * `forward`: 8,433.40 MiB, 295 ms

image

* **After (with remat)**
  * `train_step`: 15,695.13 MiB, 1128 ms
  * `forward`: 8,433.40 MiB, 295 ms

image

Note: The relationship between BaseLayer's remat, and Repeat's scan(remat(fn))
can be described as: remat(scan(remat(fn)), policy)

Until now, the inner scan was opaque to remat, so the remat policy wasn’t properly applied inside it (e.g., in RepeatedTransformerLayer). This PR applies the remat policy inside the scan as well: remat(scan(remat(fn, policy)), policy)

### DIFF
--- a/axlearn/common/base_layer.py
+++ b/axlearn/common/base_layer.py
@@ -338,7 +338,7 @@ class BaseLayer(Module):
             # pylint: disable-next=protected-access
             module._remat_methods.append(method_fn.__name__)
             # Pass both outputs and output_collection through remat(...) to avoid leaking tracers.
-            outputs, output_collection = jax.ad_checkpoint.remat(
+            outputs, output_collection = jax.checkpoint(
                 fn,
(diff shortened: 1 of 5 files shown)


ds-hwang:
@ruomingp Could you review it? From 1288

ruomingp (approved):

PR #1175, opened 2025-05-12
Add grain to core

RsEnts:


markblee (approved):
LG once internal passes.

PR #1177, opened 2025-05-12
Introduce safe_not() for ~paddings instead of 1 - paddings.

ds-hwang:
Now all of paddings.dtype are bool (1bit), after
https://github.com/apple/axlearn/pull/1133

This PR is unifying coding style between ~paddings and 1 - paddings used inconsistently across the codebase. I’d like to standardize on ~paddings, because A 1-bit flip is cheaper than a subtraction.

In detail, the common operation `x * (1 - paddings)` involves 4 steps:

1. Cast `paddings` to `int32` to perform `1 - paddings`
2. Perform the subtraction: `1 - paddings`
3. Cast the result to `float32` to match `x`
4. Perform the float multiplication: `x * (1 - paddings)`

This introduces an unnecessary cast to `int32`.
In contrast, `x * ~paddings` only requires 3 steps:

1. Bitwise NOT (`~paddings`) — a cheap bit flip
2. Cast the result to `float32` to match `x`
3. Multiply `x * (~paddings)`

Also, `float * bool` (2 and 3 steps) may (will) be specially optimized in XLA or hardware, making this path potentially more efficient, while `float * int` is hard to be optimized as internal value matters.

### DIFF
--- a/axlearn/audio/decoder_asr.py
+++ b/axlearn/audio/decoder_asr.py
@@ -8,6 +8,7 @@
 
 from typing import Callable, Optional, Union
 
+import chex
 import jax
 import jax.numpy as jnp
 import optax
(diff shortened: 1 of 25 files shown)


ds-hwang:
@ruomingp Could you take a look? From 1299

ruomingp (approved):

PR #1180, opened 2025-05-13
Add `write_every_n_steps_map` to SummaryWriter for fine-grained logging control

ds-hwang:
**Summary**

This PR enhances `SummaryWriter` by introducing support for `write_every_n_steps_map`, allowing users to configure different logging frequencies for different summary types (e.g., scalar, image, audio).

Previously, `write_every_n_steps` applied uniformly to all summary types, which made logging media like images or audio at a lower frequency difficult without also reducing scalar logging granularity.

**Key Changes**

* Introduced a new config field:

  ```
  write_every_n_steps_map: Optional[dict[SummaryKind, int]]
  ```

  which overrides `write_every_n_steps` for specific `SummaryKind` values.

* Added `SummaryKind` literal to classify summary types (`Scalar`, `Tensor`, `Text`, `Image`, `Audio`).

* Replaced hardcoded media write logic with `_time_to_write(step, SummaryKind)` check.

* Preserved backward compatibility: if `write_every_n_steps_map` is not set, behavior is unchanged.

**Example Usage**

```
SummaryWriter.Config(
    write_every_n_steps=100,
    write_every_n_steps_map={
        "Image": 1000,
        "Audio": 5000,
    }
)
```

This would log scalars and text every 100 steps, but images only every 1,000 and audio every 5,000 steps.

**Motivation**

Logging audio/image too frequently can slow down training or clutter the UI, while scalar metrics often need higher resolution. This change provides more flexible control over logging behavior without disrupting existing workflows.

### DIFF
--- a/axlearn/common/summary_writer.py
+++ b/axlearn/common/summary_writer.py
@@ -6,10 +6,9 @@ import contextlib
 import enum
 import numbers
 import os
-from collections.abc import Sequence
 from functools import wraps
 from types import FunctionType
-from typing import Any, Callable, Optional, Union
+from typing import Any, Callable, Literal, Optional, Sequence, Union
 
 import jax
 import numpy as np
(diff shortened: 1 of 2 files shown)


ds-hwang:
@markblee Could you take a look? From 1281

markblee (approved):

PR #1190, opened 2025-05-18
Add logging for dataset input and checkpoint saving

linxiulei:
This PR adds:

* logging utility classes to log when an operation exceeds a specified timeout while still running or its total execution time exceeds a specified threshold.
* use mentioned utility classes to around dataset input fetch (per iteration) and checkpoint saving to allow us to observe either execution being stalled or in poor performance


axlearn/common/trainer.py
```diff
@@ -1091,8 +1097,9 @@ def _run_step(
             train_summaries=outputs["summaries"], force_runs=force_run_evals
         )
 
-        # Checkpointer policy will decide if we should save.
-        self.save_checkpoint(evaler_summaries=evaler_summaries)
+        with LoggingContext("save_checkpoint", timeout=180, threshold=120):
```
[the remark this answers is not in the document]
linxiulei: Sure, but I am thinking which way is better to make it configurable, env var or flags?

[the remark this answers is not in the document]
apghml: The typical convention we use is that if we want to make the child configurable, then the parent config should contain a config for the child.

axlearn/common/logging_utils.py
```diff
@@ -0,0 +1,138 @@
+# Copyright © 2025 Apple Inc.
+
+"""Logging utility classes"""
+
+import logging
+import threading
+import time
+
+
+class LoggingContext:
```
apghml, line 10: Could you explain the difference of this from the other monitor functionality that exists like GPUMonitor, etc? If they serve overlapping purposes, should they be unified?

axlearn/common/trainer.py
```diff
@@ -1091,8 +1097,9 @@ def _run_step(
...
+            # Checkpointer policy will decide if we should save.
+            self.save_checkpoint(evaler_summaries=evaler_summaries)
```
apghml, line 1102: Since checkpoint syncing is asynchronous, do we expect to see timeouts inside this block?

apghml (changes requested):

axlearn/common/trainer.py
```diff
@@ -581,7 +582,12 @@ def run(
                 output = None
                 stop_trace_step = None
 
-                input_iterator = self.input.batches(self._input_iter)
+                input_iterator = LoggingIterator(
```
Ethanlm, line 585: Since it is already recorded in the Goodput, I wonder if we need a separate logging here. 

If I am reading it correctly, this LoggingIterator creates one timer thread every time `next` is called, which seems to be adding a lot of thread creation/cancellation overhead to the main process

axlearn/common/logging_utils.py
linxiulei: Can you please clarify what other monitor functionality we already have? If they overlap, we can unify them

axlearn/common/trainer.py
linxiulei: Yes, IIUC checkpoint saving is asynchronous if the previous saving is complete or it may stuck in https://github.com/apple/axlearn/blob/main/axlearn/common/array_serialization.py#L554

axlearn/common/trainer.py
linxiulei: Assuming you mean this code
```
                    self._maybe_record_event(measurement.Event.START_DATA_LOADING)
                    try:
                        input_batch = next(input_iterator)
                        self._maybe_record_event(measurement.Event.END_DATA_LOADING)
```

If the next() somehow hangs and then the whole process aborts, would Goodput measurement help catch that?

My understanding is this iterator invokes next() every a few seconds depending on the training scale, so creating and cleaning a thread for every a few seconds should have only very minimal overhead (empirically one thread takes 10 to 100 usec to create).

However, if you still have concerns on overhead, we might create a reusable thread for timer purpose.

axlearn/common/logging_utils.py
apghml: TfSummaryMonitor, GPUMonitor, GoodputMonitor. I might be forgetting some additional ones.

axlearn/common/logging_utils.py
```diff
@@ -0,0 +1,138 @@
...
+  """A context manager for monitoring the execution time of operations and
+  logging warnings based on defined timeouts and thresholds.
+
+  This context manager can be used to:
+  - Log a warning if an operation exceeds a specified timeout while still
+  running.
+  - Log a warning if an operation's total execution time exceeds a specified
+  threshold.
```
ruomingp, line 18: It's not clear that we need logging-upon-timeout, given the threading concerns raised by @Ethanlm. I wonder if a simpler semantic would be sufficient:

- LoggingContext always wait until the op finishes and log a warning if the execution time exceeds the threshold;
- We rely on the watchdog https://github.com/apple/axlearn/blob/2c4ec6ea939a12b6752ae3f95fbc4fcc58388dad/axlearn/common/trainer.py#L190 to tell us where the threads are stuck

WDYT?

linxiulei: I agree with starting something simpler so I changed this PR to only implement logging upon op completion. 

However, I feel logging-upon-timeout is important because program hangs are difficult to debug by just looking through tracebacks of so many workers printed by watchdog timeout. A clear logging message may help identify workers that cause the hangs.

axlearn/common/trainer.py
```diff
@@ -1091,8 +1093,9 @@ def _run_step(
             train_summaries=outputs["summaries"], force_runs=force_run_evals
         )
 
-        # Checkpointer policy will decide if we should save.
-        self.save_checkpoint(evaler_summaries=evaler_summaries)
+        with LoggingContext("save_checkpoint", threshold=120):
```
Ethanlm, line 1096: Can we make it configurable? 
We should be able to patch this ad-hoc in jobs we want to debug and monitor closely. But the hard-coded 120s here makes it less flexible to apply in general jobs, hence not a good candidate for PR merge

PR #1195, opened 2025-05-20
Fix CI pre-commit check failure.

raytomato:
Started to see consistent CI failure starting from last night.

The error is a sigterm with pylint
```
pylint...................................................................
Error: Process completed with exit code 143.
```

**Cause**
Initially, I thought it might be worker machine preemption, but it fails consistently across many machines, so preemption shouldn't be the reason.

Then I tried to see if OOM could be the reason, and was able to prove my assumption.
1. I first kept decreasing parallelism by changing `jobs=` in `.pylintrc`. Was able to get it working occasionally with 2 and consistently with 1.
2. Then I added a monitor process, we can see that even with 1 process, the peak memory usage is around 75%.

Thus, I was able to identify the cause.

**Implication of reduced parallelism**
While the memory impact is very obvious as I decrease the parallelism, speed is not really impacted that much.
The most recent successful run from yesterday (https://github.com/apple/axlearn/actions/runs/15127200091/job/42521313140) finished pre-commit stage in 5min and 6second, my test runs can finish within 5mins 30 seconds.

**Root cause**
I was not able to identify what triggered this change. One guess I have is that, maybe we were using a more powerful CI instances.
Since there is not much CI speed regression, I think we can prioritize getting CI fixed first.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -16,5 +16,18 @@ jobs:
       # TODO(markblee): Remove gcp,vertexai_tensorboard from CI. (needed by pytype)
       - run: pip install '.[core,dev,grain,gcp,vertexai_tensorboard,open_api]'
       # pylint uses approx 12GB of memory during this run, look into split to decrease?
-      - run: pre-commit run --all-files
+      - run: |
+          .github/scripts/monitor_memory.sh &
+          MONITOR_PID=$!
+          # A short sleep to wait for monitor process start running.
+          sleep 1
+
+          echo "====== Starting pre-commit... ======"
+          pre-commit run --all-files
+          echo "====== pre-commi complted. ======"
+
+          if kill -0 $MONITOR_PID 2>/dev/null; then
+            echo "Manually stopping monitor..."
+            kill $MONITOR_PID
+          fi
       - run: pytype -j auto .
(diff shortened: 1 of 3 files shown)


samos123:
Seems this is needed for my pr too #1163

raytomato:
> Seems this is needed for my pr too #1163

Yeah, I think CI is blocked. I know the issue now. Will try to get a proper version that we can merge tonight.

samos123:
I also tried running precommit on my Macbook but pylint causes hangs when it tries to run it for all files. It's not a full hang but everything becomes incredibly slow, making my laptop almost unusable. Probably won't help much in your investigation but sharing just in case.

.github/workflows/pre-commit.yml
```diff
@@ -16,5 +16,21 @@ jobs:
...
+          # Start memory monitor as a background process.
+          .github/scripts/monitor_memory.sh &
+          MONITOR_PID=$!
+          # A short sleep to wait for monitor process to start.
+          sleep 1
+
+          # Start pre-commit check.
+          echo "====== Starting pre-commit... ======"
+          pre-commit run --all-files
+          echo "====== pre-commi complted. ======"
```
liboyue, line 29:
```suggestion
          echo "====== pre-commit completed. ======"
```

liboyue (approved):
LGTM. I suggest to only include `.pylintrc` change for simplicity.

raytomato:
> LGTM. I suggest to only include `.pylintrc` change for simplicity.

I feel it might be good to track it?
But if we don't want to merge the monitor code in a rush, I can leave it to a separate PR.

I'm open to either way.

markblee (approved):

kelvin-zou:
Thank you @raytomato , great work!

PR #1198, opened 2025-05-21
[UPDATE JAX API] Update `jax.experimental.array_serialization.serialization` to latest JAX

Steboss:
In the latest JAX version there's been a modification of the behaviour of `jax.experiment.array_serialization.serialization`  where the attributes: 
- `_spec_has_metadata`
- `ts`
have been transferred to `ts_impl` ([here](https://github.com/jax-ml/jax/blob/main/jax/experimental/array_serialization/tensorstore_impl.py, )).
Thus I modified the `array_serialization.py` code that uses `serialization` via `ts`

@matthew-e-hopkins for visibility

### DIFF
--- a/axlearn/common/array_serialization.py
+++ b/axlearn/common/array_serialization.py
@@ -252,7 +252,7 @@ async def _async_serialize(
         and arr_inp.is_fully_addressable
     )
     # pylint: disable-next=protected-access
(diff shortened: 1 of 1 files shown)


ruomingp (approved):
@hanzhi713 Are we also trying to upstream our changes into JAX?

PR #1202, opened 2025-05-21
[UPDATE JAX API] Update `jax.tree_map`

Steboss:
As per #1199 here is an update of `jax.tree_map`

### DIFF
--- a/axlearn/common/update_transformation.py
+++ b/axlearn/common/update_transformation.py
@@ -186,7 +186,7 @@ class ConditionalUpdateTransformation(UpdateTransformation):
             return new_updates.delta_updates, new_state
 
         def stop_transform(_):
-            return jax.tree_map(jnp.zeros_like, updates.delta_updates), prev_state
+            return jax.tree_util.tree_map(jnp.zeros_like, updates.delta_updates), prev_state
 
         # We do the computation regardless of the should_update value, so we could have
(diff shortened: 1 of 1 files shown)


axlearn/common/update_transformation.py
```diff
@@ -186,7 +186,7 @@ def real_transform(_):
             return new_updates.delta_updates, new_state
 
         def stop_transform(_):
-            return jax.tree_map(jnp.zeros_like, updates.delta_updates), prev_state
+            return jax.tree_util.tree_map(jnp.zeros_like, updates.delta_updates), prev_state
```
apghml, line 189: Can we consolidate these related changes that are just renames into a single PR?

Also, I thought the outcome of the discussion in your older pr was to always prefer `jax.tree.*` apis over `jax.tree_util` when the former are available?

apghml (changes requested):

ruomingp (approved):

axlearn/common/update_transformation.py
Steboss: @apghml I'll make a single PR that groups all the `tree` related changes - makes sense, sorry about that. 
And from there we can choose if `jax.tree_util` or `jax.tree`. I may be wrong, but I spotted a mix of both in the code

PR #1205, opened 2025-05-21
Quick fix Axlearn CI due to transformers and torch version conflict.

ds-hwang:
transformers>=4.51.3 installs the latest version and conflict with the outdated torch. Before proper torch upgrade, this PR ceases the fire.


ds-hwang:
@ruomingp @markblee Could you review it? From 1350

markblee (approved):

raytomato (approved):
Thanks!

PR #1208, opened 2025-05-23
Implemented sliding window KV cache to maintain the window size KV cache to enable infinite decoding.

ds-hwang:
Currently, when using `MultiheadAttention` or `GroupedQueryAttention` for sliding window attention, the KV cache is kept for the full sequence length (`seq_len`) instead of the window length (`window_len`).

For example, a model with `window_len=1k` and `seq_len=2M` keeps a KV cache for the full 2M tokens. It then biases 1999k invalid KV tokens before calculating attention, resulting in a computational complexity of **O(2M²)** instead of the desired **O(1k²)**.

This issue persists even when using flash attention. Flash attention uses the KV cache allocated in HBM as its input. While unnecessary blocks are discarded during computation, the KV cache still occupies HBM inefficiently for the full 2M tokens.

To address this, when `MultiheadAttention` detects a sliding window mask, it stores the key-value (KV) cache in a ring buffer inside the input linear layer. As a result, downstream projects using `MultiheadAttention` automatically benefit from efficient KV cache handling in `init_states` and `extend_step`.

This PR actually moves from downstream speech/streaming/sliding_window_attention.py

**TODO**: TPU/GPU flash decoding now supports sliding window KV cache. **Note**: During FORWARD and PREFILL, its behavior is identical to vanilla KV cache. Therefore, both Splash Attention and GPU Flash Attention already cover this path in TPU and GPU.
**Note**: `enable_sliding_window_attention` is currently only used in unittests,
 so no model behavior has changed yet.

### DIFF
--- a/axlearn/common/attention.py
+++ b/axlearn/common/attention.py
@@ -99,6 +99,7 @@ from axlearn.common.attention_bias import (
     CausalAttentionBias,
     MaskFnAttentionBias,
     SegmentIdAttentionBias,
+    SlidingWindowAttentionBias,
     as_attention_bias,
     causal_mask,
     make_segment_mask,
@@ -154,6 +155,7 @@ from axlearn.common.utils import (
     flatten_items,
     get_or_none,
     save_and_offload_only_these_names_regex,
+    sequence_mask,
     shapes,
     split_prng_key,
 )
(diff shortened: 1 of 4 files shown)


ds-hwang:
@markblee Could you review it? From 1064

markblee (approved):

PR #1209, opened 2025-05-23
Compute SpecAug in jnp.bool (1 byte), not jnp.int32.

ds-hwang:
It saves the temp memory for a [batch, time, freq] mask.


ds-hwang:
@markblee Could you review it? From 1351

markblee (approved):

PR #1216, opened 2025-05-27
Add max_len arg to fake_speech_source().

ds-hwang:
fake_speech_source() generates extremely short fake audio samples (lengths like 1, 2, 3 and so on), which causes gradients to be zero in gradient unit tests.

### DIFF
--- a/axlearn/audio/spectrum_augmenter_test.py
+++ b/axlearn/audio/spectrum_augmenter_test.py
@@ -217,7 +217,7 @@ class SpectrumAugmenterTest(TestCase):
             dict(inputs=inputs, paddings=paddings), is_training=True, **kwargs
         )
         # pylint: disable-next=import-outside-toplevel
-        import matplotlib.pyplot as plt
+        import matplotlib.pyplot as plt  # pytype: disable=import-error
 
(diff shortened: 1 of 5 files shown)


markblee (approved):

axlearn/audio/spectrum_augmenter_test.py
```diff
@@ -216,7 +216,7 @@ def test_visualize(self, input_shape: Sequence[int], **kwargs):
         outputs = self._generate_masks(
             dict(inputs=inputs, paddings=paddings), is_training=True, **kwargs
         )
-        # pylint: disable-next=import-outside-toplevel
+        # pylint: disable-next=import-outside-toplevel,import-error
         import matplotlib.pyplot as plt
```
markblee, line 220: Do you need `# pytype: disable=import-error`?

markblee (approved):

axlearn/audio/spectrum_augmenter_test.py
ds-hwang: I think `# pylint: disable-next=import-outside-toplevel,import-error` contain ` # pytype: disable=import-error`.
Let me see CI work

markblee (approved):

PR #1219, opened 2025-05-27
axlearn.einops now supports `...`.

ds-hwang:
The initial implementation intentionally did not support `...`, in order to keep the code simple.
However, when trying to adopt axlearn.einops downstream, we found that `...` is used extensively, so we decided to go ahead and add support for it.

In addition, repeat() is extended to allow usage of 1 as a token, since there were downstream cases that required this.


ds-hwang:
Thank you for quick review!

PR #1234, opened 2025-06-02
Fix flash mask computation for jax 0.5.3

hanzhi713:
See code comments


hanzhi713:
@markblee Could you please take a look?

PR #1235, opened 2025-06-03
CausalLM: Metrics now return `WeightedScalar` instead of plain scalars.

ds-hwang:
This is because, in multimodal training, it's often unclear how to assign a common weight to all losses. By returning the loss weight alongside the value, we enable proper loss aggregation downstream.


ds-hwang:
@markblee Could you review? From 1378

markblee (approved):

PR #1241, opened 2025-06-05
Set default PRNG implementation on pathways

samos123:
AXLearn assumes implicitely that rbg is used as the PNRG implementation.

### DIFF
--- a/axlearn/common/launch.py
+++ b/axlearn/common/launch.py
@@ -115,6 +115,12 @@ def setup():
 
     with _init_context():
         if FLAGS.jax_backend == "proxy":
+            # AXLearn assumes rbg PRNG implementation and restore from checkpoint
+            # will fail on pathways if this isn't set. This is due shape of [4]
+            # being hardcoded here:
+            # https://github.com/apple/axlearn/blob/8bb4421e62c815ef9f1ba3679c3277b8bbc6a449/axlearn/common/trainer.py#L330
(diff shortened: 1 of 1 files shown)


muyangyuapple (approved):

axlearn/common/launch.py
```diff
@@ -115,6 +115,12 @@ def setup():
 
     with _init_context():
         if FLAGS.jax_backend == "proxy":
+            # AXLearn assumes rbg PRNG implementation and restore from checkpoint
+            # will fail on pathways if this isn't set. This is due shape of [4]
+            # being hardcoded here:
+            # https://github.com/apple/axlearn/blob/8bb4421e62c815ef9f1ba3679c3277b8bbc6a449/axlearn/common/trainer.py#L330
+            jax.config.update("jax_default_prng_impl", "rbg")
+
```
markblee, line 123: We might want to unify with `setup_spmd` in the future, which already does this (with some minor refactor of the API to decouple from spmd-specific setup).

markblee (approved):

axlearn/common/launch.py
Ethanlm: We can probably add a function in `axlearn/common/utils.py`

Ethanlm: I am merging this PR now to fix the issue. 
Will work with @samos123 on addressing this comment

PR #1250, opened 2025-06-10
Move KVCache classes to separate files

hanzhi713:
This is to avoid circular import if we add a new KV cache layer later and don't want to put it in `attention.py`

### DIFF
--- a/axlearn/common/attention.py
+++ b/axlearn/common/attention.py
@@ -98,7 +98,6 @@ from axlearn.common.attention_bias import (
     CausalAttentionBias,
     MaskFnAttentionBias,
     SegmentIdAttentionBias,
-    SlidingWindowAttentionBias,
     as_attention_bias,
     causal_mask,
     make_segment_mask,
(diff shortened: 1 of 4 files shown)


axlearn/common/attention.py
```diff
@@ -122,6 +121,9 @@
     maybe_instantiate,
 )
 from axlearn.common.flash_attention.remat import FLASH_ATTN_RESIDUAL_NAME
+from axlearn.common.kv_cache.base_kv_cache import BaseKVCache, KVState
+from axlearn.common.kv_cache.kv_cache import KVCache
+from axlearn.common.kv_cache.sliding_window_kv_cache import SlidingWindowKVCache
```
ruomingp, line 126: Not blocking this PR, but over the long run we'd like to remove this dependency.

ruomingp (approved):

axlearn/common/attention.py
hanzhi713: Done.

kelvin-zou (approved):

ruomingp (approved):

PR #1251, opened 2025-06-10
`CompositeLossMetrics` now performs a weighted sum of losses.

ds-hwang:
Currently, `CompositeLossMetrics` sums the losses without considering their weights (i.e., the number of live targets). To make this a weighted sum, downstream code has been implementing `CompositeLossWeights` to inject the number of live targets into `loss_weights`. This is essentially patching a surprising logic (initail loss sum) with complex logic (CompositeLossWeights) into a straightforward one (weighted sum).

Therefore, we’re changing the default loss aggregation logic to be straightforward from the beginning.

From now on, our standarized loss aggregation logic is
```
loss = sum(each_loss_weight * each_loss * num_each_samples) / sum(each_loss_weight * num_each_samples)
```

Historically, the complex logic was introduced because the weights of losses returned by child metrics were unknown. But now that child metrics return losses as `WeightedScalar`, we can adopt a simpler, cleaner aggregation logic.

Note: alternative formulation could be
```
loss = sum(each_loss_weight * each_loss * num_each_samples) / sum(num_each_samples)
```
However, when num_each_samples is large and each_loss_weight is small, the denominator can become disproportionately large. So we discard this option.

### DIFF
--- a/axlearn/common/causal_lm.py
+++ b/axlearn/common/causal_lm.py
@@ -229,21 +229,6 @@ def _update(x: dict, updates: dict):
     x.update(updates)
 
 
-class CompositeLossWeights(Module):
-    """Computes loss weights."""
-
-    def forward(self, child_metrics: dict[str, tuple[Tensor, Nested[Tensor]]]) -> dict[str, Tensor]:
-        """Computes per-child loss weights from child metrics.
-
-        Args:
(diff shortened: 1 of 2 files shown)


markblee (commented):
(Will approve after the internal review completes.)

markblee (approved):

PR #1260, opened 2025-06-14
Revert "Do not materialize full text/image/speech logits (10GB for 16k seq text)"

markblee:
This reverts commit 2e3c00fdbb9af83b5bd1721d2bd3dcb37b5db6ea.


ruomingp (approved):

ds-hwang (approved):
Thank you for quick action.

PR #1262, opened 2025-06-14
Update vm image.

markblee:


axlearn/cloud/gcp/vm.py
```diff
@@ -376,7 +376,7 @@ def list_disk_images(creds: Credentials) -> list[str]:
     )
     image_names = []
     for el in images["items"]:
-        if "ubuntu-2004" in el["name"] and "arm" not in el["name"]:
+        if "ubuntu-2204" in el["name"] and "arm" not in el["name"]:
```
Ethanlm, line 379: Looks like we need the disk size to be >= 100GB. Can we change default value of `--disk_size` to 128?

PR #1264, opened 2025-06-16
Fix the crash on WandBWriter.format_config() with v2 structure.

ds-hwang:
wandb.config.update() requires only python primitives, and doesn't support enum.Enum. So wandb logging crashes with Transformer norm v2, because key is not string.
e.g. norm_cfg = {NormPosition.IN_NORM: RmsNorm.default_config()}

### DIFF
--- a/axlearn/common/checkpointer_test.py
+++ b/axlearn/common/checkpointer_test.py
@@ -436,6 +436,7 @@ class CheckpointerTest(test_utils.TestCase):
             ckpt.stop()
 
     @parameterized.parameters([Checkpointer, OrbaxCheckpointer])
+    @pytest.mark.skip(reason="TODO(mark-b-lee): figure out why it fails on CI.")
     def test_input_iterator(self, checkpointer_cls):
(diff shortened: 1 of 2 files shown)


apghml (commented):
It looks like you skip one more test in the internal pr?

markblee (approved):

axlearn/common/checkpointer_test.py
```diff
@@ -436,6 +436,7 @@ def test_custom_dict(self, checkpointer_cls, custom_dict_type):
             ckpt.stop()
 
     @parameterized.parameters([Checkpointer, OrbaxCheckpointer])
+    @pytest.mark.skip(reason="TODO(mark-b-lee): figure out why it fails on CI.")
```
markblee, line 439: @samos123 it seems that after the orbax upgrade, these tests now fail with OrbaxCheckpointer? https://github.com/apple/axlearn/commit/8c699b369e96d239eaff8916de4928b6ce32a75b

PR #1270, opened 2025-06-21
Backwards compatibility for PartitionSpec config.

markblee:


axlearn/common/config.py
```diff
@@ -259,6 +258,28 @@ def validate_config_field_name(name: str) -> None:
...
+            validate_fn=lambda _: None,
+        )
+    except (ImportError, AttributeError):
+        pass
+
+
+# Register other validators for convenience and backwards compat.
+# We allow these to be optional to avoid a hard dependency.
+_maybe_register_optional_type("numpy", "dtype")
+# As of 0.6.1, PartitionSpec is not longer a tuple.
```
samos123, line 279:
```suggestion
# As of 0.6.1, PartitionSpec is no longer a tuple.
```

markblee: Thanks, fixed.

ruomingp (approved):

PR #1274, opened 2025-06-25
Reduce verbose logging in evalers.

ds-hwang:
The following log appears too frequently and makes it hard to find the actual model loss: `Skipping eval, as step (497) < min_step (10000).`


ds-hwang:
@markblee Could you take a look? From 1442

PR #1278, opened 2025-06-27
Workload Hang & Rolling Window Goodput Monitoring Support

dipannita08:
### Overview

This PR introduces significant enhancements to our Goodput measurement system, focusing on improved monitoring capabilities and robustness for long-running AxLearn training jobs.

Specifically, it adds:
1.  **Workload Hang Monitoring Support:** Detects and flags instances where the training workload might be stuck or unresponsive.
2.  **Rolling Window Goodput Monitoring Support:** Provides insights into goodput over dynamic, recent timeframes, complementing cumulative goodput metrics.
3.  **Final Upload & Safe Exit:** Ensures that all outstanding metrics are flushed and monitoring processes gracefully terminate upon job completion or shutdown.
4.  **Updated Documents & Dashboard Templates:** Documentation is updated with new metrics and points to custom dashboard templates for GCM.
5.  **More Disruption, Checkpointing & GCS Badput Breakdown:** More Badput Buckets

We also switch to context managers to demarcate start and end of recording and monitoring events. These features aim to provide more granular, real-time insights into training efficiency and promptly identify potential issues, leading to more stable and performant training workflows.

### Testing

These monitoring features have been rigorously tested on long-running AxLearn training jobs to ensure their stability and accuracy under realistic conditions.

Specifically, validation runs on Fuji 7B & test models, example results:
- Rolling Window Goodput [1d, 3d, 5d]: 
image
- Workload Hang Monitoring: 
image
- Cumulative Goodput: 
image
- Cumulative Badput: 
image


axlearn/cloud/gcp/measurement.py
```diff
@@ -4,7 +4,7 @@
 
     Example:
 
-    # Enable Goodput when launching an AXLearn training job
+    # Enable Goodput when launching an AxLearn training job
```
markblee, line 7: Unintended change?

axlearn/cloud/gcp/measurement.py
```diff
@@ -38,13 +43,22 @@ class Config(measurement.Recorder.Config):
...
         """
 
         upload_dir: Required[str] = REQUIRED
         upload_interval: Required[int] = REQUIRED
-        step_deviation_interval_seconds: int = 30  # Default to 30 seconds
+        enable_gcp_goodput_metrics: bool = True
+        enable_pathways_goodput: bool = False
+        include_badput_breakdown: bool = True
+        enable_rolling_window_goodput_monitoring: bool = False
+        rolling_window_size: Sequence[int] = ()
```
markblee, line 61: We try to avoid adding these kinds of bool feature flags in the API because they quickly become unmaintainable as we need to account for the cross product of all bool interactions.

Can `enable_rolling_window_goodput_monitoring` be inferred from whether `len(rolling_window_size) > 0`? 
Can `enable_pathways_goodput` be inferred automatically if we detect proxy backend?
When would we want `include_badput_breakdown` to be False? 
When would we want `enable_gcp_goodput_metrics` to be False? Should the user simply not configure this recorder if goodput metrics are not desired?

axlearn/cloud/gcp/measurement.py
```diff
@@ -53,68 +67,83 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
...
-        """Starts Monitoring of Goodput.
+        try:
+            yield
+        finally:
+            try:
+                if record_event_end:
+                    record_event_end(*args, **kwargs)
+            except (TypeError, ValueError, RuntimeError) as e:
+                logging.warning(
+                    "Failed to record end of event %s. Error: %s", event.name, e, exc_info=True
```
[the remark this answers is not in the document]
dipannita08: Updated.

axlearn/cloud/gcp/measurement.py
dipannita08: Updated. All new configs other than `rolling_window_size` are now inferred.

axlearn/cloud/gcp/measurement.py
```diff
@@ -123,33 +134,68 @@ def start_monitoring(self, *args, **kwargs):
...
                 )
                 self._monitor = goodput_monitoring.GoodputMonitor(
-                    job_name=cfg.name,
-                    logger_name=f"goodput_logger_{cfg.name}",
-                    tensorboard_dir=cfg.upload_dir,
-                    upload_interval=int(cfg.upload_interval),
+                    job_name=self.config.name,
+                    logger_name=f"goodput_logger_{self.config.name}",
+                    tensorboard_dir=self.config.upload_dir,
+                    upload_interval=self.config.upload_interval,
```
markblee, line 149: Same comment here.

axlearn/cloud/gcp/measurement.py
```diff
@@ -123,33 +134,68 @@ def start_monitoring(self, *args, **kwargs):
...
                     monitoring_enabled=True,
+                    pathway_enabled=pathways_enabled,
```
markblee, line 151: Is `pathway_enabled` intentional and not a typo?

axlearn/cloud/gcp/measurement.py
```diff
@@ -123,33 +134,68 @@ def start_monitoring(self, *args, **kwargs):
...
-                self._monitor.start_step_deviation_uploader(*args, **kwargs)
+            yield
+        finally:
+            if self._monitor:
+                self._monitor.stop_goodput_uploader()
+                logging.info("Flushed final metrics and safe exited from Goodput monitoring.")
+
+    @contextlib.contextmanager
+    def maybe_monitor_rolling_window_goodput(self):
+        """Monitor rolling window goodput if enabled."""
```
markblee, line 165: Please comment on the differences w/ `maybe_monitor_goodput` and when one should use one over the other.

axlearn/cloud/gcp/measurement.py
```diff
@@ -123,33 +134,68 @@ def start_monitoring(self, *args, **kwargs):
...
+        if not self.config.rolling_window_size or jax.process_index() != 0:
+            yield
+            return
+        try:
+            if self._rolling_window_monitor is None:
+                pathways_enabled = (
+                    hasattr(flags.FLAGS, "jax_backend") and flags.FLAGS.jax_backend == "proxy"
+                )
```
markblee, line 173: It may be better to infer this in `from_flags` to keep the rest of code flag-agnostic. We can use a config field `jax_backend: Optional[str] = None`, which seems better than a `is_pathways` bool since it can be extended to other backends.

axlearn/cloud/gcp/measurement.py
```diff
@@ -123,33 +134,68 @@ def start_monitoring(self, *args, **kwargs):
...
+                self._rolling_window_monitor.stop_rolling_window_goodput_uploader()
                 logging.info(
-                    "Started Step Deviation upload to Tensorboard & GCM in the background!"
+                    "Flushed final metrics and safe exited from Rolling Window Goodput monitoring."
                 )
+
+
+def create_goodput_recorder(cfg: GoodputRecorder.Config):
+    """Factory method to create GoodputRecorder."""
+    return GoodputRecorder(cfg)
```
markblee, line 201: Where is this used / do we need this? Usually we instantiate configs directly.

axlearn/common/measurement.py
```diff
@@ -60,9 +51,19 @@ def from_flags(cls, fv: Optional[flags.FlagValues]) -> "Recorder":
         raise NotImplementedError(cls)
 
     def record(self, event: Event, *args, **kwargs):
-        """Records an event with the given name."""
+        """DEPRECATED: Records a single point-in-time event."""
```
markblee, line 54: Should we just remove it? I don't think we have any other recorders.

axlearn/common/measurement.py
```diff
@@ -136,10 +137,15 @@ def initialize(fv: flags.FlagValues):
 
 def record_event(event: Event):
     """Records a global event."""
-    if global_recorder is None:
-        logging.log_first_n(logging.INFO, "No recorder configured, ignoring events.", 1)
+    # pylint: disable=unused-argument
+    if global_recorder and hasattr(global_recorder, "record_event"):
+        logging.log_first_n(
+            logging.WARNING,
+            "Global record_event call is deprecated, use the context managers instead",
```
markblee, line 144: Likewise with this fn?

axlearn/common/trainer.py
```diff
@@ -241,116 +241,125 @@ def __init__(
...
-            )
-
-        self._per_param_train_dtype = maybe_instantiate(
-            canonicalize_per_param_dtype(cfg.train_dtype)
+        init_events_manager = (
+            self._recorder.record_event(measurement.Event.ACCELERATOR_INIT)
+            if self._recorder
+            else contextlib.nullcontext()
         )
+        with init_events_manager:
```
markblee, line 249: Instead of nesting everything under a context, consider moving the block into a helper function and either decorating or wrapping only that helper. 

```
    with init_events_manager:
        self._init()

def _init(self):
    # the original block
```

It's also seems like we can limit the scope of what goes into this block so we can be more specific about what we're actually measuring in that block.

axlearn/common/trainer.py
```diff
@@ -554,6 +559,17 @@ def run(
...
+        goodput_monitor_manager = (
+            self._recorder.maybe_monitor_goodput()
+            if hasattr(self._recorder, "maybe_monitor_goodput")
+            else contextlib.nullcontext()
+        )
+        rolling_goodput_monitor_manager = (
+            self._recorder.maybe_monitor_rolling_window_goodput()
+            if hasattr(self._recorder, "maybe_monitor_rolling_window_goodput")
+            else contextlib.nullcontext()
+        )
```
markblee, line 571: The purpose of the recorder interface is to decouple the callsites from the specific recorder. It seems like the abstraction isn't quite clear since we have to explicitly call `goodput_xyz` methods.

axlearn/cloud/gcp/measurement.py
```diff
@@ -53,68 +57,75 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
...
             )
 
-    def start_monitoring(self, *args, **kwargs):
-        """Starts Monitoring of Goodput.
+        try:
+            yield
+        finally:
+            try:
+                if record_event_end:
+                    record_event_end(*args, **kwargs)
```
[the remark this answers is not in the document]
dipannita08: In the sequence you described, all the record logs will be ignored - same as before. This way will not introduce new record functionality. The goal is here to make user code easier for recording by reducing the number of lines of code in the call site.

axlearn/cloud/gcp/measurement.py
dipannita08: Intentional, not a typo.

axlearn/cloud/gcp/measurement.py
```diff
@@ -53,68 +67,83 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
...
+            except (TypeError, ValueError, RuntimeError) as e:
```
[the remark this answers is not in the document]
dipannita08: Done.

axlearn/cloud/gcp/measurement.py
```diff
@@ -53,68 +57,75 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
...
+        # Lazily instantiate the recorder if it hasn't been already.
         if self._recorder is None:
-            cfg: GoodputRecorder.Config = self.config
+            if jax.process_index() == 0:
+                logging.info("Lazily instantiating goodput recorder.")
             self._recorder = goodput.GoodputRecorder(
-                job_name=cfg.name,
-                logger_name=f"goodput_logger_{cfg.name}",
+                job_name=self.config.name,
+                logger_name=f"goodput_logger_{self.config.name}",
```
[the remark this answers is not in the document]
dipannita08: Removed redundant copies.

axlearn/cloud/gcp/measurement.py
dipannita08: Done.

axlearn/common/measurement.py
dipannita08: Removed.

axlearn/common/trainer.py
dipannita08: Simplified the `recorder` context managers. Leaving the `monitor` context managers as is due to lack of smaller abstractions.

axlearn/cloud/gcp/measurement_test.py
```diff
@@ -3,191 +3,206 @@
...
+
+    @mock.patch("jax.process_index", return_value=0)
+    def test_record_event_context_manager(self, _):
+        """Tests the record_event context manager."""
+        cfg = GoodputRecorder.default_config().set(
+            name="test",
+            upload_dir="/tmp/test",
+            upload_interval=1,
         )
-        fv.mark_as_parsed()
```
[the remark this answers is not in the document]
dipannita08: Some test scenarios which rely on the same APIs (goodput, step time deviation) are tested using multiple test params rather than separate tests, other scenarios which rely on different APIs are separated into different tests. Previous test functionality remains the same and more different scenarios are added. There is no test detail that existed before that is removed in this version.

apghml (commented):
Thanks, I've responded to your comments. Could you additionally make sure that you respond to any remaining open reviewer comments that you haven't already responded to?

axlearn/cloud/gcp/measurement.py
```diff
@@ -38,13 +42,15 @@ class Config(measurement.Recorder.Config):
         Attributes:
             upload_dir: Directory to store metrics for the monitor.
             upload_interval: Time interval (seconds) for monitoring uploads.
-            step_deviation_interval_seconds: Time interval (seconds) for step deviation metrics
-            uploads. -1 to disable step deviation uploads.
+            rolling_window_size: A sequence of integers defining the rolling window sizes in
+                seconds.
+            jax_backend: Jax backend type to infer Pathways environment.
```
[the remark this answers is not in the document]
dipannita08: Correct. If it specified and not set to Pathways backend (proxy) then Goodput for Pathways is disabled. Essentially it doesn't matter what it is set to if Pathways is not used.

axlearn/cloud/gcp/measurement.py
dipannita08: Oops, thought I addressed this but missed it. 

It is not required, we use configs to instantiate here as well. Removing.

axlearn/common/trainer.py
dipannita08: Agreed, that does make trainer cleaner. Will update!

docs/05-Goodput-Monitoring.md
```diff
@@ -94,24 +112,22 @@ To enable Goodput recording and monitoring on AXLearn, follow the example below.
...
-   axlearn gcp launch run --instance_type=tpu-v5litepod-16 \
+axlearn gcp launch run --instance_type=tpu-v5litepod-16 \
         --bundler_type=artifactregistry --bundler_spec=image=tpu \
         --bundler_spec=dockerfile=Dockerfile \
-        --name=<unique-readable-name> \
-        -- python3 -m ...training-config... \
+        -- python3 -m my_training_job \
         --recorder_type=axlearn.cloud.gcp.measurement:goodput \
-        --recorder_spec=name=<unique-readable-name> \
+        --recorder_spec=name=my-run-with-goodput \
```
Ethanlm, line 126: What's the motivation for this change? Is unique name not required?

axlearn/common/trainer.py
```diff
@@ -241,58 +241,9 @@ def __init__(
...
-                logging.warning(
-                    "xsc_check_policy was set for non-TPU XLA backend. Running without XSC."
-                )
-            else:
-                xsc_check_policy = maybe_instantiate(cfg.xsc_check_policy)
-        self._xsc_check_policy: Optional[Callable[[int], bool]] = xsc_check_policy
-        self._compiled_train_step: Optional[jax.stages.Compiled] = None
+        # Accelerator initialization.
+        with self._record_event(measurement.Event.ACCELERATOR_INIT):
+            self._device_init(devices)
```
Ethanlm, line 246: Looks like we are downscoping "ACCELERATOR_INIT" here.  I think we have been using "ACCELERATOR_INIT" event for the whole `__init__` call.   Can we keep it that way?

If we want to make it more accurate, I believe `ACCELERATOR_INIT` is only about mesh initialization.  But knowing how long it takes for the whole `__init__` call is probably more valuable

Ethanlm (commented):
Left some comments. Thanks for the work!

axlearn/cloud/gcp/measurement.py
```diff
@@ -13,10 +16,14 @@
         --recorder_spec=name=my-run-with-goodput \
         --recorder_spec=upload_dir=my-output-directory/summaries \
         --recorder_spec=upload_interval=30 \
-        --recorder_spec=step_deviation_interval_seconds=30
+        --recorder_spec=rolling_window_size=86400,259200,432000
```
Ethanlm, line 19: We decide to use last 1/3/7 rolling window.

In the examples, can we change to 
```
86400,259200,604800
```
? Tgabjs

axlearn/common/measurement.py
```diff
@@ -15,30 +16,20 @@ class Event(enum.Enum):
     """Event to be recorded.
 
     Attributes:
-        START_JOB: Start of job.
```
heyaudace, line 18: Out of curiosity, I'm wondering whether it makes sense to introduce such breaking changes without notice since we have experiments that are relying on the current API.

One option could be to improve the good put recorder implementation first and then introduce the breaking changes after?

Since axlearn is an open-source project, it might also impact other users outside of Apple?

axlearn/cloud/gcp/measurement.py
dipannita08: Done (single copy only).

axlearn/cloud/gcp/measurement.py
dipannita08: Removed.

axlearn/common/measurement.py
dipannita08: Removed.

axlearn/cloud/gcp/measurement.py
```diff
@@ -53,68 +59,78 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
...
-            self._recorder.record_custom_badput_event_start_time(*args, **kwargs)
-        elif event == measurement.Event.END_CUSTOM_BADPUT_EVENT:
-            self._recorder.record_custom_badput_event_end_time(*args, **kwargs)
-        else:
-            logging.log_first_n(
-                logging.WARNING,
-                "Ignoring unknown event %s",
-                1,
-                event,
+        start_method_name = f"record_{event.value}_start_time"
```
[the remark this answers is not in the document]
dipannita08: Yes, motivation is mainly to de-clutter the call-sites for recorder calls since those might evolve and grow in the future (as we have been experiencing so far).

There have been no changes to the recorder APIs (no new or deprecated APIs or types of events). No new recorder API calls have been added in the user-code either. GoodputRecorder (ml-goodput-measurement) functionality remains unchanged.

I added new sets of succeeding and failure scenario unit tests for every type of event we used. The tests should help integrate and catch issues in any future scenarios as well.

axlearn/cloud/gcp/measurement.py
```diff
@@ -38,13 +42,15 @@ class Config(measurement.Recorder.Config):
         Attributes:
             upload_dir: Directory to store metrics for the monitor.
             upload_interval: Time interval (seconds) for monitoring uploads.
-            step_deviation_interval_seconds: Time interval (seconds) for step deviation metrics
-            uploads. -1 to disable step deviation uploads.
+            rolling_window_size: A sequence of integers defining the rolling window sizes in
```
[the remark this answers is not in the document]
dipannita08: Correct, cumulative metrics (Goodput, Badput, Step Time Deviation, Ideal Step Time, Disruptions Count, Maximum Productive Step Count) are computed and uploaded at the same interval (`upload_interval`). Rolling window metrics (Goodput & Badput Breakdown) can be configured independently.

axlearn/common/trainer.py
dipannita08: You're right! Reverted to the original scope!

dipannita08:
Thanks for the reviews. PTAL.

axlearn/common/measurement.py
heyaudace: That's not the case actually, unless the `axlearn/common/measurement.py` changes are unintentional.

What's happening here is that if another client builds on top of the recorder api (e.g implementations a custom recorder), then the new axlearn version will start breaking.

I discussed this with @Ethanlm internally and he can fill you in on the context.

Right now this change will break internal experiments which use a different recorder than the good put recorder although the different recorder implements the same API.

My suggestion at the moment would be to only change the good put recorder but not break the API

axlearn/common/launch_trainer_main.py
```diff
@@ -13,7 +13,6 @@ def main(_):
     launch.setup()
     trainer_config = launch_trainer.get_trainer_config()
     trainer_config.set(recorder=config_for_function(lambda: measurement.global_recorder))
-    measurement.start_monitoring()
```
heyaudace, line 16: This will also be another breaking API change cc @Ethanlm

axlearn/common/measurement.py
dipannita08: Thanks for the context and discussions on the internal code-base dependencies, I wasn't aware of these (there is no way for me to test them either.)
The global recorder should now be completely backward compatible, it adds a new monitoring api `maybe_monitor_all_goodput` which is currently only implemented by the GoodputRecorder version in gcp/..

axlearn/common/trainer.py
```diff
@@ -566,6 +575,7 @@ def run(
             self.mesh(),
             jax.log_compiles(self.vlog_is_on(1)),
             self._context_manager(),
+            self._recorder.maybe_monitor_all_goodput(),
```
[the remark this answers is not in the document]
ruhanprasad: In addition, `self._recorder` could be `None`.  We should maybe return null context if that is the case.  If a launch without any recorder enabled, we'll encounter an error on this line.  Other than that LGTM

axlearn/common/trainer.py
```diff
@@ -566,6 +575,11 @@ def run(
             self.mesh(),
             jax.log_compiles(self.vlog_is_on(1)),
             self._context_manager(),
+            (
+                self._recorder.maybe_monitor_all_goodput()
```
Ethanlm, line 579: Should this be more general, instead of only about godput? Since `recorder` is not necessarily a goodput recorder?

axlearn/common/launch_trainer_main.py
Ethanlm: As @heyaudace mentioned, this also breaks the current API. Can we address it too?

dipannita08: Sure, added it back. Other RecorderTypes implementing this should work as is.

axlearn/common/trainer.py
dipannita08: Changed it to `maybe_monitor_all`.

axlearn/cloud/gcp/measurement.py
```diff
@@ -53,68 +66,81 @@ def from_flags(cls, fv: flags.FlagValues) -> "GoodputRecorder":
...
-    def record(self, event: measurement.Event, *args, **kwargs):
-        # Lazily instantiate the recorder. This avoids invoking jax before setup is complete.
+        self._recorder: Optional[goodput.GoodputRecorder] = None
+        self._monitor: Optional[goodput_monitoring.GoodputMonitor] = None
+        self._rolling_window_monitor: Optional[goodput_monitoring.GoodputMonitor] = None
+        self._job_name = cfg.name
+        self._logger_name = f"goodput_logger_{cfg.name}"
+
+    @contextlib.contextmanager
+    def record_event(self, event: measurement.EventType, *args, **kwargs):
```
ruhanprasad, line 97: Even though I see we have reverted back to the `record` API in the call-sites, I see that the GoodputRecorder class is no-oping the `record` function here and obviously `record_event` is no longer called.  Won't this affect the Goodput monitoring?  Shouldn't `record` be defined here?

dipannita08: Reverted, rebased and re-squashed.

axlearn/common/measurement.py
```diff
@@ -41,6 +42,26 @@ class Event(enum.Enum):
...
+        DATA_LOADING: Start and end of data loading.
+        CUSTOM_BADPUT_EVENT: Start and end of custom badput events.
+    """
+
+    JOB = "job"
+    STEP = "step"
+    ACCELERATOR_INIT = "tpu_init"
+    TRAINING_PREPARATION = "training_preparation"
+    DATA_LOADING = "data_loading"
+    CUSTOM_BADPUT_EVENT = "custom_badput_event"
```
heyaudace, line 62: Are we actually aiming to keep this `EventType` class (as well as `Event`), or is this leftover code?

I'm not seeing this used anywhere in the trainer or tested in `axlearn/common/measurement_test`.
If it's only used in goodput recorder, maybe move the abstraction into goodput recorder implementation?

axlearn/cloud/gcp/measurement.py
```diff
@@ -109,47 +258,8 @@ def record(self, event: measurement.Event, *args, **kwargs):
...
-        Goodput and Badput at the upload_interval and upload to the specified TensorBoard
-        directory.
-        Note: This function requires initialization of distributed JAX before it is called.
-        If there are internal GCP errors from querying and uploading data, these will be
-        logged without affecting the workload. GoodputMonitor logs will provide further
-        information if data is not being uploaded correctly.
-
-        Default behavior is to push metrics to Google Cloud Monitoring.
-        This behavior can be overridden by configuring `goodput_monitoring.GCPOptions`
+    def start_monitoring(self, **kwargs):
```
Ethanlm, line 261: https://github.com/apple/axlearn/blob/a0144986dc6eb049107f99e4884f36cfeeb0cd87/axlearn/common/launch_trainer.py#L132

Since `start_monitoring` in GoodputRecorder is a dummy function in this PR,  and monitor initialization is done by `maybe_monitor_all`, but `maybe_monitor_all` is invoked in the `trainer.run` method,  we are calling

```
measurement.record_event(measurement.Event.START_JOB)
```
before monitor initialization.  This doesn't seem correct.

Can we fully revert back to original implementation, so we can get this functionality merged in, and worry about API changes in a standalone follow-up PR(s)?

Ethanlm: Thanks

Ethanlm (approved):
@ruhanprasad will do some final validation with internal CI. If works well, @ruhanprasad please approve and put a `ready-to-merge` label so this PR will be merged. 

Thanks

ruhanprasad (approved):
Internal CI passes, approving

axlearn/common/measurement.py
```diff
@@ -60,13 +81,56 @@ def from_flags(cls, fv: Optional[flags.FlagValues]) -> "Recorder":
...
 
+    @contextlib.contextmanager
+    def record_event(self, event: EventType, *args, **kwargs):
+        """A context manager to record the start and end of an event.
+
+        This is the preferred method for recording events. Child classes
+        should implement this context manager to handle timed operations.
+
+        Example:
+            with recorder.record_event(EventType.ACCELERATOR_INIT):
```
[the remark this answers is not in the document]
dipannita08: Yes, because the call-sites for GoodputRecorder and other internal/custom recorders are shared.

Summarizing, this PR:
- Introduces new common interfaces to manage recording and monitoring life-cycles better.
- Implements the new interfaces for Goodput.
- Call sites are backward compatible for internal/custom recorder types.
- GoodputRecorder (for now) continues using the older record API, and the newer monitor API to support both cumulative and rolling window monitoring.
- Functionality diff: 
    - With GoodputRecorder: Additional metrics can be monitored (Workload Hang, Rolling Window, Disruption, StepInfo etc.)
    - With Custom/Internal Recorders: None

In following PRs, we aim to:
- Implement context manager style APIs for record & monitor for internal/custom recorders.
- Update the call-sites.
- Deprecate the old style APIs & EventTypes, which each grow 2x for evert new type of event and makes the codebase complex with no real benefit.

changlan (approved):

axlearn/common/measurement.py
```diff
@@ -41,6 +42,26 @@ class Event(enum.Enum):
     END_CUSTOM_BADPUT_EVENT = "END_CUSTOM_BADPUT_EVENT"
 
 
+class EventType(enum.Enum):
+    """Event to be recorded.
```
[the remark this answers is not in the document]
dipannita08: Can we move these type of changes to following Refactoring type PRs?

PR #1288, opened 2025-07-08
Pathways Updates

lukebaumann:
Creates a Pathways-TPU image and separating the Pathways dependencies into a different extra


Ethanlm (commented):
This adds a new docker image and might not be necessary. 

Can I ask what's the motivation behind it?

lukebaumann:
Pathways TPU and McJAX TPU have different dependencies. Specifically `jax[tpu]` is only needed for McJAX TPU and `pathwaysutils` is only needed for Pathways TPU. Separating Pathways TPU and McJAX TPU as two packaging extras was the main intent of the PR and as a consequence I created a new docker image.

Today, Pathways TPU _can_ install `jax[tpu]` and McJAX TPU _can_ install `pathwaysutils` so this PR is not necessary but it does more accurately denote dependencies of different execution modes.

Ethanlm:
> Pathways TPU and McJAX TPU have different dependencies. Specifically `jax[tpu]` is only needed for McJAX TPU and `pathwaysutils` is only needed for Pathways TPU. Separating Pathways TPU and McJAX TPU as two packaging extras was the main intent of the PR and as a consequence I created a new docker image.
> 
> Today, Pathways TPU _can_ install `jax[tpu]` and McJAX TPU _can_ install `pathwaysutils` so this PR is not necessary but it does more accurately denote dependencies of different execution modes.


I would suggest to hold off splitting the dependencies for now, since this will break our internal setup. 

Making a separate image doesn't seem necessary and add maintenance overhead. It is nice to have both Pathways TPU and McJAX TPU working in one single image for easier testing.

We can go with other changes in the PR

axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -429,6 +429,11 @@ def _build_pathways_head_job(self):
         annotations = _LoadBalancer(
             jobset_name=cfg.name, replicated_job_name=_PATHWAYS_HEAD_REPLICATED_JOB_NAME
         ).metadata
+        annotations.update(
+            {
+                "alpha.jobset.sigs.k8s.io/exclusive-topology": "kubernetes.io/hostname",
```
Ethanlm, line 434: Is this to avoid scheduling two head pod into one node?

PR #1298, opened 2025-07-14
Retire remat_in_scan in repeat layer.

ds-hwang:
`remat_in_scan` is redundant because it can be fully replaced by `BaseLayer.remat_spec`.

For example, in `RepeatedTransformerLayer`, `remat_in_scan` is used in the form `scan(remat(layer))`. But if `layer.remat_spec` is already defined, then `layer.call` ends up applying `remat` redundantly. Even if we retire `remat_in_scan`, the same behavior can be fully replicated using `layer.remat_spec`.
Therefore, we are retiring `remat_in_scan`.

### DIFF
--- a/axlearn/common/repeat.py
+++ b/axlearn/common/repeat.py
@@ -67,7 +67,6 @@ from axlearn.common.config import (
     Required,
     config_class,
     config_for_function,
-    maybe_instantiate,
 )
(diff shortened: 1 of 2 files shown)


ds-hwang:
@markblee Could you review it? From 1434

PR #1299, opened 2025-07-15
Sliding window KV cache supports the flash attention path via standard attention fallback.

ds-hwang:
There are two ways to support the flash attention path with sliding window KV cache:
1. Using standard attention fallback [1], or
2. Modifying the TPU/GPU flash decoding to support sliding window KV cache.

[1] https://github.com/apple/axlearn/blob/bf6f65a7b53510bef0a621917f9b7d9f58bf1964/axlearn/common/flash_attention/layer.py#L199-L201

In this PR, I chose the first approach. I judged that the alternative does not provide significant speed gain relative to the added code complexity.

Flash attention gains speed primarily by eliminating O(T²) HBM read/write bottlenecks and skipping unnecessary blocks.

In the decoding path, since `q_len = 1`, there is no O(T²) HBM read/write bottleneck to begin with. Thus, most of the speed benefit comes from skipping unnecessary blocks. According to benchmarks [2], flash attention is about 5–20% faster than standard attention fallback in this case: [2] https://github.com/apple/axlearn/blob/main/axlearn/common/flash_attention/gpu_decoding.py#L34

This was measured in scenarios where the causal mask is used together with KVCache — where the KVCache holds up to `max_seq_len`, and the speed gain comes from skipping KV cache beyond the current time step.

However, in SlidingWindowKVCache, there is no cache to skip, since only the valid window is stored. Therefore, in the decoding path, there is neither an O(T²) HBM bottleneck nor blocks to skip, meaning there is no reason to use flash attention. The additional code complexity required to support flash attention in this case does not justify the minimal speed benefit, so this code path falls back to standard attention.

Moreover, when PagedKVCache is used in serving in the future, it will handle both local and global attention. Thus, there is no need to implement flash attention specifically for SlidingWindowKVCache.

**Benchmark**
I compared Flash Attention decoding and Standard Attention decoding on TPU v5p, checking both 16k (global) and 4k (local) KV cache configurations. On TPU, there are no benchmark cases where Flash decoding is faster than Standard decoding, in
[tpu_attention_benchmark.py](https://github.com/apple/axlearn/blob/main/axlearn/common/flash_attention/tpu_attention_benchmark.py).

Flash Attention decoding vs Standard Attention decoding (1-step decoding, TPU v5)

Model Size | Standard (16k) | Flash (16k) | HBM (16k) | Standard (4k) | Flash (4k) | HBM (4k)
-- | -- | -- | -- | -- | -- | --
1.2B | 0.4019 ms | 0.5910 ms | — | 0.2569 ms | 0.3755 ms | —
12.6B | 0.6049 ms | 1.1137 ms | 256.43M | 0.3214 ms | 0.6032 ms | 64.36M
29.6B | 0.2542 ms | 0.2652 ms | 640.46M | 0.1977 ms | 0.2143 ms | 160.38M
65.2B | 1.0380 ms | 1.7251 ms | 1.12G | 0.4452 ms | 0.8791 ms | 288.49M
134B | 1.1095 ms | 2.0664 ms | 1.38G | 0.4678 ms | 1.0134 ms | 352.51M
261.7B | 1.5863 ms | 2.4353 ms | 1.72G | 0.5682 ms | 1.2005 ms | 440.46M

This PR switches sliding window decoding from Flash (16k) to Standard (4k), reducing KV cache memory usage by 4× and improving speed by more than 4×.

While it's theoretically possible to modify the TPU/GPU decoding code to support sliding window KV cache as an alternative to Flash (4k), it offers no advantage over Standard (4k), so the additional effort isn't justified.

For 134B 4k case, this is profile results: Flash is 3-4 times slower.
* Standard (4k): 246us image

* Flash (4k): 800us image

**Conclusion**: Sliding window KV cache supports the flash attention path via standard attention fallback.

### DIFF
--- a/axlearn/common/flash_attention/common.py
+++ b/axlearn/common/flash_attention/common.py
@@ -19,7 +19,9 @@ from jax.experimental import pallas as pl
 from axlearn.common.attention import compute_gqa_context, compute_gqa_logits, softmax_with_biases
 from axlearn.common.attention_bias import BaseAttentionBias, MaskFn, SegmentIdAttentionBias
 from axlearn.common.config import Configurable, config_class
-from axlearn.common.kv_cache.paged_kv_cache import reconstruct_kv
+from axlearn.common.kv_cache.base_kv_cache import BaseKVCache
+from axlearn.common.kv_cache.kv_cache import KVCache
+from axlearn.common.kv_cache.paged_kv_cache import PagedKVCache, reconstruct_kv
 from axlearn.common.layers import dropout
 from axlearn.common.utils import Nested, Tensor, validate_contains_paths
 
@@ -143,7 +145,6 @@ class BaseFlashAttention(Configurable):
         """Configures BaseFlashAttention.
 
         Attributes:
-            is_decoding: Whether we're in decoding/inference mode.
             softmax_scale: Scale factor to apply to QK.
             dropout_rate: Dropout rate for attention probs.
             interpret: Whether to use interpret mode for Pallas kernels.
@@ -151,7 +152,6 @@ class BaseFlashAttention(Configurable):
             gpu_block_size: Block size for GPU pallas kernels.
         """
 
-        is_decoding: bool = False
         softmax_scale: float = 1.0
         dropout_rate: float = 0.0
         interpret: bool = False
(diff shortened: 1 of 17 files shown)


changlan (approved):

ds-hwang:
@ruomingp, @markblee could you still review? :')

PR #1309, opened 2025-07-18
Add axlearn/experiments/logistic_regression_tutorial.py

wangkuiyi:


changlan (approved):

axlearn/experiments/logistic_regression_tutorial.py
```diff
@@ -0,0 +1,229 @@
...
+
+This tutorial demonstrates how to train a basic logistic regression model
+on synthetic 2D data for binary classification.
+"""
+
+# ============================================================================
+# Step 1: Create the Tutorial File
+# ============================================================================
+
+# conda create -n axlearn python=3.10
```
apghml, line 13: It might create extra work to keep these installation instructions in sync with the official docs. Would it make sense just to link the installation docs instead?

axlearn/experiments/logistic_regression_tutorial.py
```diff
@@ -0,0 +1,229 @@
...
+# Step 3: Model Definition
+# ============================================================================
+
+
+class LogisticRegressionModel(BaseModel):
+    """Logistic regression model."""
+
+    @config_class
+    class Config(BaseModel.Config):
+        model: Required[InstantiableConfig] = REQUIRED
```
apghml, line 109: Add a comment here for what this is with an example? (And mention this field does't need to be an instance of BaseModel, despite the name?)

apghml (commented):
Thanks!

axlearn/experiments/logistic_regression_tutorial.py
wangkuiyi: Thanks! Done.

axlearn/experiments/logistic_regression_tutorial.py
wangkuiyi: Thanks! Renamed this field from model into backbone.

axlearn/experiments/logistic_regression_tutorial.py
apghml: This doesn't seem to have been done?

PR #1334, opened 2025-08-22
Adds pause-resume elasticity if the proxy backend is used.

lukebaumann:
* Added the changes to the jobset for elastic training to enable elasticity.
* Added changes to launch_trainer so that the pause_resume decorator is used.
* Set logging.raiseExceptions=True so that DATA_LOSS errors that occur in debug/info/other log calls are raise exceptions immediately.

As written, this will use Pause-Resume elasticity if Pathways is enabled.

This should be merged after #1335

### DIFF
--- a/axlearn/cloud/gcp/pathways_utils.py
+++ b/axlearn/cloud/gcp/pathways_utils.py
@@ -341,6 +341,7 @@ class PathwaysReplicatedJob(BaseReplicatedJob):
             f"--resource_manager_address=localhost:{_PATHWAYS_RESOURCE_MANAGER_PORT}",
             f"--server_port={_PATHWAYS_PROXY_PORT}",
             f"--gcs_scratch_location={staging_location}",
+            f"--num_elastic_slices={cfg.accelerator.num_replicas}",
         ]
         cmd_args.extend(xla_flags_from_options(self._xla_options).split())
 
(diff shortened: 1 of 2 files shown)


axlearn/cloud/gcp/pathways_utils.py
```diff
@@ -609,14 +610,19 @@ def _build_pathways_worker_job(
...
             {"alpha.jobset.sigs.k8s.io/exclusive-topology": "cloud.google.com/gke-nodepool"}
         )
+        # When a slice is restarted, each restarted VM in the slice counts
+        # against the backoff limit
+        backoffLimit = system.vms_per_slice * _PATHWAYS_BACK_OFF_LIMIT
+
+        # For elasticity, we want the slices to be able to restart many times.
+        # There is no way to set this to be unlimited so we set the backoffLimit
+        # very high.
+        backoffLimit *= 1000
```
samos123, line 620: Should this only be done when elasticity is enabled?

muyangyuapple: The only concern is that if there is a user code error, it will take very long for the job to eventually report failure. Which may cause user confusion.

axlearn/common/launch_trainer.py
```diff
@@ -159,8 +162,24 @@ def run_trainer(trainer_config: SpmdTrainer.Config) -> Any:
...
+        output = trainer.run(prng_key)
+        measurement.record_event(measurement.Event.END_JOB)
+        return output
+
+    if FLAGS.jax_backend == "proxy":
+        # pylint: disable-next=import-error,import-outside-toplevel
+        from pathwaysutils.elastic import manager
+        elastic_manager = manager.Manager()
+        max_retries = 5
+        timeout = 10 * 60  # ten minutes
```
muyangyuapple, line 177: What does this timeout impact? How does it affect the multi-host inference case?

lukebaumann: This `timeout` argument is passed to `wait_for_slices` and is how long the JAX workload will wait for all slices to be ready after one (or more) of them fail before raising a `TimeoutError`. See [wait_for_slices](https://github.com/AI-Hypercomputer/pathways-utils/blob/7c52d6f245bfce6c0ab6729f776ccd40074195c1/pathwaysutils/elastic/manager.py#L668) for more details.

The multi-host inference case should not enable elasticity or use `elastic_manager.pause_resume`. Instead, it should rely on `LeaderWorkerSet` for resiliency mechanisms.

PR #1360, opened 2026-02-09
Upgrade to python 3.12 in pre-commit test.

raytomato:
Update pre-commit setup following axlearn Python upgrade.


wangkuiyi (commented):
Could @changlan plase also take a look?

